PR #1, opened 2024-03-08
Compute fluxes of country aggregations

melodb:
- Create dictionary of regions (country aggregations)
- Correct scaling of covariance matrix
- Compute regional fluxes from sum of individual country fluxes
- Compute regional prior uncertainty from sum of prior sigma2 (approximation)
- Compute regional posterior uncertainty from covariance matrix

### DIFF
--- a/PARIS_inversion_results.py
+++ b/PARIS_inversion_results.py
@@ -106,13 +106,27 @@ model_q_indices = {'intem':[0,1],
                    'rhime':[1,2],
                    'elris':[0,1]}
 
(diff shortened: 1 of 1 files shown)


mo-aliceramsden:
This looks great to me, thanks for adding this option in. 
I've tested this a few ways with different region aggregations, including some test options I added to regions_dict and everything is working for me. I like how the posterior uncertainty is only calculated if the covariance is included in the file, but the posterior mean is always plotted. That should allow for some basic comparison whilst we're still working on the InTEM covariance output. I'll merge this into devel now, as everything is working well.

PR #7, opened 2024-04-04
HFC updates

melodb:
Python script:
- Add HFC-32/125/227ea to dictionaries
- Correction in plot_spatial_flux to plot distribution of one model only
- Add lon/lat of ITALY and SWITZERLAND in plot_spatial_flux

Notebook:
- Add uYmod to list of available variables in the timeseries plot
- Add instructions for timeseries plotting
- Add ITALY and SWITZERLAND to list of available regions for plotting of spatial distribution


aliceramsden (approved):
This is all working really well for me, so this can be merged into devel with no changes.
I like the additional option for diff_include in the mf plots, that should be really useful.

PR #8, opened 2024-04-12
Add features to country flux and mf plotting

melodb:
- Add number of obs per model on mf timeseries plot
- Add option to remove plotting of prior uncertainty in country flux timeseries
- Add option to plot one legend for the entire figure in country flux timeseries


aliceramsden (approved):
The 'set_global_leg' variable wasn't working for me so I just updated the line where it checks this variable. That could just have been a Python version thing, but hopefully it should work for everyone now. And the plots look so much better with a single legend now, good idea!

All your other updates are working really well so this can definitely be merged in.

PR #9, opened 2024-04-19
Plot inventory of country aggregates + other updates

melodb:
- Compute inventory data of country aggregates from sum of individual countries
- Look for alternative file if std* does not exist

Also:
- correct bug in LUX
- edit some warning messages

### DIFF
--- a/PARIS_inversion_results.py
+++ b/PARIS_inversion_results.py
@@ -181,7 +181,7 @@ countrycodes_dict = {'IRELAND':'IRL',
                      'AUSTRIA':'AUT',
                      'ITALY':'ITA',
                      'BELGIUM': 'BEL',
-                     'LUXEMBURG': 'LUX'}
+                     'LUXEMBOURG': 'LUX'}
 
 regions_dict = {'BELUX':'BEL-LUX',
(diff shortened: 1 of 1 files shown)


aliceramsden (approved):
This all works well for me. Apart from when using the 'CW_EU' region, when the code can't find inventory data for Czechia, so it doesn't output any inventory data. But your error messages deal with this well and the plots are still output with no other issues.

So I can make a note to check if we have inventory data for Czechia and I'll receate the inventory files if we do. But for now I think this can be merged in as everything else is working well.

melodb:
Other option is to make the code ignore the countries that do not exist in the inventory and plot the sum of the remaining countries. However, this can lead to misinterpretation of the results if the country missing is a big contributor.

mo-aliceramsden:
> Other option is to make the code ignore the countries that do not exist in the inventory and plot the sum of the remaining countries. However, this can lead to misinterpretation of the results if the country missing is a big contributor.

Yes, I think you're right, that would most likely lead to misinterpretation of the results, in a lot of cases. I've made a note to check if there's other countries that we're missing inventory data for, so I can try and fix that next week.

PR #11, opened 2024-04-19
Add species database

mo-aliceramsden:
Species info is now read in from a species_info.json file. This should hopefully make it easier to add more species in the future.

Some of the info I added may not be correct (e.g. the units variables) and will need updating once we start plotting results for these gases. But those should be easy fixes to make if we do find any issues.


melodb (approved):
I just corrected the previous developments (plot inventory of country aggregates and find alternative file) to be in agreement with the new approach. All working fine from my side. Ready to merge.

PR #13, opened 2024-05-03
added fix for error in CW_EU name and updated obs plotting to be clearer

mo-aliceramsden:
The code should now be able to find the CW_EU flux region from both old InTEM files (which contain the incorrect CW_EU definition) and the new InTEM files (where this definition is fixed.

There's also some minor updates to the obs plotting and I've increased the size of the colours in the legend as this was tricky to see when sharing results during meetings.

### DIFF
--- a/PARIS_inversion_results.py
+++ b/PARIS_inversion_results.py
@@ -1475,7 +1475,7 @@ def plot_spatial_flux(ds_all,species,plot_area,model_labels):
     cbar_diff.set_clim(difflim[species])
 
     color_bar3 = fig.colorbar(cbar_diff,orientation='vertical',extend='both',ax=ax[2,...],shrink=0.9,pad=0.005)
-    color_bar3.set_label(f'Prior - posterior {s_data[species]["species_print"]}\n{time_out}\n(mol m$^{{-2}}$ s$^{{-1}}$)')
+    color_bar3.set_label(f'Posterior - prior {s_data[species]["species_print"]}\n{time_out}\n(mol m$^{{-2}}$ s$^{{-1}}$)')
     
     return fig
 
--- a/PARIS_inversion_results.py
+++ b/PARIS_inversion_results.py
@@ -494,19 +494,26 @@ def plot_obs_modelled_separate(ds_all,species,site,model_labels,
                 if len(include) == 1:
                     ax.scatter(ds_all[m].time.values,
                                ds_all[m]['Yobs'].values,
-                               color='black',label=f'Obs ({model_labels[m]})',s=8,alpha=0.8,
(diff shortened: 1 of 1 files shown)


PARIS_inversion_results.py
```diff
@@ -492,20 +494,19 @@ def plot_obs_modelled_separate(ds_all,species,site,model_labels,
                 if len(include) == 1:
                     ax.scatter(ds_all[m].time.values,
                                ds_all[m]['Yobs'].values,
-                               color=model_colors[m][var_colors[var]],label=f'Obs ({model_labels[m]})',s=8,alpha=0.8)
-
+                               color='black',label=f'Obs ({model_labels[m]})',s=8,alpha=0.8,
+                               marker='s')
```
melodb: The purpose of this ax.scatter inside if (var == 'Yobs') & if (len(include) == 1) is to plot the obs of each model on its own color (model_colors[m][var_colors[var]]). This would be ok because Yobs is the only variable being plotted. If we don't want to do that, I would recommend to call ax.scatter outside "if len(include) == 1" and to delete the else of this if.

PARIS_inversion_results.py
```diff
@@ -492,20 +494,19 @@ def plot_obs_modelled_separate(ds_all,species,site,model_labels,
...
+                    
                     if add_unc:
-                        ax.fill_between(ds_all[m].time.values,
-                                        ds_all[m]['Yobs'].values - ds_all[m]['uYobs'].values,
-                                        ds_all[m]['Yobs'].values + ds_all[m]['uYobs'].values,
-                                        color=model_colors[m][var_colors[var]],alpha=0.2)
+                        ax.errorbar(ds_all[m].time.values,
+                                    ds_all[m]['Yobs'].values,
+                                    ds_all[m]['uYobs'].values,
+                                        color='black',alpha=0.4,fmt='none')
```
melodb: Similarly, here we can also consider using the color of each model for the error bars because Yobs is the only variable being plotted. Both are fine to me. I'm just highlighting this in case black was chosen to avoid confusion with other variables like Yapost.

melodb (approved):
Everything is working well from my side. I fixed a small bug in the legend of the spatial map. I also left 2 comments regarding the colors used to plot Yobs. Check what you think.

mo-aliceramsden:
Thanks for catching that bug! And good point about the plot colours when only plotting Yobs. 
I've edited this so that Yobs is plotted in the model colours when there's only one variable in 'include'. But when more than one variable is used, Yobs is always in black. Hopefully that makes these plots clearer when we're trying to plot lots of variables.

PR #15, opened 2024-05-16
Options to plot flux totals from more countries and options to choose colourmaps

mo-aliceramsden:
- Extended the country code dictionary so we can plot other country's fluxes in the future if needed (we'll need to look at Hungary in future years, so thought I'd add lots of other countries whilst I was doing this).

- Added an option in the notebook to choose colourmaps for the spatial fluxes.

- When running the notebook locally, we (at the Met Office, so this probably isn't an issue for anyone else!) found that the notebook kernel kept dying when rerunning 'func.slice_flux'. So I've split up this process slightly, which seems to have helped a little. This doesn't change the code at all, so older versions of the notebook should still work.


melodb (approved):
Everything is working fine on my end. Ready to merge from my side.

PR #16, opened 2024-05-21
option to over-ride standard inversion periods when reading in data

mo-aliceramsden:
A minor update to the notebook so we can override the inversion period that's specified in the species_info.json.
There's now a 'period_override' variable in the notebook which is used by the read_flux and read_mf functions to choose the inversion_period name at the end of the netCDF filenames.

E.g. if you want to plot both a monthly and annual PFC-218 inversion, this isn't possible with the old version of the code as this would assume the netCDFs end with 'yearly'. But with this update, the 'period' part of the filenames can be manually chosen using period_override.


mo-aliceramsden:
I've also added a bug fix for calculating inventory totals for aggregated regions:
- When a country had a mix of nans and values in its inventory timeseries, this country is now skipped in the total, as it was producing nan totals (I think this only impacts the CW_EU region for PFC-218 because France has nans inventory values for only some years).
- Fixed the tab spacing for plotting the inventory bars as this was repeating the plotting step for every country in the aggregated regions (we only found this bug when the aggregated region was used as the first subplot as the inventory label was plotted mulitple times in this setup for some reason!).

melodb (approved):
All is working well on my end. Ready for merging.

PR #17, opened 2024-05-29
Adapt timeseries plotting to new conc file

melodb:
Python script:
- Modifications were made to the three timeseries functions (separate, together, diff)
- Error bar in Yobs is now uYobs_repeatability (or previous uYobs variable if uYobs_repeatability does not exist)
- If uYmod does not exist, it is computed from qYmod as before
- The diff timeseries only works if the variable exists in both models

Notebook:
- List of possible variables in conc file was updated
- Bug fix for period_override


hdelongueville (approved):
Looks great!

mo-aliceramsden (approved):
Everything's working really well, happy for this to be merged!

PR #19, opened 2024-05-31
Accept regions reported as country instead of countrycodes_dict definition

hdelongueville:
Fix plot_country_flux when regions are reported as, for e.g., 'BENELUX' instead of 'BEL-LUX-NLD'. This allows the uncertainties of the RHIME regional emissions to be plotted.


melodb (approved):
All good from my side!

PR #20, opened 2024-06-06
Function to plot flux maps per time interval

melodb:
This function is similar to 'plot_spatial_flux', however, it allows the plotting of only one variable (prior, posterior or difference).
Differently from 'plot_spatial_flux', the average is not performed over the full time window specified in start_date/end_date but considering the number of time stamps specified in dt. If dt is an integer, the same value is considered for all models. If dt is a list, one value per model must be specified. A value of dt higher that the number of time stamps throws an error. This was made flexible so that yearly inversions can be easily compared with monthly inversions.

The design of the figure differs slightly from function 'plot_spatial_flux': the model name only shows up at the top; the time window of the averaging is used as subtitle; the white space between subplots was reduced; one color bar is shown for the whole figure.

To be solved:
1. The end date of the time window is misleading: the 1st day of the year is shown instead of the last. This is also the case for 'plot_spatial_flux'. Do you have any suggestion to solve this issue for both yearly and monthly inversions?

Potential improvements:
2. The average time window is specified according to the number of time stamps. For example, if I want the average at each 2 years, the value of 2 should be used for yearly inversions and the value of 24 should be used for monthly inversions. Is there a better way?
3. I did my best to define the colorbar/white space settings so that it would work for all possible number of lines and columns. It is not perfect but I would say it is good enough. Is there a more clever way of doing this?

### DIFF
--- a/PARIS_inversion_results.ipynb
+++ b/PARIS_inversion_results.ipynb
@@ -588,7 +588,7 @@
     "plot_area = 'CWEU'   #options for: UK, FRANCE, GERMANY, ITALY, SWITZERLAND, NWEU, CWEU, EUROPE\n",
     "cmap = 'coolwarm'    #flux colour (e.g. 'viridis', 'coolwarm')\n",
     "c_border = 'dimgrey' #country boundary colour (e.g. 'floralwhite', 'dimgrey')\n",
-    "dt = 2               #number of time stamps to use in the average\n",
+    "dt = 2               #number of time stamps to use in the averaging (use a list of integers to specify one value of dt per model)\n",
     "var = 'posterior_prior_diff'  #variable to be plotted.\n",
     "                              #options for: 'flux_total_posterior', 'flux_total_prior', 'posterior_prior_diff'\n",
     "###################################\n",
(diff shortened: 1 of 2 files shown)


aliceramsden (commented):
This looks great, I tested this on both yearly and monthly inversions with different averaging periods (different values of dt) and this all worked really well. 

I've pushed two commits which fix the print out of the end date of the averaging period. This is calculated based on either the inversion period given in the species_info.json file or the time periods listed in 'period_override', if this is used. I've added that fix to the two other spatial plotting functions too. There's probably a better way of coding up what I did, but it does seem to work for now!

I think the 'dt' option for averaging period works well as it is, and it is very customisable so should be useful for lots of different comparisons in the future, so I'm ok with this staying as it is.

The spacing of the plots and colour bar looks good too. The only thing I would suggest is to switch the layout of the subplots, so all the plots from one model are on the same row. Then, for example, if we had 5 years of annual plots, there would be 5 columns and 2 rows. This would be easier to view on the screen, rather than the current setup which would produce a 2 column 5 row plot. But that's partly just personal preference so happy to hear Helene's suggestions too!

melodb:
Thanks a lot for working on this! The end dates are just perfect!
Regarding the layout, I agree. I was not liking the look of it when plotting one single model. I will wait for Helene's feedback before changing it.

PR #23, opened 2024-06-13
Combine model results for region fluxes

mo-aliceramsden:
I've added plot_separate and plot_combined options to plot_country_flux. So individual models can be plotted as separate lines and/or averaged results from all models can be plotted as single lines.
For now, I've included 4 methods for averaging the results:
- mean of all posterior means
- mean of all posterior percentiles as the overall poterior uncerainty
- min/max of all posterior percentiles as the overall posterior uncertainty
- create posterior PDFs for each model by sampling from a Gaussian distribution with the posterior mean and posterior standard deviation from each model. These PDFs are then sampled to produce a 'combined' posterior PDF and the mean and standard deviation of this is taken. This method (for now) assumes that all models have Gaussian posteriors.

I've also restructured the plot_country_flux function, pulling out the sections which either search for the correct country index or calculate the aggregated region by summing/using the covariance matrices of individual countries. I've put all those steps into a separate function, so plot_country_flux only contains the plotting parts. This made it much easier to calculate the combined results and will hopefully this will make editing the code easier in the future.

### DIFF
--- a/PARIS_inversion_results.ipynb
+++ b/PARIS_inversion_results.ipynb
@@ -248,15 +248,15 @@
     "add_prior_unc = False #if True: plots prior uncertainty as shaded area\n",
     "set_global_leg = True #If True, plots one single legend instead of one legend per subplot.\n",
     "country_codes_as_titles = False #If True, lists 3-letter country codes under region names in subplot titles\n",
-    "include_separate = True #If True, includes all model results as separate lines\n",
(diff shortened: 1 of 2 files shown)


melodb (approved):
Thank you for adding this in. Everything is working very well. The rearrangement of plot_country_flux in several functions is very much welcome. I have added a small fix to the number of columns in the legend so that it looks nicer. Ready for merging from my side!

PR #24, opened 2024-06-17
Bug fix in timeseries plotting

melodb:
- Fix in the timeseries plotting to deal with old netcdf variables.
- To be merged after combine_model_results.


mo-aliceramsden (approved):
All looks good to me, thanks for spotting that bug!

PR #26, opened 2024-06-19
Fix to spatial flux lon/lon position relative to country borders

mo-aliceramsden:
The country borders and coastlines are plot assuming that the lat/lon arrays give the centre of the each grid cell.
So I've fixed the pcolormesh call which plots the grids of spatial fluxes so that it assumes the same coordinate setup. Before, pcolourmesh was assuming that the lat/lon arrays were giving the bottom left of each grid cell.

This moves fluxes half of one grid cell, and the prior fluxes now sit inside country borders as we'd expect.


melodb (approved):
All working well on my side. I didn't test with a corrected InTEM file, but at least for ELRIS the map seems better. Thanks for catching this! Mystery solved :)

PR #29, opened 2024-06-27
Various plotting updates

mo-aliceramsden:
Updates include:

- Option to plot point markers for key locations, either using a keyword (ne_england, london, paris) or a [lat,lon] location)
- Option to manually choose y axis limits in mole fraction plots
- Switched x and y axes of spatial fluxes by time plots (so there is a row for each model and time increases along the columns)
- Updated significant figures on stats in mole fraction histograms, so there's more significant figures for low concentration gases or small differences
- Added older inventory data and added option to plot inventory bars from mulitple years
- Added option to plot spatial fluxes averaged over seasons (e.g. Dec-Fev from all years)


melodb (approved):
Ready to merge from my side

PR #37, opened 2024-08-08
Plain notebook

melodb:
**1st commit:**
- only one file per inversion system in the header
- clean output cells
- add "season" to list of input variables before call to "plot_spatial_flux"
- add markdown before call to "plot_sites_timeseries"
- deleted call to set_model_colors (only to allow full flexibility, we can rethink this later)

The idea was to make it clean and easier to track future changes.

**2nd commit:**
- Replace model_filenames and model_labels in the header by a json file
- The json file is just an example, we can discuss name tags and labels. Or each of us can have their local copy and do as they please.
- I felt that it was not a good idea to add colors to the json file (not practical to edit)

**3rd commit:**
- Deleted model_colors from header and added more example experiments
- Following Alex's idea, I have also implemented an alternative function set_model_colors_2 which is expected to work for all configurations. The principle behind it is that in general we are interested in analyzing the outputs of one single model and sometimes we want to look at all models at the same time. In the last case, sometimes we want the colors to be coherent between model. Each person can set their color preferences as they want. The various example experiments allow you to check how the function behaves. In the notebook, the function is called at each cell where "models" is defined.
- If you think this functions works well for you, we can later delete set_model_colors.


mo-aliceramsden (approved):

hdelongueville:
This is brilliant! Thanks for putting this together. It makes the use of the notebook much more fluid.

hdelongueville (approved):

PR #38, opened 2024-09-04
Function to calculate and plot total HFC or PFC in CO2eq

mo-aliceramsden:
- Added new function to plot total HFC or total PFC emissions in CO2 equivalents.

  - This reads in data for the list of provided regions, sums regional fluxes and calculates the root mean square average of the uncertainties from each species.
  - The filenames used to read in data are based on a list of 'standard' filenames, with the prior specified independently for each species.

- Updated plot_country_flux, to also work with total HFCs or total PFCs.

- I've also updated the notebook with some documentation on how to produce these plots.


hdelongueville (commented):
For RHIME (and I think for ELRIS too), all the HFCs standard runs are made with EDGAR.

hdelongueville (commented):
It is not working for RHIME. 
It seems that it can read the file but not find the country emissions.
For example I get
Attempting to read data from rhime_name_edgar
Reading data from: /project/paris/inverse_modelling/RHIME/hfc134a/RHIME_NAME_EUROPE_EDGAR_rhime_obs_rhime_baseline_optimized_hfc134a_yearly.nc
Done!
ERROR: Could not find GERMANY emissions for rhime.

When plotting HFC-134a individually it finds the German emissions so it might come from a bug somewhere..

mo-aliceramsden:
Ah, ok. That's for checking on this. I'll download RHIME's latest netcdf files and see if I can fix that bug.

And I'll add in an if statement too, so if it can't find the default filename (e.g. FLAT for some gases, EDGAR for others), it'll look for the standard 'NAME_EDGAR' file

mo-aliceramsden:
I've just pushed an update which fixes the read in of RHIME data. The issue was with the filenames, as you suspected. Because it was trying to read from the list of standard filenames, which might not exist for all models, it was failing at some points.
The fix for this includes a try/except statement where it first looks for the standard filename, then tries the model_edgar_name filename, then if this fails too, it just skips that species.
The read in of all 3 models' data now works for me, but let me know if you're still having issues though.

Although, I noticed that in some of RHIME's emission netcdfs, the timestamps aren't at the same centre point of the year as the other two models (e.g. RHIME's 2020 timestamp is at 2020-07-01T22:00:00 and ELRIS's is at 2020-07-02T00:00:00). So when empty arrays representing data from missing timestamps are created for RHIME this step fails, but it works for the other models. This failing messes up some of the plotting, adding some random zeros in at timestamps that don't match. So for now, I can put in a fix to replicate the timestamps used by RHIME in this step, but in the future it might be useful to agree on the timestamps used for the middle of each inversion period, incase we run into something like this again.

I've also put in some updates to:

- allow for any number of plots to be created by 'plot_country_flux'. So there's no empty subplots if you only specify one region.
- fixed the x axis for yearly averaged monthly runs (before it had moved the timestamp back to the start of the year)
- changed the plot titles from 'NW_EU2' to 'NW EUROPE' etc. for all 'grouped' areas

melodb (approved):
Everything is working fine from my side. I just fixed the bug on the filenames and took the freedom to change RHIME's prior tag from FLAT to EDGARMINVAL for PFC-116 and CF4 in the portal. I have some minor suggestions but I will only be able to work on them after the conference. In the meanwhile please feel free to merge if you wish.

PR #39, opened 2024-09-26
New initialization function and other edits

melodb:
List of edits to the code:

1) Restructure reading of json files:
- Move reading of json files and definition of standard plotting colors to function initialize_settings
- Make necessary changes in function headers
- This avoids restarting the kernel after changing the files. Re-running the first cell will be enough.

2) Simplify prints when reading all f-gases:
- Add verbose logical flag in read_flux/extract_region_flux and set it to false when reading all F-gases

3) Option to use bigger font size:
- This is useful for producing plots for presentations
- flag 'ppt_mode' is defined in the first cell
- font size and annotate_coords are now defined inside function initialize_settings
- Some adjustments were made in some functions to accommodate bigger fonts
- Necessary changes were made to the functions header

4) Move standard run names to species_info.json:
- Delete option to look for alternative std file

5) Small edits:
- Add option to plot 6 regions in 2x3 subplots
- Add import sys (needed for sys.exit())
- Delete unused variable region_time_out
- Delete old fix for ELRIS
- add 'flux_total_posterior_inversion_grid' to skip_var
- do not plot separate prior nor separate posterior uncertainty when combined = True
- improve format of global legend
- Add "Save plot" cell in the notebook after all plots
- Move print statement in the notebook inside function read_flux_total_fgases

5) Bug fixes:
- Replace ds_all by ds_all.copy()
- Fix bug when plotting prior uncertainty
- Fix bug when plot_inventory == True but inventory_flux == None
- Restrict plotting of inventory between start_date and end_date to avoid partial boxes close to the y-axis
- Fix bug: variable 'c' instead of 'country'

### DIFF
--- a/PARIS_inversion_results.ipynb
+++ b/PARIS_inversion_results.ipynb
@@ -323,7 +323,7 @@
    "outputs": [],
    "source": [
     "fig = func.plot_obs_modelled_together(ds_all_mf_sliced,species,site,\n",
-    "                                    model_colors,s_data,m_data,annotate_coords,\n",
+    "                                    model_colors,s_data,m_data,annotate_coords,ppt_mode,\n",
     "                                    include=['Yapost'],\n",
     "                                    diff_include=['Yapost'],\n",
     "                                    y_lim=None)"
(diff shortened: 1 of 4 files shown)


aliceramsden (approved):
This is all working really well, everything is definitely a lot easier to use now, thanks! 

The 'initialize_settings' function is working well, I can update model_info.json without having to reload the notebook and the ppt_mode option should be really useful.
And thanks for keeping track of all the updates we'd discussed at the last meeting and gettting those changes in.

I just pushed a tiny fix to the print out of missing species, as I realised this wasn't working. Otherwise, everything looks great so I'm happy for this to be merged in straight away

hdelongueville (approved):

PR #41, opened 2024-10-03
Annex plot generator

melodb:
Script to automatically generate draft annex plots.


hdelongueville:
It already looks great! 
I would suggest to remove the run information in the legend? For e.g., InTEM instead of InTEM NAME 33 sites (e+w). 
This information is important internally but I am not sure those details are required in the NIR annexes. (maybe I am wrong)

hdelongueville:
I know we've decided to just show InTEM and Mean Posterior on the 'short period' plot but it seems odd to me not to show ELRIS and RHIME on those plots. This comment can be ignore for now and we can discuss that for next year.

melodb:
> It already looks great! I would suggest to remove the run information in the legend? For e.g., InTEM instead of InTEM NAME 33 sites (e+w). This information is important internally but I am not sure those details are required in the NIR annexes. (maybe I am wrong)

Thanks Helene, you're absolutely right. However, I would prefer not to commit those changes in models_info. To produce the plots, I have manually changed all InTEM longrun labels to "InTEM" but I guess the long labels are useful in our daily work and therefore should stay in models_info. Maybe when producing the annexes, a branch can be created with the necessary changes in models_info.

hdelongueville:
> > It already looks great! I would suggest to remove the run information in the legend? For e.g., InTEM instead of InTEM NAME 33 sites (e+w). This information is important internally but I am not sure those details are required in the NIR annexes. (maybe I am wrong)
> 
> Thanks Helene, you're absolutely right. However, I would prefer not to commit those changes in models_info. To produce the plots, I have manually changed all InTEM longrun labels to "InTEM" but I guess the long labels are useful in our daily work and therefore should stay in models_info. Maybe when producing the annexes, a branch can be created with the necessary changes in models_info.

Yes, you're absolutely right. That sounds like a good solution!

alexdanjou (approved):
Looks good to me!

melodb:
Great work!

Code suggestions (just suggestions, feel free to ignore them):
- Move block of code "Store results for csv and table" to a function to avoid code duplication. We could also think about moving the table functions to a new file and import it in annex_plot_generator to make the file lighter.
- Similarly to function "set_flux_limits", in all spatial map functions, the default of set_fluxlim can be set to "default" instead of "None".
- Function "set_flux_limits" does not fully act on the variables being plotted (e.g., when using the combined option or when plotting the seasonal average of only one month). The limits selection will not be very precise, but the spatial plots look good anyway. I don't have a suggestion that would quickly work for all 3 spatial map functions so I will leave it as is for now.

hdelongueville:
Regarding set_fluxlim flag, you're right. I initially set option to None as a precaution, in case someone intuitively used None. But for consistency, I think I'll switch it to 'default'.

hdelongueville:
Regarding set_flux_limits, see Issue #46. I agree that it's open to further discussion and improvements.

hdelongueville (approved):

hdelongueville:
[celebrate]     Helene De Longueville reacted to your message:
________________________________

PR #48, opened 2024-10-23
Option to plot inversion_grid spatial fluxes

mo-aliceramsden:
The plot_inversion_grid_flux variable can be set to 'True' to plot spatial fluxes using the 'inversion_grid' variables. If the plotting functions cannot find the 'inversion_grid' varaibles, it reverts back to plotting the standard 'flux_total_posterior' variable, which I thought would be useful if we still want to compare current RHIME results with the other models.

UPDATE:
I've also included a first draft of the recalculation of country flux uncertainties during resampling from monthly to annual (or seasonal). This seems to work ok for me, but it would be good if someone could check the maths!


hdelongueville (commented):
Nice job! I have few suggestions:
- Renaming `update_uncert_in_resample` to be bit more descriptive. I find "update" a bit vague and might not be suited on the long term. 
- Creating a dedicated function to calculate the resample uncertainties to make the code easier to read and more flexible on the long term
- Avoiding `tmp` as a variable name for a better long term readability 

I would suggest a dedicated function named `calculate_resample_uncertainty` with the option `resample_uncert_correlation` OR something like that..

I am aware this requires a lot of changes so feel free to ignore.

hdelongueville:
Actually do we want different options to calculate the resample uncertainties? 

In what case would we prefer to use the mean instead of the standard error of the mean?

From my understanding, using the standard error of the mean is always what we want no? In that case, I would remove the mean option. And calculate the standard error of the mean directly when resample = True.

hdelongueville:
Regarding the math behind,

I could be wrong but shouldn't we divide by the square root of the number of periods? `np.sqrt(n_periods)`?

alexdanjou:
> Actually do we want different options to calculate the resample uncertainties?
> 
> In what case would we prefer to use the mean instead of the standard error of the mean?
> 
> From my understanding, using the standard error of the mean is always what we want no? In that case, I would remove the mean option. And calculate the standard error of the mean directly when resample = True.

Don't think that mean is necessary if it has no mathematical justification. I think I'm the one that put the mean, but it's mostly because I didn't think about this...

mo-aliceramsden:
Thanks for having a look through this.

- Good point about renaming `update_uncert_in_resample`, and `tmp`! I can definitely do that.
- And I'll look into moving the uncertainty recalculation to a separate `calculate_resample_uncertainty` function, you're right that'd make the code a lot easier to read/edit (and start the process of clearing up the code, that you opened an issue about!)
- Yes, I think you're right about the square root of the number of periods! When I tried that originally, this recalculation didn't actually change the overall uncertainties so I thought I'd made a mistake in the maths. But I'll trust your maths so I'll add that correction in.
- We might want to keep the original calculation using the mean for calculting the resampled uncertainty as an option for now, as I think this might have to be used for ELRIS in the long term as they do have some correlation between their monthly inversions (if I can remember this correctly)?

hdelongueville:
Yes you are right! The resampled uncertainties are indeed correlated when, for example, 2-year-inversions are used to estimate yearly emissions. So definitely an option we want to keep.

hdelongueville:
Also, I would not trust my math so maybe a third opinion would be nice.

melodb:
> Regarding the math behind,
> 
> I could be wrong but shouldn't we divide by the square root of the number of periods? `np.sqrt(n_periods)`?

I was trying to understand the calculation of the combined standard deviation and I found a good explanation here:
https://stats.stackexchange.com/questions/168971/variance-of-an-average-of-random-variables

If the website is correct, I would say that we should divide `sqrt(sum(var_i))` by `n_periods` or by `np.sqrt(n_periods**2)` (which is equal to `n_periods`). But I'm not the best source on these issues.

aliceramsden:
> > Regarding the math behind,
> > I could be wrong but shouldn't we divide by the square root of the number of periods? `np.sqrt(n_periods)`?
> 
> I was trying to understand the calculation of the combined standard deviation and I found a good explanation here: https://stats.stackexchange.com/questions/168971/variance-of-an-average-of-random-variables
> 
> If the website is correct, I would say that we should divide `sqrt(sum(var_i))` by `n_periods` or by `np.sqrt(n_periods**2)` (which is equal to `n_periods`). But I'm not the best source on these issues.

Ok, this makes more sense to me now, thanks for finding that explaination. When I use `n_periods` instead of `np.sqrt(n_periods)` in the calculation, the output uncertainties reduce during resampling and match some of InTEM's previous results plots, which used different code to calculate the same monthly to annual resampling. Which is also a good sign that we should use `n_periods` without the square root (if we assume that InTEM's older calculations were correct)

aliceramsden:
I've completed a few updates based on all the comments above:
- Moved the resampled uncertainty recalculation to a separate function. For now this just has the new default (recalculate using the square root of the variances) and the old (mean) methods. But should be easy to expand in the future.
- Renamed a few variables: `update_uncert_in_resample` to `resample_uncert_correlation` and `tmp` to `ds_all_original`.
- Updated the maths back to `n_periods` instead of `np.sqrt(n_periods)`

I've made the same updates in the pull request into annex_generator #49

melodb:
Can this PR be closed? I have the feeling these changes got into the code but maybe through a different PR.

aliceramsden:
> Can this PR be closed? I have the feeling these changes got into the code but maybe through a different PR.

Yes, I'm happy for this pull request to be closed, you're right that these changes must have already made it into the code.

PR #49, opened 2024-10-23
Options to plot inversion_grid spatial fluxes in annex plots

mo-aliceramsden:
This pull request includes the same updates as in #48, but also includes edits to annex_plot_generator.py, so that by default spatial fluxes are created from only InTEM and ELRIS, using the inversion_grid flux variables.

UPDATE:
I've included a first draft of the recalculation of country flux uncertainties during resampling from monthly to annual (or seasonal), which is also available in #48 .

### DIFF
--- a/PARIS_inversion_results.py
+++ b/PARIS_inversion_results.py
@@ -3037,7 +3037,6 @@ def plot_spatial_flux_per_timestamp(ds_all,species,plot_area,end_date,s_data,m_d
                                     chop_by='year',dt=1,period_override=None,
                                     plot_site_locations=False,plot_point_markers=False,set_fluxlim='default',
                                     plot_inversion_grid_flux=False):
-                                    plot_site_locations=False,plot_point_markers=False):
     """
     Plots posterior fluxes, prior fluxes or difference between these
     for all models and specific time intervals.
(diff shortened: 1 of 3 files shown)


PARIS_inversion_results.py
```diff
@@ -3240,17 +3345,31 @@ def plot_spatial_flux_per_timestamp(ds_all,species,plot_area,end_date,s_data,m_d
...
+                        print(f'Cannot find inversion_grid variables for {m} so using standard flux output.')
+                        slice_apost   = ds_all[m]['flux_total_posterior'].sel(time=slice(t0_date[m][i],t1_date[m][i]))
+                        slice_apriori = ds_all[m]['flux_total_prior'].sel(time=slice(t0_date[m][i],t1_date[m][i]))
                     var_plot      = np.mean(slice_apost,axis=0) - np.mean(slice_apriori,axis=0)
                     var_plot[np.where(var_plot) == np.nan] = 0.
                 elif var == 'posterior_mean_diff':
-                    mean_slice_apost = np.mean(ds_all[m]['flux_total_posterior'].sel(time=slice(t0_date[m][i],t1_date[m][i])),axis=0)
-                    mean_apost       = np.mean(ds_all[m]['flux_total_posterior'],axis=0)
+                    try:
+                        mean_slice_apost = np.mean(ds_all[m]['flux_total_posterior'].sel(time=slice(t0_date[m][i],t1_date[m][i])),axis=0)
```
melodb, line 3411: I think here it should be [f'flux_total_posterior{var_append}']? I did not commit the fix myself because I'm afraid you're still working on this branch

aliceramsden: Yes you're right, thanks for spotting that! I'll add that fix in now

melodb: From what I see in the code, I think this fix got forgotten. That's why the seasonal plots are at a higher resolution. I will add it now ;)

PR #55, opened 2024-11-12
Joe dca hfo edits

joe-pitt:
Changes to json files for hfo species/run names, plus some hacks for Saurabh's flexinvert files.


melodb (approved):
All good from my side. When restructuring the code, I think we should try our best to avoid model specific settings, even though I know that there will be always model specific correction to do.

PR #57, opened 2024-11-29
Apply rolling_mean to selected models

melodb:
Before, the rolling mean was applied to the combined time series only.
Now, this option can also be used for the individual models. In addition, a list of boolean can be used to specify the models to smooth. This is useful for comparing the smoothed 1-year inversion with the 2-year/3-year inversions.


hdelongueville (approved):
Looks good!

mo-aliceramsden (approved):
I've tested this out with a few species, and the rolling mean option is looking good. 

I noticed I was getting a bug in the global legend, where it was printing mulitple times, so I've fixed the indentation of that part of the code. And the 'inventory' legend label was missing sometimes because the legend labels are set by the last country read in, so if there's missing inventory data for this country, the inventory label is skipped. So I've put a basic fix in for that. Thought I'd get that minor fix in whilst I remembered!

Otherwise, this is all working well for me!

PR #93, opened 2025-04-08
fix_label_plot_sites_timeseries

alexdanjou:
If a model does not use the first site, the label s not plot in plot_sites_timeseries. Fix to prevent that

### DIFF
--- a/fluxy/plots/mf_timeseries.py
+++ b/fluxy/plots/mf_timeseries.py
@@ -297,15 +297,13 @@ def plot_sites_timeseries(
             )
 
         for i, m in enumerate(models):
-            # Define label
-            if iSite == 0:
-                label = model_labels[m]
-            else:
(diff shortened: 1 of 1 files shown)


fluxy/plots/mf_timeseries.py
```diff
@@ -315,6 +313,7 @@ def plot_sites_timeseries(
                     s=2,
                     label=label,
                 )
+                model_labels[m] = None
```
melodb: This works for this function but deletes model_labels[m] for good. Therefore, you will have no label in the subsequent plots.

melodb (changes requested):

melodb:
Thanks for finding this bug! I was having the same issue but I was not understanding why it was not working. Maybe make a copy of model_labels instead of changing it directly

PR #94, opened 2025-04-08
Update flux map functions

hdelongueville:
- [x] Split prepare `flux_perpare_map_data` into `align_map_data` and `combine_map_dataset`. This allows to only apply `align_map_data` if `plot_combined = True` in `plot_flux_map_over_time`. This is needed when comparing models which don't use the same grid.
- [x] Fix the seasons order in `average_over_seasons`, which was in alphabetic order.

NOTE: `flux_combine.py` has now two similar functions `combine_dataset` and `combine_map_dataset`. `combine_dataset`, which is currently flux timeseries specific, should be modified to be flexible enough to be merged with `combine_map_dataset`.


fluxy/operators/flux_combine.py
```diff
@@ -28,3 +28,25 @@ def combine_dataset(
         }
     )
     return {"combined": ds_output}
+
+def combine_map_dataset(
+    ds_all: dict[str, xr.Dataset]
+) -> dict[str, xr.Dataset]:
+    """
```
melodb, line 35: Maybe add a short function description.

melodb (approved):
All working fine on my end!

Issue #96, opened 2025-04-15
Fix x-axis of country fluxes when plotting long timeseries


melodb:
The problem occurred when plotting N2O monthly fluxes from 2010 to 2023.

PR #98, opened 2025-04-16
Fix issue #96

alexdanjou:
If plotting more than 8 years, the xticks will be displayed with a step equal to "time range divided by 8 plus 1".

Closes issue #96


melodb (approved):
Plots looking nice!

Issue #95, opened 2025-04-15
Error when computing uncertainty of country aggregates from covariance matrix


melodb:
This occurred for BENELUX for most HFCs (ELRIS runs).

PR #99, opened 2025-04-16
Fix issue 95

alexdanjou:
Add a selection on dimension "country_2" when exists in extract_region_flux (needed for covariance matrix)
Fix issue #95


melodb (approved):
Working fine, thanks!

PR #104, opened 2025-04-18
Updated mf simulation performance plot

ayanu:
Changed way of how stats are calculated and returned: previously dict of dict of dict -> now: DataFrame
Changed options for calculating stats on prior, posterior and above background simulations 

Changed scatter to barplot. Old scatter plot could be removed. 

Open issue: Plot not usable when presentation_mode True.


fluxy/operators/mf.py
```diff
@@ -67,7 +68,7 @@ def compute_mf_difference(
     return ds_diff
 
 
-def stats_mf(ds_all: dict[str, dict]) -> dict[str, dict]:
+def stats_mf(ds_all: dict[str, dict], type='prior') -> pd.DataFrame:
```
lionel42: I would not name the variable `type`. In python this is reserved for the `type()` function. 
You can either call it `stats_type` or anything else.

fluxy/operators/mf.py
```diff
@@ -78,17 +79,24 @@ def stats_mf(ds_all: dict[str, dict]) -> dict[str, dict]:
...
+        type : 
+            type of statistics to be computed. One of 'prior', 'posterior' for 
+            statistics on the absolute mole fractions and 'prior_above_BC', 
+            'posterior_above_BC' for regional part of mole fraction, i.e. with 
+            BC contribution subtracted from both observation and simulation. 
     Returns:
-        stats (dictionary of dictionaries):
+        stats (pandas.DataFrame):
             Statistical measures, for each site and for each model.
     """
```
lionel42: When returning a data frame, it would be nice to describe what are the columns and what are the indices.

fluxy/operators/mf.py
```diff
@@ -97,21 +105,47 @@ def stats_mf(ds_all: dict[str, dict]) -> dict[str, dict]:
...
+                elif (type=="posterior"):
+                    obs = dc["Yobs"]
+                    sim = dc["Yapost"]
+                elif (type=="prior_above_BC"):
+                    obs = dc["Yobs"]     - dc["YaprioriBC"]                    
+                    sim = dc["Yapriori"] - dc["YaprioriBC"]                    
+                elif (type=="posterior_above_BC"):
+                    obs = dc["Yobs"]   - dc["YapostBC"]
+                    sim = dc["Yapost"] - dc["YapostBC"]
+
```
lionel42: add here else: raise ValueError()

fluxy/operators/mf.py
```diff
@@ -97,21 +105,47 @@ def stats_mf(ds_all: dict[str, dict]) -> dict[str, dict]:
...
+                
+                # additional derived stats
+                tmp['nrmse'] = tmp['rmse'].values/np.mean(obs).item(0)                 
+                
+                # concatenate DataFrames after each iteration 
+                if stats is None:
+                    stats = tmp
+                else: 
+                    stats = pd.concat([stats, tmp], ignore_index=True)
+
```
lionel42: Also include lin 88,89 before the loop. 

Here I would rather store all the `tmp` into a list and then after exiting the loop calling `pd.concat` on the list.

This way you would call `pd.concat` only once.

fluxy/plots/mf_stats.py
```diff
@@ -140,6 +140,103 @@ def plot_stats_mf(
...
+    stats_to_plot: list[str],
+    species: str,
+    model_colors,
+    model_labels,
+    config_data,
+    mf_units_print: str,
+    type: str,
+    stats_ylim=None, 
+    start_date=None,
+    end_date=None,
```
lionel42: add missing type hints

fluxy/plots/mf_stats.py
```diff
@@ -140,6 +140,103 @@ def plot_stats_mf(
...
+
+def plot_stats_pd_mf(
+    stats: pd.DataFrame,
+    stats_to_plot: list[str],
+    species: str,
+    model_colors,
+    model_labels,
+    config_data,
+    mf_units_print: str,
+    type: str,
```
lionel42: same comment with type variable as above

fluxy/plots/mf_stats.py
```diff
@@ -140,6 +140,103 @@ def plot_stats_mf(
...
+        start_date (str) and end_date (str):
+            Dates used to title the plot.
+    Returns:
+        fig (figure):
+            Plot showing each model's fit statistics, for each site.
+    """
+    from fluxy import config
+    
+    import matplotlib.pyplot as plt
+    import numpy as np
```
lionel42: Import statements should be at the top of the file, not inside function

fluxy/plots/mf_stats.py
```diff
@@ -140,6 +140,103 @@ def plot_stats_mf(
...
+
+    models = np.unique(stats['model'].to_numpy())
+    # make sure model_labels remain in correct order
+    model_labels = [model_labels[k] for k in models]
+    
+    colors = [None] * models.size
+    for i, model in enumerate(models):
+        colors[i] = model_colors[model][0]
```
lionel42: use list comprehension for this, as done one the line above with model_labels

fluxy/plots/mf_stats.py
```diff
@@ -140,6 +140,103 @@ def plot_stats_mf(
...
+
+    if type in ['prior', 'posterior']: 
+        mf_str = ""
+    else: 
+        mf_str = " above BC"
+        type = type.split("_")[0]
```
lionel42: A little comment to say what this is doing would be nice

fluxy/plots/mf_stats.py
```diff
@@ -140,6 +140,103 @@ def plot_stats_mf(
...
+        
+    long_stats = pd.melt(stats, id_vars=['model', 'site'], value_vars=stats_to_plot)
+    nrows = len(stats_to_plot)
+    fig, ax = plt.subplots(nrows, 1, figsize=(10, 3 * nrows), tight_layout=True)
+    for i, stat in enumerate(stats_to_plot): 
+        tmp = long_stats[long_stats['variable']==stat].pivot(index='site', columns='model', values='value')
```
lionel42: I don't like the name tmp, I prefer something like `df_this_stats` so that you know it is a dataframe (or a serie ?) and you know what it contains

fluxy/plots/mf_stats.py
```diff
@@ -140,6 +140,103 @@ def plot_stats_mf(
...
+             zorder=3
+            )
+        ax[i].grid(zorder=0)
+        if stats_ylim is not None:
+            if stat in stats_ylim.keys():
+                ax[i].set_ylim(stats_ylim[stat][0], stats_ylim[stat][1])
+        if stat in ['pearson', 'nrmse', 'nn']:
+            ax[i].set_ylabel(config.stat_labels[stat])
+        else:
+            ax[i].set_ylabel(config.stat_labels[stat]+" ("+mf_units_print+")")
```
lionel42: Change logic. Add the thing at the end only if the stat is not in the thing, in order to remove the else statement and to put the set_ylabel oustide of the if else statement

fluxy/plots/mf_stats.py
```diff
@@ -140,6 +140,103 @@ def plot_stats_mf(
...
+    
+    leg = ax[0].legend(ncol=3, borderpad=0.2, columnspacing=1.0, loc="upper center", bbox_to_anchor=(0.5, 1.25), labels=model_labels)
+
+    species_info = config_data["species_info"][species]
+    fig.suptitle(
+        (
+            f'{species_info["species_print"]} {type} model performance versus mole fraction observations{mf_str}'
+        )
+        + f" \n{start_date} to {end_date}"
+    )
```
lionel42: You can concatenate string without the + sign if they are all inside parentheses. just do 
```
( 
   "sting1"
   "string2"
)

**Review by lionel42 — COMMENTED** [review]
I haven't tried to run the code, so there are just minor comments about the code from my side. Hope they are all clear ;)

File: fluxy/config.py [hunk 13]
```diff
@@ -59,8 +59,13 @@
 stat_labels = {
     "pearson": "Pearson correlation coefficient",
     "rmse": "RMSE",
+    "bias": "Bias",
+    "crmse": "Centered RMSE",
+    "sd_sim": "Simulated StDev",
+    "sd_ref": "Observed StDev",
```
melodb: Maybe `"sd_ref"` can be renamed `"sd_obs"`. Moreover, "std" can be renamed `"sd_res"` (residuals) for consistency.

fluxy/operators/mf.py
melodb: Moreover, the options can be already stated in the function definition:

`from typing import Literal`
(...)
`stats_type: Literal["prior", "posterior", "prior_above_BC", "posterior_above_BC"] = "prior",`

fluxy/operators/mf.py
```diff
@@ -78,17 +79,24 @@ def stats_mf(ds_all: dict[str, dict]) -> dict[str, dict]:
...
-
+    type_options = ['prior', 'posterior', 'prior_above_BC', 'posterior_above_BC']
+    assert type in type_options, f"'{type}' is not in {type_options}"
     sites_all = get_unique_sites(ds_all)
 
     # Implemented statistics
-    stat_vars = ["pearson", "rmse", "nrmse", "std"]
-
-    stats = {stat: {site: {} for site in sites_all} for stat in stat_vars}
+    stat_vars = ["pearson", "rmse", "bias", "nrmse", "std", "sd_sim", "sd_ref"]
```
melodb: This list is no longer needed

fluxy/operators/mf.py
```diff
@@ -97,21 +105,47 @@ def stats_mf(ds_all: dict[str, dict]) -> dict[str, dict]:
...
-                for stat in stat_vars:
-                    stats[stat][site][model] = np.nan
-
-        # Delete site from dict if first statistics is NaN
-        if all([np.isnan(v) for v in stats["pearson"][site].values()]) == True:
-            for stat in stat_vars:
-                del stats[stat][site]
-
+                # xarray for single site
+                dc = ds.isel(nsite=site_index).dropna(dim="time")
```
melodb: Maybe `dc` can be renamed `ds_site` to make it more clear?

fluxy/operators/mf.py
```diff
@@ -97,21 +105,47 @@ def stats_mf(ds_all: dict[str, dict]) -> dict[str, dict]:
...
+                d = {
+                    'model': model, 
+                    'site': site, 
+                    'pearson': np.corrcoef(obs, sim)[0, 1], 
+                    'rmse': np.sqrt(np.mean((sim-obs)**2)).item(), 
+                    'crmse': np.sqrt(np.mean((sim-obs-np.mean(sim - obs))**2)).item(), 
+                    'bias': np.mean(sim-obs).item(), 
+                    'sd_sim': np.std(sim).item(), 
+                    'sd_ref': np.std(obs).item(), 
+                    'std': np.std(sim - obs).item(),
```
melodb: In my view, instead of using `.item()` in the middle of the statistics, it would be better to use `.values` when computing `obs` and `sim`

fluxy/operators/mf.py
```diff
@@ -97,21 +105,47 @@ def stats_mf(ds_all: dict[str, dict]) -> dict[str, dict]:
...
+                    sim = dc["Yapost"]
+                elif (type=="prior_above_BC"):
+                    obs = dc["Yobs"]     - dc["YaprioriBC"]                    
+                    sim = dc["Yapriori"] - dc["YaprioriBC"]                    
+                elif (type=="posterior_above_BC"):
+                    obs = dc["Yobs"]   - dc["YapostBC"]
+                    sim = dc["Yapost"] - dc["YapostBC"]
+                    
+                # calculate stats
+                d = {
```
melodb: I would also rename `d` by something else, for example `stats_site`

fluxy/plots/mf_stats.py
```diff
@@ -140,6 +140,103 @@ def plot_stats_mf(
     )
 
     # Print stats to screen
-    print_stats(stats_all.copy(), stats_to_plot)
+#    print_stats(stats_all.copy(), stats_to_plot)
```
melodb: Maybe we can add a flag to print the stats to the screen

melodb (commented):
Thanks a lot for adding these (long overdue) statistics!
I have tried the code for one HFC and the bar plot does not look great with only 4 sites. Maybe that can be improved. I can help addressing the comments if you wish.

fluxy/operators/mf.py
ayanu: Included. Removed test of values of stats_type as this should be covered by Literal.

fluxy/operators/mf.py
ayanu: Done. But since we test for the possible values of stats_type before, this should never happen.

fluxy/operators/mf.py
ayanu: Done

fluxy/operators/mf.py
ayanu: Agreed. Changed.

fluxy/operators/mf.py
ayanu: I come from the time when a line of code was not suppose to have more than 80 characters ... so short variable names were key. But if you insist ...

fluxy/operators/mf.py
lionel42: Yes right, but it might be helpful if someone is adding a new one later on.

fluxy/operators/mf.py
melodb: Please note that the use of `Literal` in the header is only informative. It does not check for consistency. You still need to use `assert`.

fluxy/operators/mf.py
ayanu: Done

fluxy/operators/mf.py
ayanu: Right. Removed.

fluxy/config.py
ayanu: Done

fluxy/plots/mf_stats.py
ayanu: This could just be in the notebook. display() works nicely on a DataFrame

fluxy/operators/mf.py
ayanu: Done. Much nicer now. I obviously did not realise that pd.concat can handle a long list, but I should have known from the inputs I used. So much to learn.

fluxy/plots/mf_stats.py
ayanu: Done

fluxy/plots/mf_stats.py
melodb: The same logic that Lionel proposed below can be applied here:
 
```
mf_str = ""
if stats_type in ['prior_above_BC', 'posterior_above_BC']: 
        mf_str = " above BC"
        stats_type = stats_type.split("_")[0]
```

Moreover, I think it would be cleaner not to re-write `stats_type` but to define a new variable, for example, `stats_str`

fluxy/plots/mf_stats.py
ayanu: Done. But required another local variable, which I usually try to avoid.

fluxy/plots/mf_stats.py
melodb: Here you should not re-write `model_labels` because it is used in other functions.

melodb (commented):
Sorry for the extra comments. I did not have the time to revise the plotting function yesterday. Regarding the width of the bar plots, one possibility is to adapt the length of the figure according to the number of sites to plot and the width of the bars in plot(). But then the title and legend would have to be adjusted as well. This can be improved later.

melodb:
From my side you can delete the old function `plot_stats_mf` and use this name for your function.  Moreover, if you could adapt the notebook accordingly, as well as test `test_plot_stats` inside `tests/test_plots.py` so that the tests would pass that would be great. Otherwise, I can also do it at the end.

lionel42:
@ayanu Would be nice to pass a round of black formatter, maybe @melodb  can show you how to do it

fluxy/plots/mf_stats.py
ayanu: Oh no. Call by reference in python. I don't like that ...

ayanu:
Thanks for all your comments and suggestions. 
Once Daniela has run the formatting wizard, this should be ready to be merged.

ayanu:
In terms of making the plot more usable for less sites, I would suggest to turn it into a horizontal bar plot. Then the height would be determined by the number of sites and not the width. Most likely one will always plot more than one parameter, making the plot wide enough.

melodb:
Thanks for all the changes! I have marked all comments as "Resolved" except two:
- From Lionel: move imports to the top of the file
- From myself: rename variables in config.py ("sd_ref" and "std")

ayanu:
Thanks for keeping track. Imports and stat names should be fixed now.

melodb (approved):
Working perfectly. Thanks again! I will merge it.

Issue #105, opened 2025-04-23
flask modelled mf are not plotted with plot_mf_timeseries


hdelongueville:
When I try plotting mf timeseries for CBW or TOB flask data including `'Yapost' : 'qYapost'`, nothing shows up in the plot for `Yapost` while `qYapost` is being plotted (see figure below).

Is it because the flask data are sparser, and the current plotting settings aren't well-suited for visualizing them? When I am plotting the data simply using `ax.scatter(ds_all_mf_sliced['InTEM_NAME_EDGAR'].time.values, ds_all_mf_sliced['InTEM_NAME_EDGAR']['Yapost'].values)`, it is working fine. 
I had a look at `plot_mf_timeseries` but I haven't spotted anything. Maybe I am missing something obvious...

![Image](https://github.com/user-attachments/assets/27fe2324-3baf-46ba-9958-3ed7fcc0a120)

melodb:
The issue is exactly what you describe. `ax.plot()` is only able to connect points by lines if the dataset is not full of NaN. The time variable has one data point per 4h and the flask data has one per week. The time variable needs to be rearranged for it to work nicely. I had to do it for the CIF inversions (which had daily obs). I will try to adapt to the code.

hdelongueville:
Thanks, that would be great!

melodb:
Solved in PR #134

PR #106, opened 2025-04-25
Read new format

melodb:
**WARNING:** This branch should **NOT** be merged to devel

For use when dealing with data in old and new format. Currently, it reshapes the new format into the old format. I will try to keep it updated with devel.

**In summary:**
- There are two functions in io.py that convert the flux and concentration data into the old format
- Currently, these functions are only called for "CIF-EnKS" and "CIF-4DVAR"
- When selecting mf from one site, one function is called in select.py that cleans the NaN and adds them back in only in data gaps. This solves issue #105
- Percentiles for country fluxes are set to NaN when non existing (needed for CIF-4DVAR)
- dimension country_2 was renamed to country2 to be in agreement with new format
- Small fix in stats_mf to prevent deleting all mf data when one variable is NaN for all timestamps

**Notes for N2O intercomparison with sister projects:**
- The CIF inversion results are already in the JupyterHub in the correct folders.
- **Plotting country fluxes:** all options are working properly. There will be warnings for CIF-4DVAR due to the absence of uncertainties.
- **Plotting mf:** keep `baseline_site = None`; read_model_output is quite slow due to my lousy implementation - consider moving it to other cell in the notebook; available variables in CIF datasets: Yobs, Yapriori, Yapost, YaprioriBC, YapostBC, uYtotal; uYmod exists in both CIF datasets, however, it is all NaN for CIF-EnKS, therefore the code will break. Please refrain from plotting variables that do not exist in all datasets (as uYmod and qYapost). If this becomes too much of a limitation, a fix can be implemented.
- **Plotting mf difference:** not working; bug when "-" exists in the model name (to be fixed). Anyway, it would only compute the difference at exactly equal timestamps.
- **Plotting stats:** working fine
- **Plotting flux maps:** keep `add_sites = False` and `plot_inversion_grid_flux = False`; difference between two flux maps will not work because the grid is not exactly the same; last flux map plot (plot of different timestamps) will only work if there is data in all datasets (2018-2020); do not set `var` to "*_inversion_grid" variables; keep `plot_combined = False` (don't work due to different grids).

**Notebook setup in JupyterHub:**
```
data_dir = '/project/paris/inverse_modelling/'

experiments = {'egu':['ELRIS_NAME_EUROPE_EDGAR_28sites',
                      'ELRIS_FLEXPART_EUROPE_EDGAR_28sites',
                      'InTEM_NAME_EUROPE_EDGAR_26sites_std',
                      'RHIME_NAME_EUROPE_EDGAR_longrun_24sites',
                      'CIF-EnKS_ICON-ART_EUROPE_TNO-AVENGERS',
                      'CIF-4DVAR_FLEXPART_EUROPE_GAINS'],
              }

species = 'n2o'
models = experiments['egu']
period = 'monthly'
mf_units_print = 'ppb'
```

Enjoy!


melodb:
All changes already included in b7b17e6aa105cfc6e1400e3697bdf54c3ddbbce1 and d46ab2f885bc2f0f1018887403582cf5b7b0908d

PR #112, opened 2025-05-05
Bug fix

melodb:
- it was preventing diff map colors of being used

### DIFF
--- a/fluxy/plots/flux_map.py
+++ b/fluxy/plots/flux_map.py
@@ -359,7 +359,7 @@ def plot_flux_map_model_comparison(
         var_plot = get_flux_mean(var_plot, season)
 
         # Determine plot settings
-        is_diff = "diff" in model
+        is_diff = ("diff" in var) or ("diff" in model)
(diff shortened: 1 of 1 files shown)


hdelongueville (commented):
I'm not sure I agree with this change. When we are plotting, for example, `posterior_prior_diff`, don't we want to use cmap_diff for model1, model2 as well as model2-model1?

melodb:
> I'm not sure I agree with this change. When we are plotting, for example, `posterior_prior_diff`, don't we want to use cmap_diff for model1, model2 as well as model2-model1?

Good point!
Should it then be `is_diff = ("diff" in var) or ("diff" in model)`?
Because `is_diff = "diff" in var` does not allow the third plot (which is always a difference plot) to be in diff colors when plotting flux_total_posterior, flux_total_prior, etc.

hdelongueville:
Ah I see your point! Yes, you are right!

hdelongueville (approved):

PR #120, opened 2025-05-12
improve `plot_sites_timeseries`

alexdanjou:
Change the spacing between models so it is dependent of the number of models


hdelongueville:
Sounds good to me! 
Maybe @melodb has an opinion on this commit as she worked on that function?

lionel42:
I notice I did the same in #121  
Can you check if it makes sense ?
Would you be okay to drop this PR, as merging will get tricky ?

alexdanjou:
outdated by #121

PR #121, opened 2025-05-14
Updating new format

### LINKED ISSUE #90: Update fluxy to deal with new format
It would be nice to have this done before EGU.

lionel42:
This PR changed the naming convention of the variables to new format.

I wrote many new tests to fix things where I found issues.

Fixes #90


lionel42:
@melodb  If you can test with your data and check also the actual plot contents that would be nice, as the tests only check the code, but not much the content ;)

fluxy/io.py
```diff
@@ -20,6 +21,35 @@
...
 
 
+# Naming of variables that might have changed in the past
+legacy_names: dict[str, str] = {
+    "country_flux_total_prior": "flux_total_prior_country",
+    "country_flux_total_posterior": "flux_total_posterior_country",
+    "percentile_country_flux_total_prior": "percentile_flux_total_prior_country",
+    "percentile_country_flux_total_posterior": "percentile_flux_total_posterior_country",
+    "nsite": "number_of_identifier",
+    "nsites": "number_of_identifier",
```
melodb: "nsites" can be deleted. I think all old outputs used "nsite" **[fixed in my PR]**

fluxy/io.py
```diff
@@ -20,6 +21,35 @@
...
+    "sitenames": "platform",
+    "Yobs": "mf_observed",
+    "Yapriori": "mf_prior",
+    "Yapost": "mf_posterior",
+    "YapriorBC": "mf_bc_prior",
```
melodb: "YapriorBC" can be deleted. "YaprioriBC" (already there) is the correct naming. **[fixed in my PR]**

fluxy/io.py
```diff
@@ -20,6 +21,35 @@
...
+    "YaprioriBC": "mf_bc_prior",
+    "YapostBC": "mf_bc_posterior",
+    "Yaprior_bias": "mf_bias_prior",
```
melodb: Same here for "Yaprior_bias" (correct one is "Yapriori_bias") **[fixed in my PR]**

fluxy/io.py
```diff
@@ -542,15 +572,25 @@ def edit_vars_and_attributes(
...
+    name_dict = {
+        var: legacy_names[var]
+        for var in itertools.chain(
+            ds.data_vars.keys(), ds.coords.keys(), ds.sizes.keys()
+        )
+        if var in legacy_names
+    }
+
+    ds = ds.rename(name_dict)
+
```
melodb: Maybe a small comment can be added before this block of code **[fixed in my PR]**

fluxy/io.py
```diff
@@ -542,15 +572,25 @@ def edit_vars_and_attributes(
...
         # Easy fix for InTEM ("units" attribute is wrongly set to "unit")
         vars_to_check = [
-            "country_flux_total_prior",
-            "country_flux_total_posterior",
-            "percentile_country_flux_total_prior",
-            "percentile_country_flux_total_posterior",
+            "flux_total_prior_country",
+            "flux_total_posterior_country",
+            "percentile_flux_total_prior_country",
+            "percentile_flux_total_posterior_country",
```
melodb: Maybe this fix for InTEM only (L.589-600) can be moved inside the if-statement `elif m0 == "intem":` (L.636) **[fixed in my PR]**

fluxy/io.py
```diff
@@ -567,6 +607,8 @@ def edit_vars_and_attributes(
             ds = ds.set_index(countrynumber="country").rename(
                 {"countrynumber": "country"}
             )
+
+        if m0 in ["elris", "elris-new"]:
```
melodb: Taking into account that you have renamed all variables before applying the model specific corrections, there is no need to distinguish "elris" from "elris-new" anymore. L.611 (`if m0 in ["elris", "elris-new"]:`) can be deleted and L.607-609 can be replaced by:
```
if "countrynumber" in ds.dims.keys():
    ds = ds.set_index(countrynumber="country").rename({"countrynumber": "country"})
```
The test files "ELRIS-NEW*" then need to be renamed "ELRIS*" and moved to folder "ELRIS" (folder "ELRIS-NEW" can be deleted). **[fixed in my PR]**

fluxy/io.py
```diff
@@ -567,6 +607,8 @@ def edit_vars_and_attributes(
             ds = ds.set_index(countrynumber="country").rename(
                 {"countrynumber": "country"}
             )
+
+        if m0 in ["elris", "elris-new"]:
             var_to_change = "covariance_country_flux_total_posterior"
             if var_to_change in ds and ds[var_to_change].dims == (
                 "time",
```
melodb: This correction must be applied to all datasets that have this variable and not only to ELRIS. Therefore, it must be called immediately after `if file_type == "flux":` or at the end **[fixed in my PR]**

fluxy/io.py
```diff
@@ -654,25 +696,67 @@ def edit_vars_and_attributes(
...
+        ds = ds.assign_coords({var: ds[var] for var in ["number_of_identifier", "time", "platform"]})
+
+    if 'covariance_flux_total_posterior_country' in ds:
+        # Drop for now 
+        ds = ds.drop_vars("covariance_flux_total_posterior_country")
+        logger.warning(
+            "covariance_flux_total_posterior_country is not supported yet. "
+            "The issue is that the duplicated dimension 'country' is not supported in xarray.\n"
+            "Please check the input data."
+        )
```
melodb: Delete this part. This variable is needed. Let's rename the second dimension to "country2" as described in L.612-627. **[fixed in my PR]**

fluxy/operators/convert.py
```diff
@@ -147,7 +147,7 @@ def get_variables(ds_model: xr.Dataset, unit_type: str) -> tuple[list[str], str
             # Particular case of mole fractions:
             if unit_type == "mf":
                 if unit in config.units_scale["mf"].keys():
-                    if var == "sitenames":
+                    if var == "platform":
```
melodb: This correction can be moved to "edit_vars_and_attributes":
`ds["platform"].attrs.pop("units", None)` **[fixed in my PR]**

scripts/PARIS_inversion_results.ipynb
```diff
@@ -321,41 +321,41 @@
...
     "\n",
     "###################################\n",
     "### options for variables to include\n",
-    "# Yobs                - total observed mole fraction\n",
-    "# uYobs_repeatability - observed repeatability mole fraction uncertainty\n",
-    "# uYobs_variability   - observed variability mole fraction uncertainty\n",
-    "# uYmod               - model mole fraction uncertainty\n",
+    "# mf_observed                - total observed mole fraction\n",
+    "# umf_observed_repeatability - observed repeatability mole fraction uncertainty\n",
+    "# umf_observed_variability   - observed variability mole fraction uncertainty\n",
```
melodb: Fix typo: delete "u"

scripts/PARIS_inversion_results.ipynb
```diff
@@ -321,41 +321,41 @@
...
+    "# mf_posterior_bias         - posterior bias added to site\n",
+    "# mf_priorOUTER       - prior mole fractions only from outer regions\n",
+    "# mf_posteriorOUTER         - posterior mole fractions only from outer regions\n",
     "\n",
     "### options for uncertainties to plot as error bars/shaded area\n",
-    "# uYobs_repeatability - observed repeatability mole fraction uncertainty\n",
-    "# uYobs_variability   - observed variability mole fraction uncertainty\n",
-    "# uYmod               - model mole fraction uncertainty\n",
+    "# umf_observed_repeatability - observed repeatability mole fraction uncertainty\n",
+    "# umf_observed_variability   - observed variability mole fraction uncertainty\n",
```
melodb: Fix typo: delete "u"

scripts/PARIS_inversion_results.ipynb
```diff
@@ -321,41 +321,41 @@
...
+    "# stdev_mf_model               - model mole fraction uncertainty\n",
     "# uYtotal             - total mole fraction uncertainty\n",
-    "# qYapriori           - prior mole fraction uncertainty\n",
-    "# qYapost             - posterior mole fraction uncertainty\n",
+    "# qmf_prior           - prior mole fraction uncertainty\n",
+    "# qmf_posterior             - posterior mole fraction uncertainty\n",
```
melodb: Fix typo: replace "qmf" by "percentile_mf"

fluxy/operators/regions.py
```diff
@@ -115,15 +115,15 @@ def extract_region_flux(
...
+                var_percentile = f"percentile_flux_total_{v}_country"
                 if var_percentile in ds_region.variables:
                     da = ds_region[var_percentile]
                     ds_region[f"{v}_lower"] = da.isel(percentile=min_percentile_index)
                     ds_region[f"{v}_upper"] = da.isel(percentile=max_percentile_index)
                 else:
-                    da = ds_region[f"country_flux_total_{v}"]
+                    da = ds_region[f"flux_total_{v}_country"]
                     ds_region[f"{v}_lower"] = da
                     ds_region[f"{v}_upper"] = da
```
melodb: An elif-statement must be added to check for variables "stdev_flux_total_prior_country" and "stdev_flux_total_posterior_country". In this case, variables "*lower" and "*upper" are computed from subtracting/summing the stdev variables to `ds_region[v]`

fluxy/operators/select.py
```diff
@@ -176,7 +186,6 @@ def slice_mf(
                 ds_all.pop(m)
                 continue
         else:
-            ds_all[m] = ds_all[m].sel(time=slice(start_date, end_date))
 
             if len(ds_all[m]["time"]) == 0:
                 logger.warning(f"No {m} obs found between {start_date} and {end_date}.")
```
melodb: Maybe there is no need to nest this if-statement. It can be moved after masking the time (L.167-168).

fluxy/plots/mf_timeseries.py
```diff
@@ -286,11 +294,16 @@ def plot_sites_timeseries(
     # Create figure
     fig, ax = plt.subplots(1, 1, figsize=(0.7 * len(siteList), 8))
 
-    for iSite, site in enumerate(siteList):
-        if iSite != 0:
+    assert margin < 0.5, "Margin must be smaller than 0.5"
+    assert margin > 0, "Margin must be positive"
+
+    model_offset = (1 - 2 * margin) / (len(models) - 1)
+    print(model_offset)
```
melodb: I guess this print statement can be deleted.

fluxy/plots/utils.py
```diff
@@ -322,6 +322,7 @@ def add_site_markers(ax, site_info, color):
 def get_sites_coordinates(
     ds_all: dict[xr.Dataset],
     config_data: dict,
+    fallback_sites: list[str]  | None = None,
```
melodb, line 325: Add description of fallback_sites to arguments description

fluxy/operators/mf.py
```diff
@@ -36,34 +36,72 @@ def compute_mf_difference(
...
+    # Get the common index
+    common_index: pd.MultiIndex = ds_left_index.intersection(ds_right_index)
+
+    # Get the common time and site indices
+    get_common_index = lambda ds, indices: ds.assign_coords(
+        xr.Coordinates.from_pandas_multiindex(
+            indices,
+            "index",
+        )
+    ).sel(index=common_index)
```
melodb, line 64: Should we pass `common_index` as an input for clarity?

fluxy/operators/mf.py
```diff
@@ -36,34 +36,72 @@ def compute_mf_difference(
...
+    ds_diff[key_name] = xr.Dataset(
+        coords={
+            "time": ("index", common_index.get_level_values("time").values),
+            "platform": ("platform", unique_platforms),
+            "number_of_identifier": ("index", platform_indices),
+        },
+        attrs={
+            "description": f"Difference between {model_left} and {model_right}",
+        },
+    )
```
melodb, line 84: In my humble opinion, the code would be simpler to follow if we would first re-structure the two datasets from index to time and platform and then use `xr.align`.

fluxy/operators/mf.py
```diff
@@ -117,50 +155,53 @@ def stats_mf(
...
+            elif stats_type == "prior_above_BC":
+                obs = ds_site["mf_observed"] - ds_site["mf_bc_prior"]
+                sim = ds_site["mf_prior"] - ds_site["mf_bc_prior"]
+            elif stats_type == "posterior_above_BC":
+                obs = ds_site["mf_observed"] - ds_site["mf_bc_posterior"]
+                sim = ds_site["mf_posterior"] - ds_site["mf_bc_posterior"]
+            else:
+                raise ValueError()
+            
+            obs, sim = obs.values, sim.values
```
melodb, line 183: So elegant! <3

fluxy/plots/mf_timeseries.py
melodb, line 1: L.165 needs to be adapted to new naming convention. For example:
`if unc_var.split("_")[0] == "percentile":` **[fixed in my PR]**

fluxy/operators/mf.py
```diff
@@ -36,34 +36,72 @@ def compute_mf_difference(
...
+    ds_right = ds_all[model_right]
+    # Get the common time and site indices
+    make_index = lambda ds: pd.MultiIndex.from_arrays(
+        [ds["time"].values, ds["platform"].values[ds["number_of_identifier"].values]],
+        names=["time", "platform"],
     )
+    ds_left_index = make_index(ds_left)
+    ds_right_index = make_index(ds_right)
+    # Get the common index
+    common_index: pd.MultiIndex = ds_left_index.intersection(ds_right_index)
```
melodb, line 56: I'm getting an empty plot when running `compute_mf_difference` and `plot_mf_timeseries`. From what I checked, `common_index` is empty even though `ds_left_index` and `ds_right_index` have common items. Not sure what's going wrong in intersection().

fluxy/operators/mf.py
```diff
@@ -117,50 +155,53 @@ def stats_mf(
...
+                "site": site,
+                "pearson": np.corrcoef(obs, sim)[0, 1],
+                "rmse": np.sqrt(np.mean((sim - obs) ** 2)),
+                "crmse": np.sqrt(np.mean((sim - obs - np.mean(sim - obs)) ** 2)),
+                "bias": np.mean(sim - obs),
+                "sd_sim": np.std(sim),
+                "sd_obs": np.std(obs),
+                "sd_res": np.std(sim - obs),
+                "nn": np.size(sim),
+            }
```
melodb, line 197: I'm getting only NaN after this step because there are still NaNs in the timeseries after `ds_site = ds.where(mask_site, drop=True)` (L.163). Applying `.dropna("index")` solves the issue.

melodb (commented):
Thanks a lot for working on this! I have highlighted some parts which are not yet working. While testing, I fixed some things myself (comments with "[fixed in my PR]"). You might want to check my PR and potentially merge it before working on my comments.

fluxy/operators/select.py
```diff
@@ -148,22 +157,23 @@ def slice_mf(
...
-                - ds_all[m].time.values[0].astype("datetime64[h]")
+                ds_all[m]['time'].values[1].astype("datetime64[h]")
+                - ds_all[m]['time'].values[0].astype("datetime64[h]")
             ).astype(int)
 
         # Round time to seconds (for consistency between models)
         ds_all[m]["time"] = ds_all[m]["time"].dt.round("s")
 
+        mask_time = (ds_all[m]['time'] >= start_date) & (ds_all[m]['time'] <= end_date)
+        ds_all[m] = ds_all[m].where(mask_time, drop=True)
```
melodb: When slicing the data, one must also mask the data according to variable assimilation_flag (`mask = ds_all[m]["assimilation_flag"] == 1`). This won't make a difference for our results (assimilation_flag is always equal to 1), but is important for other users which output both assimilated and non assimilated observations (e.g. Joël).

fluxy/operators/regions.py
lionel42: So you mean we should compute those 2 varialbes (upper and lower) if they do not exist in the ds ? @melodb

fluxy/operators/select.py
lionel42: I refactored now the whole thing around, have a look should be fixed: ee9669401829cd60b8fbced2cee3c05090a99a4e

fluxy/operators/mf.py
lionel42: I actually think it would make it less clear, because then it means you can kind of choose what the common index is, but in this context we want to enforce the common index to be this one. (It is not a function made for general use, but more for local context)

fluxy/operators/mf.py
lionel42: Right, it feels like I almost reimplemented the align function. 

Not sure if this should be let for a future refactoring or for now, I feel the align function might make some trouble as well.

fluxy/operators/mf.py
lionel42: This is wierd, I might need to check. Can you have a look and make sure the dates are the same values (maybe rounding errors ?) and that the platforms are also named the sames ?

fluxy/operators/mf.py
lionel42: I am wondering where the nan values are comming from. Are you sure there are not in the dataset before this function ?

fluxy/operators/select.py
lionel42: This variable seems to not be in the old format

We have two possiblities 
- add it when reading the file
- slicing only if the variable is in the dataset

This might be important whether or not you want to use this elsewhere

fluxy/operators/select.py
melodb: Looks perfect! I have just updated the comments in [d33849b](https://github.com/openghg/fluxy/pull/121/commits/d33849b7218071cf61efb905588dfcfc1138f039)

fluxy/operators/regions.py
melodb: Yes. Variables "prior/posterior_lower/upper" do not exist in any dataset but their values are stored in index [0] and [1] of variables "percentile_flux_total_prior/posterior_country". These "percentile*" variables exist in both the old and new format. However, in the new format there is also the option to output the variables "stdev_flux_total_prior/posterior_country" instead of "percentile*". In this case, variables "prior/posterior_lower/upper" can be computed from subtracting/summing the "stdev*" variables to `ds_region[v]`. I can take care of this.

fluxy/operators/mf.py
melodb: I found the problem. In some output files, the platform is a char array instead of a str. Therefore ds_left_index and ds_right_index were not exactly the same. There were some conversions to str in some parts of the code but I'm now ensuring that it is a string directly in io.py (see c96aa99ce4046bb2133564d5c152364d3a411e74 and 921263ad0492d94a64e5f6877711c72b5963c82f). This should no longer be a problem with the new format.

fluxy/operators/mf.py
melodb: The NaN values are already in the dataset. That's common in the old format because we store data at a 4h time step and sometimes there is no data for a given site and specific timestamps. In the new format, whish is index based, there is no need to store NaNs. Nevertheless, in the template of the new format we accept NaN as missing values so they may appear in some datasets. Honestly, I like NaNs in the dataset because they allow us to identify data gaps (which should not be connected by a line in a line plot) so I would prefer to add `.dropna("index")` here instead of deleting all NaNs from the timeseries.

fluxy/operators/select.py
melodb: Both are fine for me. First option is the safest one and the second is the easier one (I'm not expecting this variable to be used elsewhere). In the template of the new format, this variable is marked as mandatory, but I have already received datasets without it. Regardless of our choice , we will be assuming that all observations in the dataset were assimilated if `assimilation_flag` does not exist, which I think is ok.

lionel42: Good, i added the variable if not in the dataset in cd8184de8afdea1987828aa64643136250c49aaa and removed the non assimilated data in b291ed7e8b396016fe33a10a7976d75060776ab6

fluxy/operators/regions.py
lionel42: Sounds good if you take care of this because I feel even more lost 😅

lionel42:
@melodb I am (or at least i think I am) done with this, anything to do before we merge ? (I think you might need to fix that prior posterior stdev thing)

fluxy/operators/regions.py
melodb: Sorry, once I decided to do it myself I did not make an effort to be very clear. Hopefully, it is clear from commit [4e3a314](https://github.com/openghg/fluxy/pull/128/commits/4e3a314c17ed9d4e9d1486a6759111e5482d4118)

melodb (approved):
Ready to merge from my side! Thanks a lot for this!

fluxy/config.py
```diff
@@ -16,34 +17,34 @@
...
-    "Yapost_bias": "posterior bias",
-    "YaprioriOUTER": "prior outer region mf",
-    "YapostOUTER": "posterior outer region mf",
-    "Yobs": "observed",
-    "uYobs_repeatability": "obs repeatability mf uncertainty",
-    "uYobs_variability": "obs variability mf uncertainty",
-    "uYmod": "model uncertainty",
-    "uYtotal": "total uncertainty",
+    "mf_bc_posterior": "posterior baseline",
+    "mf_bias_posterior": "posterior bias",
```
hdelongueville, line 24: It would be nice to also add "mf_bias_prior" :)

fluxy/config.py
```diff
@@ -16,34 +17,34 @@
 bel_pop_r = np.round(np.mean(bel_pop / (bel_pop + lux_pop)), 3)
 
 mf_labels = {
-    "Yapriori": "prior",
-    "Yapost": "posterior",
+    "mf_prior": "prior",
+    "mf_posterior": "posterior",
     "YaprioriBC": "prior baseline",
```
hdelongueville, line 22: not mf_bc_prior?

fluxy/config.py
```diff
@@ -16,34 +17,34 @@
...
 
 mf_color_index = {
-    "Yapriori": 1,
-    "Yapost": 0,
+    "mf_prior": 1,
+    "mf_posterior": 0,
     "YaprioriBC": 1,
-    "YapostBC": 0,
+    "mf_bc_posterior": 0,
     "Yapriori_bias": 1,
```
hdelongueville, line 39: not mf_bias_posterior? (same question for YaprioriBC)

fluxy/plots/utils.py
hdelongueville, line 325: Nice option! This fallback_sites is missing from flux_map functions.

scripts/PARIS_inversion_results.ipynb
```diff
@@ -321,41 +321,41 @@
...
-    "# uYobs_repeatability - observed repeatability mole fraction uncertainty\n",
-    "# uYobs_variability   - observed variability mole fraction uncertainty\n",
-    "# uYmod               - model mole fraction uncertainty\n",
-    "# uYtotal             - total mole fraction uncertainty\n",
-    "# qYapriori           - prior mole fraction uncertainty\n",
-    "# qYapost             - posterior mole fraction uncertainty\n",
+    "# mf_observed_repeatability - observed repeatability mole fraction uncertainty\n",
+    "# mf_observed_variability   - observed variability mole fraction uncertainty\n",
+    "# stdev_mf_model            - model mole fraction uncertainty\n",
+    "# uYtotal                   - total mole fraction uncertainty\n",
```
hdelongueville, line 356: shouldn't uYtotal also change?

fluxy/test_utils/models.py
hdelongueville, line 1: shouldn't test_utils go in fluxy/tests/?

hdelongueville (commented):
Thanks a lot for all the work!

scripts/PARIS_inversion_results.ipynb
melodb: If I have fix this in my latest PR (see [854e222](https://github.com/openghg/fluxy/pull/131/commits/854e222ad5ddec583b694d3ea8188add832a4d2f))

fluxy/test_utils/models.py
lionel42: In fluxy/test we put the scripts for unit tests, test_utils is being part of the fluxy package (fluxy/fluxy/test_utils), so that we can import the things useful for testing.
If we put test_utils in fluxy/test, the testing framework will recognise it as a file containing unit test functions, which could be annoying

PR #130, opened 2025-06-04
Add last year to xticks

melodb:
This part of the code is only used when the timeseries has more than 8 years.
This fix adds the last year to the plot x-axis.


alexdanjou (commented):
Good spot! Indeed if we plot 2012 to 2024 data, it will stop in 2022. But I think it is an issue only when the diff between the max and min dates is a multiple of step (eg if for 2012 to 2025 plot, the last tick is 2024 and I don't think adding 2025 is good in that case).

So I'm suggesting to add a condition before this addition : if (max_x-min_x)%np.timedelta64(step, "Y")==0:

melodb:
@alexdanjou, please approve the PR if you think it can be merged. Thanks!

alexdanjou (approved):

PR #131, opened 2025-06-04
Add corrections for CIF-EnKS and other fixes

melodb:
Corrections for CIF-EnKS:
- rename "country2" to "country_2"
- rename platform names from "mhd_c" to "MHD" and "mhd_f" to "MHD_F"
- fix country flux timestamps

Other fixes:
- adapt calculation of sigma_prior to use either percentile, either stdev
- replace dictionary key separator from "-" to "--" so that it works when "-" exists in the model name
- fix var name in notebook


fluxy/operators/regions.py
```diff
@@ -85,17 +85,26 @@ def extract_region_flux(
...
+                    ).sum(dim="country")
+                )
+            elif "stdev_flux_total_prior_country" in ds.variables:
+                ds_region["sigma_prior"] = np.sqrt(
+                    ((ds["stdev_flux_total_prior_country"]) ** 2).sum(dim="country")
+                )
+            else:
+                ds_region["sigma_prior"] = xr.zeros_like(
+                    ds_region["flux_total_prior_country"]
                 ).sum(dim="country")
```
lionel42, line 107: What happens if someone specifies both variables ? Do you want to add a check for that ? that only one is accepted ?

lionel42 (approved):

fluxy/operators/regions.py
melodb: In my view, the modeler is free to specify both variables (they should contain similar information). For plotting, I would give priority to percentile and if it's not found we use stdev

PR #134, opened 2025-06-05
Fixing timeranges in timeseries plots

lionel42:
In the timeseries plot, we sometimes have gaps in the measurements. 

This PR attempts to fix which gaps are shown and which are not.

With the old format this was easy, becauase we could resample the time axis to the freqency we want.
However in the new format this is more tricky.

We can do this behaviour by adding nan values between gaps. so that they don't show on the plot

### DIFF
--- a/fluxy/operators/select.py
+++ b/fluxy/operators/select.py
@@ -347,7 +347,7 @@ def clean_timeseries_missing_data(
     )
     new_times = []
 
-    wrong_times = time[np.where(dtime > min_freq)[0] + 1]
+    wrong_times = time[np.where(dtime > min_freq)[0]]
     new_times = wrong_times + min_freq
 
     max_index = len(ds.index.values)
--- a/fluxy/operators/select.py
+++ b/fluxy/operators/select.py
@@ -1,9 +1,11 @@
+import logging
+import os
+from datetime import timedelta
+from pathlib import Path
+
+import numpy as np
 import pandas as pd
 import xarray as xr
-import numpy as np
-import os
-from pathlib import Path
-import logging
 
 from fluxy.operators.convert import scale_variables
 
@@ -177,15 +179,12 @@ def slice_mf(
 
         # Slice data according to site
         if site is not None:
-            site_index = get_site_index(ds_all[m], site)
-
-            if site_index is None:
-                logger.warning(f"No {m} obs found for {site}.")
+            try:
+                ds_all[m] = slice_site(ds_all[m], site)
+            except ValueError as e:
+                logger.warning(f"Error slicing site {site} from {m}: {e}")
                 ds_all.pop(m)
                 continue
-
-            mask_site = ds_all[m]["number_of_identifier"] == site_index
-            ds_all[m] = ds_all[m].where(mask_site, drop=True)
 
         if len(ds_all[m]["time"]) == 0:
             # Remove model if no data left after time slicing
@@ -215,6 +214,31 @@ def slice_mf(
             ds_all[m] = ds_all[m].sel(time=both_times)
 
     return ds_all
+
+
+def slice_site(ds: xr.Dataset, site: str) -> xr.Dataset:
+    """
+    Slices the dataset to only include data for a given site.
+
+    Args:
+        ds (xarray dataset):
+            Dataset with mf data of a given model.
+        site (str):
+            Site of interest.
+    Returns:
+        ds (xarray dataset):
+            Dataset with mf data of a given model, sliced to only include data for the given site.
+    """
+
+    site_index = get_site_index(ds, site)
+
+    if site_index is None:
+        raise ValueError(f"Site {site} not found in dataset.")
+
+    mask = ds["number_of_identifier"] == site_index
+    ds = ds.where(mask, drop=True)
+
+    return ds
 
 
 def get_site_index(ds: xr.Dataset, site: str) -> int | None:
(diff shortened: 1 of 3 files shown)


lionel42:
@melodb I tried to do something similar to #106 , can you check and see if it works well for you ?

I had to move the function after the site selection, otherwise we might have issues if we apply this method on mulitple sites, i hope it is no problems

fluxy/operators/select.py
```diff
@@ -258,3 +282,80 @@ def get_unique_sites(ds_all: dict[str, xr.Dataset]) -> list[str]:
...
+
+    if min_freq is None:
+        # Calcuate a minimum frequency based on the median time difference
+        # between data points, assuming that the median is a good representation
+        # of the time difference between data points.
+        # 2x is a good value for data gaps, 1.9x is used to avoid approximation errors.
+        min_freq = 1.9 * dt_median
+    else:
+        # Comvert timedelta to numpy timedelta64 for consistency
+        min_freq = np.timedelta64(min_freq).astype(dtime.dtype)
```
melodb, line 336: If the data has a frequency of 1h, the user would set `min_freq = '1H'`. In this case, shouldn't we set `min_freq = 1.9 * min_freq`? Otherwise, if the data points are distanced by 1.1H, this will be considered a data gap. In the way I see it, we would consider it a data gap only if they are distanced by 2H (or 1.9H).

fluxy/operators/select.py
```diff
@@ -258,3 +282,80 @@ def get_unique_sites(ds_all: dict[str, xr.Dataset]) -> list[str]:
...
+
+    # Check for data gaps
+    if np.all(dtime <= min_freq):
+        return ds
+
+    logger.info(
+        f"Adding NaN between data gaps using dt={dt_median.astype('timedelta64[h]')}."
```
melodb: Maybe here we should output `dt_median` when `min_freq = None` and `min_freq` when `min_freq != None` (should we rename the input variable `min_freq` equal to `dt_median`?)

fluxy/operators/select.py
```diff
@@ -258,3 +282,80 @@ def get_unique_sites(ds_all: dict[str, xr.Dataset]) -> list[str]:
...
+    )
+    new_times = []
+
+    wrong_times = time[np.where(dtime > min_freq)[0] + 1]
+    new_times = wrong_times + min_freq
```
melodb, line 348: I guess here, you would sum `dt_median` if `min_freq = None` when entering the function, right?

fluxy/plots/mf_timeseries.py
```diff
@@ -114,14 +123,22 @@ def plot_mf_timeseries(
...
             model_color = model_colors[mdiff0]
 
+        ds_plot = ds_all[m]
+        # Select only the site to plot
+        ds_plot = slice_site(ds_plot, site)
+        # Clean the time dimension
+        ds_plot = clean_timeseries_missing_data(
+            ds_plot, variables_nans=vars_to_plot, min_freq=time_freq_min
+        )
+
```
melodb: Why do we need to call `slice_site` in the plotting function? Wasn't it already called inside `slice_mf`? Moreover, why don't we call `clean_timeseries_missing_data` inside `slice_mf`? As far as I understood, we wanted to separate the plotting functions from the selection/processing functions.

melodb: I see that you use the `vars_to_plot` as input. However, I think that's not needed. The data gaps and NaNs will be the same for all variables: if the obs are NaN, the prior mf, posterior mf, BC mf, bias mf, etc, will all be NaN. So you can apply the cleaning to all mf variables based on `mf_observed`.

fluxy/operators/select.py
```diff
@@ -258,3 +282,80 @@ def get_unique_sites(ds_all: dict[str, xr.Dataset]) -> list[str]:
...
+    # Check for data gaps
+    if np.all(dtime <= min_freq):
+        return ds
+
+    logger.info(
+        f"Adding NaN between data gaps using dt={dt_median.astype('timedelta64[h]')}."
+    )
+    new_times = []
+
+    wrong_times = time[np.where(dtime > min_freq)[0] + 1]
```
melodb: Are you sure the "+1" is needed? I only get the right plots when I replace `min_freq` by `dt_median` in L.344 (see comment above) and if I delete +1.

With +1 (current version):
image

Without +1 (my suggestion):
image

melodb: Sometimes there are isolated data points (between two data gaps) - see for instance March 2022. You can see that those points do not show up in the line plot (you can only see the error bar). They would only show up in a scatter plot. I would ignore this for now. Those isolated points are rare and I would keep the line plot anyway because it improves visualization.

melodb (commented):
Thanks a lot for this! I have added some comments.

fluxy/operators/select.py
lionel42: I think not. If you specify '1H' and the frequency is exactly 1 hour the line will be shown. I think if you let the user decide, the user could specify 1.9 * timedelta(hour=1) if they want this behaviour.

fluxy/operators/select.py
lionel42: I have added a more meaningful log message in  f58c4365d32a641890b76aa90d243505d1357c6b, what do you think ?

fluxy/operators/select.py
lionel42: Not sure what you mean exactly, here the goal is to create the timestamps where the none value should be given to break the lines

fluxy/plots/mf_timeseries.py
lionel42: There is a huge logic issue if you add nan values in time with the new 'index' axis, because now the index axis is containing all stations. So if you add nan values to break the lines, you would need to add them to all stations separately, which makes things more complex. 

Also the function `plot_mf_timeseries` should also work if you give it a dataset where more than one site is present (at least in my opinion), therefore i decided to add the slice_site in there, even though it might be applied earlier in other processing steps.

For the `vars_to_plot` I would keep it. I am not sure if that makes sense with whole inversion concept, but it might be in the future that other people come with other datasets where maybe observations and simulations are given for different timesteps and are not always given together.

fluxy/operators/select.py
lionel42: Yes, right the +1 was a mistake

lionel42: We can maybe add a scatter plot of the points to show the single points ?

lionel42: image
 somthing like that ?

melodb: I guess you have just added markers in the whole line plot, right? I think that's the best way for now. From my quick chat with ChatGPT, the only way to add markers only to the isolated points would be to detect them and plot them separately. But I don't think it's worth the work.

lionel42: Yes exactly, I commited the marker thing . I made them small so that they should not look larger than the lines

fluxy/operators/select.py
melodb: So I guess we can delete the multiplication by 1.9, right? If the frequency is exactly 1h and we don't specify `min_freq`, we would obtain `dt_median = 1h`. I guess we should then assign `min_freq = dt_median`. My point is that the code should be similar for an equally spaced data set regardless if we specify `min_freq` or not .

fluxy/plots/mf_timeseries.py
melodb: Currently, `plot_mf_timeseries` is not ready to plot several sites. There is no for-loop over the sites and you can see that variable site is only used to print the y-axis label. There was the idea of expanding the function so that it could plot the timeseries of several stations in the same plot. However, this new development was not considered a priority.

Currently in devel, in order to plot the timeseries, you need to slice the data (run `slice_mf`) with `site != None`. The option to slice the data with `site=None` was made for the stats calculation only. With your implementation, the user is able to slice/scale the data using `site=None`, and only in the plotting phase decide on the site to plot (@alexdanjou will like this). However, it breaks a bit the idea of separating the plotting functions from the processing functions which underlines the whole code. Maybe I'm being too picky on this, but in my view, we should try to be faithful to the structure already in place.

Therefore, my suggestion is to call `clean_timeseries_missing_data` inside `slice_site` (I think we always want to clean the timeseries after slicing, so there is no need to make these two steps independent) and to call `slice_site` in the notebook cell before calling `plot_mf_timeseries` (the user already needs to call the reading function anyway and potentially `slice_mf` if they want to decide on the time window and units).

Regarding your point on `vars_to_plot`, which could be an argument to keep the call to `slice_site` inside `plot_mf_timeseries`, I really don't see the need to keep it. We can only simulate mole fractions for which there are observations. There is indeed the option to add to the dataset observations that don't have associated simulated mf (that's what `assimilation_flag` is used for). But the other way around is impossible. I don't see the point of preparing the code for something that will never happen.

Sorry for the long text. If I'm being too picky regarding the structure already in place, just leave it as is.

fluxy/operators/select.py
lionel42: Yes I guess we can, because median will alread take this into account

fluxy/plots/mf_timeseries.py
lionel42: * for the site, I agree, I will just add a check that we have one site only
* clean stays inside, as it is kind of 'hack' of the data, only used for the plot
* vars_to_plot we keep, because maybe we have some simulation and no observations or the opposite

lionel42: Should be good now

fluxy/operators/select.py
```diff
@@ -327,10 +327,9 @@ def clean_timeseries_missing_data(
         # Calcuate a minimum frequency based on the median time difference
         # between data points, assuming that the median is a good representation
         # of the time difference between data points.
-        # 2x is a good value for data gaps, 1.9x is used to avoid approximation errors.
         dt_median = np.median(dtime)
-        min_freq = 1.9 * dt_median
-        logger.info(f"Using median time difference 1.9 * {dt_median=} as min_freq.")
+        min_freq = dt_median
```
melodb: Can we get rid of L.330 and directly set `min_freq = np.median(dtime)`?

fluxy/operators/select.py
```diff
@@ -258,3 +285,82 @@ def get_unique_sites(ds_all: dict[str, xr.Dataset]) -> list[str]:
...
+    else:
+        # Comvert timedelta to numpy timedelta64 for consistency
+        min_freq = np.timedelta64(min_freq).astype(dtime.dtype)
+
+    # Check for data gaps
+    if np.all(dtime <= min_freq):
+        return ds
+
+    logger.info(
+        f"Adding NaN between data gaps using dt={min_freq.astype('timedelta64[h]')} hours."
```
melodb: The word "hours" can be deleted at the end of the sentence. Otherwise it shows up twice.

fluxy/operators/select.py
lionel42: yes ;)

lionel42:
Should be good now :)

melodb (approved):
All working very well! Ready for merging :D

PR #137, opened 2025-06-06
added prior_lower and prior_upper to combined dataset, so combined prior can be plotted

mo-aliceramsden:
Very minor update, just adding two missing lines to `combine_dataset` in flux_combine.py`


melodb (approved):
Thanks for adding this fix!

Issue #85, opened 2025-02-25
CLI


lionel42:
I see that in fluxy.cli (command line interface), nothing was changed.

Is there a plan to tackle this ?

melodb:
This script needs to be operational in October so we must indeed update it. But maybe we can take care of it once the main code is finished.

Issue #123, opened 2025-05-15
combine_map_dataset does not work with some emissions netcdfs


mo-aliceramsden:
plot_combined = True sometimes fails for plot_flux_map_over_time(), if not all emissions datasets contain the same variables.

When combine_dataset or combine_map_dataset functions (from flux_combine.py) are run, xr.concat(ds_list...) sometimes fails, if one datasets contains a variable not present in the others.

This issue could be solved by dropping variables that are only included in one model's netcdfs, before the xr.concat(). Using, for example:

```
for d in list(ds_all.keys()):
    for v in ds_all[d]:
        if '_out' in v or 'percentile_flux_total_posterior_inversion_grid' in v:
            ds_all[d] = ds_all[d].drop_vars(v,errors='ignore')
```

Or, this issue could be solved by dropping all variables, other than those to be used immediately in the plotting function.

hdelongueville:
The function `align_map_data` prepares the flux datasets for flux maps by filtering variables, removing unused dimensions, and aligning time and spatial coordinates. Another step could be added there to keep only the common variables.

hdelongueville:
OR we add this step in `combine_map_dataset`, which makes more sense I think

hdelongueville:
I'd prefer to remove as least variables as possibles in case we need them in future development. 
Something like this?

```
def combine_map_dataset(ds_all: dict[str, xr.Dataset]) -> dict[str, xr.Dataset]:
    """
    Combine multiple xarray datasets along the 'model' dimension and return the mean dataset.

    Args:
        ds_all: xarray datasets of fluxes indexed by model names.

    Returns:
        A dictionary with a single key 'combined', containing the mean of all datasets along the 'model' dimension.
    """
    models = list(ds_all.keys())
    ds_list = list(ds_all.values())

    # Step 1: Find common variables across all datasets
    common_vars = set(ds_list[0].data_vars)
    for ds in ds_list[1:]:
        common_vars.intersection_update(ds.data_vars)

    # Step 2: Subset each dataset to contain only the common variables
    ds_list = [ds[sorted(common_vars)] for ds in ds_list]

    # Step 3: Combine and average
    ds_dict = {
        "combined": xr.concat(ds_list, dim="model", combine_attrs="override").mean(
            dim="model", keep_attrs=True
        )
    }

    return ds_dict
```

alexdanjou:
[...]

Issue #118, opened 2025-05-08
Units not found when using "all_hfc"


hdelongueville:
When using flux_timeseries with "all_hfc", I get the following error:

```
File /user/work/qq24644/fluxy/fluxy/plots/flux_timeseries.py:230, in plot_country_flux(ds_all, species, plot_regions, config_data, model_colors, model_labels, start_date, end_date, annex_mode, plot_inventory, inventory_years, data_dir, fix_y_axes, add_prior, add_prior_unc, set_global_leg, country_codes_as_titles, plot_separate, plot_combined, resample, resample_uncert_correlation, plot_resample_and_original, return_res, rolling_mean)
    [227] n_cols, n_rows = determine_subplots_arrangement(len(plot_regions))
    [229] if "all" in species:
--> [230]     units = {ds.posterior.units for ds in ds_all.values()}
    [231] else:
    [232]     units = {ds.country_flux_total_posterior.units for ds in ds_all.values()}

File /user/work/qq24644/miniconda3/envs/parisenv/lib/python3.13/site-packages/xarray/core/common.py:305, in AttrAccessMixin.__getattr__(self, name)
    [303]         with suppress(KeyError):
    [304]             return source[name]
--> [305] raise AttributeError(
    [306]     f"{type(self).__name__!r} object has no attribute {name!r}"
    [307] )

AttributeError: 'DataArray' object has no attribute 'units'
```
Somehow it seems that units in `ds_all_flux_scaled['ELRIS']['posterior'].attrs['units']` and `ds_all_flux_scaled['ELRIS']['prior'].attrs['units']` are missing.

I haven’t found the source of the problem yet, but it likely arises when all the HFC datasets are combined.

alexdanjou:
@hdelongueville is the bug still present for you? I don't manage to reproduce it. If it is, can you check if the PR #138 resolves it (and if yes put in the comment of the PR that it resolves it).

hdelongueville:
I don't have that issue anymore, no idea why.

hdelongueville:
Let's say it is fixed then

melodb:
[...]

Issue #116, opened 2025-05-06
chop_by = year/month not working properly


melodb:
In functions average_over_years and average_over_months, the averaging is performed by making the floor division of the year/month by `N` (variable `dt` in the notebook):
https://github.com/openghg/fluxy/blob/8318741922c54e10c0bcac68a86c0db1e5a8fc60/fluxy/operators/flux_map_resample.py#L202

For a time series ranging from 2018-2023, this works nicely for `dt=1` and `dt=2`. For `dt=3`, the floor division aggregates 2018 in one group, 2019-2021 in another, and the remaining years (2022-2023) in the last group (see figure).
![Image](https://github.com/user-attachments/assets/55c9bd23-b3b7-4b78-a349-9e695c191a85)

I think it would make more sense to create the groups in order, for instance 2018-2020 followed by 2021-2023, instead of according to the floor division. The same non expected grouping occurs for dt=4, 5 and 6. Maybe a different approach can be used to define the groups.

hdelongueville:
Great point!

alexdanjou:
PR #138 fixes this I think. Can you verify @melodb ? (I change the line in `groups = (ds.time.dt.year-ds.time.dt.year.min()) // N`).

PR #138, opened 2025-06-06
Update script generating draft annex plots and tables

### LINKED ISSUE #81: `for` loop for slicing : necessary for memory issues. Could we hide it?

alexdanjou:
Closes issues #85 and (as a side effect) #81, #118, #123 and #116.

Stuff done : 
- Restructure `annex_plot_generator.py` : the files are read once and not multiple times. There is still a performance issue (#139 )
- Adapt  `annex_plot_generator.py` to recent changes of fluxy
- Change the order of operations done in the flux map functions to select first the variables to be plotted and then do less unnecessary computing (also solves #123 )
- Fix bug in `select.py` and closes #81


alexdanjou:
You can view the produces files in `project/paris/users/AlexDanjou/images/` on the jupyter hub. Tell me if you spot any issues!
I have also tested the classic notebook and it seems to work fine.

fluxy/cli/annex_plot_generator.py
```diff
@@ -1,912 +1,435 @@
...
-        "pfc218": 0.95,
-        "pfc318": 0.95,
-        "sf6": 0.95,
-    },
-}
-
 ###########################################
 
+def define_model_list(models, species, config_data, json_exp_to_use):
+    print("WARNING : each team should check that the right files are used. Still to be checked by RHIME, ELRIS, InTEM.")
```
lionel42: Use logging module instead of print statements

fluxy/cli/annex_plot_generator.py
```diff
@@ -1,912 +1,435 @@
...
-        "pfc116": 0.95,
-        "pfc218": 0.95,
-        "pfc318": 0.95,
-        "sf6": 0.95,
-    },
-}
-
 ###########################################
 
+def define_model_list(models, species, config_data, json_exp_to_use):
```
lionel42: Document function

fluxy/cli/annex_plot_generator.py
```diff
@@ -1,912 +1,435 @@
...
+        else:
+            model_read = f"{model.split('_')[0]}_{config_data['models_info']['standard_run']['default'][species]}"
+            if "longrun" in model:
+                model_read = f"{model_read}_longrun"
+                
+        models_std.append(model_read)
+        
+    return models_std
+
+def dict_to_str_dataframe(res, inventory_years, species):
```
lionel42: document function

fluxy/cli/annex_plot_generator.py
```diff
@@ -1,912 +1,435 @@
...
+        fig, res_dict = plot_country_flux(ds_all_flux_scaled,
+                        species = species,
+                        model_colors = model_colors,
+                        model_labels = model_labels,
+                        start_date = start_date,
+                        end_date = end_date,
+                        resample = None,
+                        **kwargs_country_flux_general,
+                        **kwargs_country_flux_yearly_species)
+        full_path = os.path.join(output_path, f"{species}_country_flux_annual_longrun_{region}.png")
```
lionel42: Try using pathlib for the paths, then you can simply do 
`fullpath= output_path / filename`

fluxy/cli/annex_plot_generator.py
```diff
@@ -1,912 +1,435 @@
...
-    print("\n\nTABLE HFC\n\n")
+    print("\nTABLE HFC")
     hfc_res = annual_res[
         annual_res.species.apply(lambda x: x[:3].lower() == "hfc")
     ].copy()
     hfc_res["species"] = hfc_res.species.apply(lambda x: x.replace("hfc", "HFC-"))
-    make_table(hfc_res, f"{output_path}/hfc_res_{regions[0]}.tex")
-    hfc_res.to_csv(f"{output_path}/hfc_res_{regions[0]}.csv", index=False)
+    make_table(hfc_res, f"{output_path}/hfc_res_{region}.tex")
+    hfc_res.to_csv(f"{output_path}/hfc_res_{region}.csv", index=False)
```
lionel42: pathlib also in this context would make sense

fluxy/cli/filenames.json
```diff
@@ -0,0 +1,85 @@
+{
```
lionel42, line 1: document what is in this file

fluxy/operators/flux_align_dataset.py
```diff
@@ -12,17 +12,24 @@ def align_time(ds_list: list[xr.Dataset]) -> list[xr.Dataset]:
...
     if any(abs(dtime - np.median(dtime)) > 0.1 * np.median(dtime)):
         raise ValueError("Unable to infer period from dataset")
     period = np.median(dtime)
 
+    # Select the data
+    if np.unique([ds.time.size for ds in ds_list]).size != 1:
+        min_date = max([x.time.min() for x in ds_list]) - period/2
+        max_date = min([x.time.max() for x in ds_list]) + period/2 
+        ds_list = [ds.sel(time = slice(min_date, max_date)) for ds in ds_list]
+
```
lionel42: Not very clear what you are selecting here

fluxy/operators/flux_align_dataset.py
```diff
@@ -83,33 +90,34 @@ def align_lat_lon(
...
-                for var in ds.data_vars
-                if not {"time", "latitude", "longitude"}.issubset(ds[var].dims)
-            ]
-        )
-        # Step 2: Remove unused coordinates (dimensions that are no longer used)
-        unused_dims = set(ds.dims) - set(
-            dim for var_i in ds.data_vars for dim in ds[var_i].dims
-        )
-        ds_all[key] = ds.drop_dims(unused_dims)
+        if type(ds) == xr.Dataset: # Applied only if Dataset and not DataArray
```
lionel42: use isinstance to check type

fluxy/operators/flux_align_dataset.py
```diff
@@ -83,33 +90,34 @@ def align_lat_lon(
     return aligned_ds_list
 
 
-def align_map_data(ds_all: dict[xr.Dataset]) -> dict[xr.Dataset]:
+def align_map_data(ds_all: dict[xr.Dataset | xr.DataArray]) -> dict[xr.Dataset | xr.DataArray]:
     """
```
lionel42: As I understand here, you have simply added that the function can take datarrays as well, but is basically doing nothing when a datarray is given right ?

fluxy/operators/flux_align_dataset.py
```diff
@@ -83,33 +90,34 @@ def align_lat_lon(
...
+            Dictionary of model names and corresponding xarray Datasets/DataArrays.
 
     Returns:
-        dict[xr.Dataset]:
-            Aligned datasets, after removing non-geographic variables.
+        dict[xr.Dataset | xr.DataArray]:
+            Aligned Datasets/DataArrays, after removing non-geographic variables.
     """
 
     for key, ds in ds_all.items():
```
lionel42, line 107: Change the loop logic to 
```
for ...:
    if ds is datarray:
         continue 
    # process dataset
```
This will remove an indentation level which is nice

fluxy/operators/flux_map_diff.py
```diff
@@ -27,19 +27,27 @@ def define_var_plot(
 
     if var == "posterior_prior_diff":
         var_plot = ds["flux_total_posterior"] - ds["flux_total_prior"]
+        var_plot.attrs["units"] = ds['flux_total_posterior'].attrs.get("units")
```
lionel42: Create a function `copy_unit` to remove all the duplicated accessing code here and below.

Maybe also make a unit variable and assign it the the attribute at the end of the if else statement

fluxy/operators/flux_map_resample.py
```diff
@@ -26,17 +26,22 @@ def get_flux_mean(
...
-        dim="time"
-    )
+        if season not in seasonal_means.season.values:
+            raise ValueError(f"Season '{season}' not found in the dataset.")
 
-    if season not in seasonal_means.season.values:
-        raise ValueError(f"Season '{season}' not found in the dataset.")
+        ds_output = seasonal_means.sel(season=season)
+    ds_output.attrs["start_date"] = data.time.values[0]
+    ds_output.attrs["end_date"] = data.time.values[-1]
```
lionel42: Assumes time is sorted, are you sure that is true ?

fluxy/operators/flux_map_resample.py
```diff
@@ -171,7 +176,7 @@ def average_over_seasons(
     ordered_seasons = [s for s in desired_order if s in ds_avg.time.values]
     ds_avg = ds_avg.sel(time=ordered_seasons)
 
-    time_labels = ds_avg.time.values.tolist()
+    time_labels = ["Dec - Feb", "Mar - May", "Jun - Aug", "Sep - Nov"]
```
lionel42: Use arrows instead of `-` ?

fluxy/operators/flux_prepare_inventory.py
```diff
@@ -36,8 +36,8 @@ def retrieve_inventories(
     """
     inventories_list = list()
 
-    if inventory_years is None:
-        inventory_years = [None]
+    if inventory_years is None or not isinstance(inventory_years, list):
```
lionel42: can remove the none check in this case as None is not instance of list

fluxy/operators/flux_timeseries_resample.py
```diff
@@ -113,6 +113,7 @@ def resample_flux(
     ds_all_original = {m: ds_all[m].copy() for m in ds_all.keys()}
 
     ds_all_resampled = calculate_resampled_flux(ds_all, rtime)
+    new_keys = [m + "_resample" if r else m for m, r in zip(ds_all_resampled.keys(),rtime)]
```
lionel42: If this is used only for the return statement, move it below to where it is used

fluxy/operators/select.py
```diff
@@ -11,7 +11,7 @@
 
 
 def slice_flux(
-    ds_all: dict[str, xr.Dataset],
+    ds_all_p: dict[str, xr.Dataset],
```
lionel42: dangerous to change signature like this, also this is not according to convention other functions, I would switch it back

fluxy/operators/select.py
```diff
@@ -49,19 +49,21 @@ def slice_flux(
             xarray datasets, scaled, converted, and sliced between chosen dates.
 
     """
-
+    ds_all = dict()
```
lionel42: give this variable another name instead of changing the signature

fluxy/plots/flux_timeseries.py
```diff
@@ -259,12 +260,20 @@ def plot_country_flux(
...
+            else :
+                end_date_inv = end_date
             inventories_to_plot = retrieve_inventories(
                 data_dir,
                 country,
                 species,
-                start_date,
-                end_date,
+                str(start_date_inv),
+                str(end_date_inv),
```
lionel42: Would make more sense to do this conversion in the retrieve_inventories function

fluxy/plots/flux_timeseries.py
```diff
@@ -259,12 +260,20 @@ def plot_country_flux(
         ax = axes if (n_rows, n_cols) == (1, 1) else axes.flatten()[i]
 
         if plot_inventory:
+            if isinstance(start_date,list):
+                start_date_inv = min([np.datetime64(date) for date in start_date])
+            else :
+                start_date_inv = start_date
+            if isinstance(end_date,list):
+                end_date_inv = min([np.datetime64(date) for date in end_date])
```
lionel42: should be max ?

fluxy/plots/flux_timeseries.py
```diff
@@ -364,13 +373,14 @@ def plot_country_flux(
...
+       
         # set legend if needed
         if not set_global_leg:
-            ncol = 3 if annex_mode else 2
+            ncol = len(ds_to_plot) + 1 if annex_mode else 2
             leg = ax.legend(ncol=ncol, borderpad=0.4, columnspacing=1.0)
-            for l in leg.legend_handles[: (-1 if plot_inventory else None)]:
-                l.set_linewidth(3.0)
+            # for l in leg.legend_handles[: (-1 if plot_inventory else None)]:
+            #     l.set_linewidth(3.0)
```
lionel42: Why commenting this ?

fluxy/plots/utils.py
```diff
@@ -631,10 +631,8 @@ def plot_country_borders(ax, lines, border_color):
 
 
 def set_flux_limits(
-    ds_all: dict[xr.Dataset],
-    var: str,
+    ds_all: dict[xr.DataArray],
```
lionel42: If you move to dataarray, then I would change the name to `da_all`, just make sure the function is not used in other parts of the code than the ones you cahnged

lionel42 (changes requested):
That is quite a big piece of work.
Additionally to the code comments, I have few general comments:

* the code in the cli seems to be very specific to your PARIS use case. It there a plan to change that in the future or do you keep it for PARIS ? 
* Apply the black formatter on your code
* I haven't run the code yet, might do it once you apply some of the changes

fluxy/operators/flux_align_dataset.py
alexdanjou: One dataset can go from 2008 to 2023 and the other from 2018 to 2023. So I'm trying to select the overlapping range.
I've changed "# Select the data" into "# Reduce datasets to their overlapping time range ". Is that clearer?

fluxy/operators/flux_align_dataset.py
alexdanjou: When a dict of datasets is given, the function drops the variables in the datasets that have not "time","latitude" and "longitude" in their dimensions. Then it timely and spatially align datasets in the dict. When there is a datarray in this dict, it just skip the first bit (dropping the unwanted variables) but still align it with the other datasets/dataarrays.

fluxy/operators/flux_map_resample.py
alexdanjou: I assume so. And thought that indexing was quicker than sorting. But for the amount of data we have there, there is no point. So changed it to min()/max() for security.

fluxy/operators/flux_timeseries_resample.py
alexdanjou: I broke the list comprehension and make this happens in the for loop, which has more sense I think.

fluxy/plots/flux_timeseries.py
alexdanjou: Moved it up, in the if statement that converted them first into numpy.datetime64

fluxy/plots/flux_timeseries.py
alexdanjou: Because this is not compatible with one of the version of matplotlib that I'm using (either the old one of the ICOS Jupyter Lab or the recent one on my laptop, I don't remember). Also I don't really know what it is doing and I'm not shocked by the figure I have when I remove this, so I assumed that it wont be missed (@hdelongueville @melodb @mo-aliceramsden who wrote that?)

fluxy/plots/utils.py
alexdanjou: I've checked and change the behaviour everywhere :)

alexdanjou:
> * the code in the cli seems to be very specific to your PARIS use case. It there a plan to change that in the future or do you keep it for PARIS ?

yes it is very specific, and now there is 3-4 files in it that are only used fir the annex generation. Maybe we should consider putting this in its own folder (inside `cli` or elsewhere)

fluxy/plots/flux_timeseries.py
mo-aliceramsden: At some point these lines used to be:
```
try:
    for l in leg.legend_handles:
        l.set_linewidth(3.0)

except:
    for l in leg.legendHandles:
        l.set_linewidth(3.0)
```
Because different xarray versions used `legend_handles` or `legendHandles`, so this was a way of making sure this still worked with different xarray setups. But if we're moving towards specifying an xarray version, then we can just use `legend_handles` which is the more recent name.

Without `l.set_linewidth(3.0)`, the coloured lines in the legend are very narrow and are hard to see (especially when using the plots in a presentation) so I think we should try and keep this in, if we can.

configs/models_info.json
```diff
@@ -35,8 +35,8 @@
     },
 
     "standard_run":{
-        "default": {"ch4": "NAME_EUROPE_EDGARWETCHARTS_31sites",
-                    "n2o": "NAME_EUROPE_EDGAR_26sites",
+        "default": {"ch4": "NAME_EUROPE_EDGAR-WetCHARTs_longrun_31sites",
+                    "n2o": "NAME_EUROPE_EDGAR_longrun_24sites",
```
melodb: Is there a need to commit these filename changes? The longrun does not exist for ELRIS so I would not use it for the standard run.

fluxy/plots/flux_timeseries.py
alexdanjou: Great thanks! I put your code back (but change the `try` `except` by a check on the version of matplotlib)

fluxy/plots/flux_map.py
```diff
@@ -492,16 +487,22 @@ def plot_flux_map_over_time(
...
-    ds_chopby = {}
+    ds_chopby, time_labels = {}, {}
     for key, ds in ds_dict.items():
-        ds_chopby[key], time_labels = average_over_period(ds, dt, chop_by)
+        ds_chopby[key], time_labels[key] = average_over_period(ds, dt, chop_by)
+
+    if all([v == time_labels[key] for v in time_labels.values()]):
+        time_labels = time_labels[key]
+    else:
+        raise ValueError(f"Uncoherent `time_labels` derived : {time_labels}.")
```
lionel42: Some one might get lost with such an error, would be nice to say what can be the cause of getting this uncoherents time labels

fluxy/plots/flux_timeseries.py
```diff
@@ -367,10 +377,15 @@ def plot_country_flux(
...
             leg = ax.legend(ncol=ncol, borderpad=0.4, columnspacing=1.0)
-            for l in leg.legend_handles[: (-1 if plot_inventory else None)]:
-                l.set_linewidth(3.0)
+
+            if float(matplotlib_version[:3]) > 3.7:
+                for l in leg.legend_handles[: (-1 if plot_inventory else None)]:
+                    l.set_linewidth(3.0)
+            else:
+                for l in leg.legendHandles[: (-1 if plot_inventory else None)]:
+                    l.set_linewidth(3.0)
```
lionel42: Here you could do the if statetement to get the name of the attribute and then use getattr to access that attribute.
This way you can take the for loop out of the if else statement

fluxy/plots/utils.py
```diff
@@ -92,24 +92,25 @@ def print_cbar_label(
...
+                (e.g. "posterior_prior_diff", "flux_total_prior", ...)
+            species_info (dict):
+                A dictionary with metadata for species, including display names.
+            var (str, optional):
+                The variable name in the dataset.
+            season (str, optional):
+                The season to include in the label (e.g., 'DJF', 'MAM').
+            format (list[str], optional):
+                Specifies the components to include in the label.
+                Options: ['variable', 'species', 'units', 'time']. Default includes all.
```
lionel42: Why indenting here ? Did the formatter do that  ?

fluxy/plots/utils.py
```diff
@@ -240,11 +236,16 @@ def print_period(
             A formatted string representing the dataset's time period, optionally including a season.
     """
 
-    time = ds.time
     datetime_format = f"datetime64[{freq}]" if season is None else "datetime64[Y]"
 
-    start_date = time.values[0].astype(datetime_format)
-    end_date = time.values[-1].astype(datetime_format)
+    if "time" in ds.dims:
+        start_date, end_date = ds.time.values[0, -1]
```
lionel42: Don't assert the time is sorted, not true with the new indexing format

lionel42:
@alexdanjou Seems to work fine for me, just the few additional changes and this should be good. Since it is a big PR I think a review from someone else might be necessary.

fluxy/plots/flux_timeseries.py
alexdanjou: In fact, the former name is not deprecated (yet ?) so there is no need for this `if` statement if we keep to the old naming.

lionel42 (commented):
Good to me know, I would like to have someone else also doing a review

lionel42 (approved):

melodb:
> Good to me know, I would like to have someone else also doing a review

I will revise it on Friday.

hdelongueville:
I'll also revise it later this week :)

alexdanjou:
> Good to me know, I would like to have someone else also doing a review

Thanks a lot @lionel42 !

mo-aliceramsden (approved):
I've tested running annex_plot_generator.ipynb for a couple of example species (both monthly and annual) using my local copy of fluxy and this is all working really well. The 'config_annex_plot' and 'filenames.json' options are very useful.

I also tried plotting all_hfc when some of the individual HFC files were missing and this gave all the expected (very helpful!) error messages. 

So this looks to all be working as I'd expect, I'm happy for this to be merged in :)

fluxy/operators/flux_align_dataset.py
```diff
@@ -23,6 +24,12 @@ def align_time(ds_list: list[xr.Dataset]) -> list[xr.Dataset]:
         raise ValueError("Unable to infer period from dataset")
     period = np.median(dtime)
 
+    # Reduce datasets to their overlapping time range
+    if np.unique([ds.time.size for ds in ds_list]).size != 1:
```
melodb: As far as I understand, the code will only enter this if-statement if the size of the time array is different between the datasets. E.g. 2018-2023 (size=6) and 2012-2023 (size=12). This does not cover the case for which one data set has data from 2017-2022 and other from 2018-2023 (both with 6 years of data). I guess in this case, we also want to enter the if-statement and reduce both datasets to 2018-2022. Maybe the min and max should be checked instead.

fluxy/operators/flux_align_dataset.py
```diff
@@ -83,20 +90,25 @@ def align_lat_lon(
     return aligned_ds_list
 
 
-def align_map_data(ds_all: dict[xr.Dataset]) -> dict[xr.Dataset]:
+def align_map_data(
+    ds_all: dict[xr.Dataset | xr.DataArray],
+) -> dict[xr.Dataset | xr.DataArray]:
```
melodb, line 94: The closing bracket "]" is in the wrong position. I guess you want:
```
ds_all: dict[xr.Dataset] | xr.DataArray,
) -> dict[xr.Dataset] | xr.DataArray:
```

fluxy/operators/flux_align_dataset.py
```diff
@@ -83,20 +90,25 @@ def align_lat_lon(
...
     """
-    Prepare flux datasets for flux maps by filtering variables, removing unused dimensions, and aligning time and spatial coordinates.
+    Prepare flux datasets/dataarray for flux maps by filtering variables, removing unused dimensions, and aligning time and spatial coordinates.
 
     Args:
-        ds_all (dict[xr.Dataset]):
-            Dictionary of model names and corresponding xarray datasets.
+        ds_all (dict[xr.Dataset | xr.DataArray]):
```
melodb, line 99: Same here and in the Returns part

fluxy/operators/flux_map_diff.py
```diff
@@ -3,7 +3,7 @@
 
 def define_var_plot(
     ds: xr.Dataset,
-    var: str,
+    var: str | list[str],
```
melodb, line 6: Just to check: why was this option to accept a list implemented? As far as I can tell, `define_var_plot` is only used in `plot_flux_map` (where `var` is an element of `vars_list` in a for-loop),` plot_flux_model_comparison` and `plot_flux_map_over_time`. For the latter 2 functions, the code breaks when setting `var` to a list, so I have the feeling it is not used anywhere. Maybe I'm missing something.

fluxy/operators/flux_map_resample.py
```diff
@@ -171,7 +176,7 @@ def average_over_seasons(
     ordered_seasons = [s for s in desired_order if s in ds_avg.time.values]
     ds_avg = ds_avg.sel(time=ordered_seasons)
 
-    time_labels = ds_avg.time.values.tolist()
+    time_labels = ["Dec -> Feb", "Mar -> May", "Jun -> Aug", "Sep -> Nov"]
```
melodb: I don't think this should be hard-coded. It can happen that the datasets have data only from March to August. I don't remember the specific case, but there was a reason for checking if the seasons exist in L.176 (`ordered_seasons = [s for s in desired_order if s in ds_avg.time.values]`). Maybe @hdelongueville remembers.

One option would be to retrieve time_labels from `ds_avg.time.values.tolist()` and then have a dictionary that translates that into something else.

Btw, I find the arrows quite ugly (sorry @lionel42 xD). In general, we use "-" to define time ranges (e.g. "2018 - 2020") and everyone understands. So I would also use "-" here.

fluxy/operators/flux_map_resample.py
```diff
@@ -199,7 +204,7 @@ def average_over_years(
             Labels for each period (e.g., "2020", "2020—2022").
     """
 
-    groups = ds.time.dt.year // N
+    groups = (ds.time.dt.year - ds.time.dt.year.min()) // N
```
melodb, line 212: Nice! It's working beautifully!

fluxy/plots/flux_timeseries.py
```diff
@@ -367,9 +377,12 @@ def plot_country_flux(
 
         # set legend if needed
         if not set_global_leg:
-            ncol = 3 if annex_mode else 2
+            ncol = len(ds_to_plot) + 1 if annex_mode else 2
             leg = ax.legend(ncol=ncol, borderpad=0.4, columnspacing=1.0)
-            for l in leg.legend_handles[: (-1 if plot_inventory else None)]:
+
+            # handle_name = "legend_handles" if float(matplotlib_version[:3]) > 3.7 else "legendHandles"
+            # for l in leg.__getattribute__(handle_name)[: (-1 if plot_inventory else None)]:
```
melodb: Just to clarify: are these comments needed?

fluxy/cli/annex_plot_generator.py
```diff
@@ -1,931 +1,639 @@
...
+from fluxy.config import set_model_colors
+from fluxy.plots.flux_timeseries import plot_country_flux
+from fluxy.plots.flux_map import plot_flux_map_over_time
+
+logger = logging.getLogger(__name__)
 
 ###########################################
 ### GENERAL SETTINGS
 ###########################################
 # Species to plot
```
melodb: I guess these comments can be deleted

fluxy/cli/annex_plot_generator.py
```diff
@@ -1,931 +1,639 @@
...
+        "InTEM_longrun",
+        "InTEM",
+        "ELRIS",
+        "RHIME",
+    ]  # NOTE: only options are basic model names w/ and w/o longrun
+    models_yearly_species = [
+        "InTEM",
+        "ELRIS",
+        "RHIME",
+    ]
```
melodb: This is not important from the code review point of view, but in the previous annex we were always plotting intem_longrun, intem, elris and rhime (except for N2O for which there were no rhime results but that does not apply anymore). In all plots, the blue line was intem_longrun (and not intem) and the black line was the combination of intem, elris and rhime. Is intem and intem_longrun currently the same file for the yearly species?

fluxy/cli/annex_plot_generator.py
```diff
@@ -1,931 +1,639 @@
...
+    )
 
-        ### Define plotting colors
-        model_colors = func.set_model_colors(models_std, m_colors)
+    kwargs_maps_seasonnal = dict(
+        var="posterior_mean_diff_inversion_grid",
+        cmap="coolwarm",
+        c_border="dimgrey",
+        chop_by="season",
+        dt=[[12, 1, 2], [3, 4, 5], [6, 7, 8], [9, 10, 11]],
```
melodb: `dt` is not needed anymore. `chop_by="season"` takes care of it.

fluxy/cli/annex_plot_generator.py
```diff
@@ -1,931 +1,639 @@
...
+        json_exp_to_use :
+            Path to json file containing specific exp. This should be only used for testing.
+
+    Returns:
+        models_std :
+            List of exp names to be used for the plots and tables for the annex for the species passed as input.
+    """
+    logger.warning(
+        "Each team should check that the right files are used. Still to be checked by RHIME, ELRIS, InTEM."
+    )
```
melodb: Can we move this warning somewhere else? I think it is ok if it shows up only once at the start of the code.

fluxy/cli/annex_plot_generator.py
```diff
@@ -1,931 +1,639 @@
...
+        # 1.1) Plot annual country fluxes from 2008 to 2023 from intem_longrun and combined from 3 std_run
+        print(f"- Annual country fluxes {start_date[0]} - {end_date}")
+        fig, res_dict = plot_country_flux(
+            ds_all_flux_scaled,
+            species=species,
+            model_colors=model_colors,
+            model_labels=model_labels,
+            start_date=start_date[0],
+            end_date=end_date,
+            resample=[None, "year", "year", "year"],
```
melodb: First I was wondering why you were not resampling the intem_longrun. I get now that you chose to plot the yearly run. If it's not too much work, I think it would be nicer to join these settings in config so that we can easily discuss them in a PARIS meeting.

fluxy/cli/annex_plot_generator.py
```diff
@@ -1,931 +1,639 @@
...
+
+        if json_exp_to_use:
+            with open(Path(json_exp_to_use), "r") as f:
+                json_data = json.load(f)
+            model_read = f"{model.split('_')[0]}_{json_data[model][species]}"
+
+        else:
+            model_read = f"{model.split('_')[0]}_{config_data['models_info']['standard_run']['default'][species]}"
+            if "longrun" in model:
+                model_read = f"{model_read}_longrun"
```
melodb, line 59: I just realized that fluxy is not ready to pick the `standard_run` from a given species and that this feature is only used when plotting all_hfc or all_pfc. That's why you had to manually do it here. I cannot think about a better solution at the moment. However, I think it would be better to align this part of the code with what is done when reading all hfcs/pfcs (see io.py, L.339-362). In this way, we can specify different standard runs for different models. Otherwise, we would be forced to use filenames.json.

fluxy/cli/annex_plot_generator.py
```diff
@@ -1,931 +1,639 @@
...
-            m0 = model.split("_")[0]
+    # Settings for seasonal difference to the mean
+    kwargs_maps_general = dict(
+        config_data=config_data,
+        region=region,
+        set_fluxlim="auto",
+        plot_combined=True,
+        add_sites=True,
+        add_markers=annex_config.point_markers[region],
+    )
```
melodb: I would suggest moving all these dictionaries to config_annex_plot, including the model lists, start/end dates and other species specific settings. I think it would make it easier to change them afterwards if needed.

fluxy/cli/annex_plot_generator.py
```diff
@@ -1,931 +1,639 @@
...
+        start_date = ["2008-01-01", "2018-01-01", "2018-01-01", "2018-01-01"]
+        end_date = "2024-01-01"
+
+        models_std = define_model_list(
+            models_monthly_species, species, config_data, json_exp_to_use
+        )
+        period = [
+            "monthly" if "longrun" not in m else "yearly"
+            for m in models_monthly_species
+        ]
```
melodb: This is not relevant from the code review point of view, but it's not clear to me why we use the yearly intem longrun instead of the monthly. Maybe this type of settings can also go to config_annex_plot so that they are not hidden within the code.

melodb (commented):
Thank you so much Alex for all this work! I only had problems getting the N2O spatial maps plotted but this is nothing to do with your implementation, I might have the wrong files. I left some comments. See what you can do. Thanks a lot once again! (and thanks for your patience in getting a review from my side)

fluxy/operators/flux_map_resample.py
hdelongueville: You are right, sometime a month or a season can be missing. 
Silly question but why don't we keep, eg, "JJA"? Is an esthetical reasol?
Anyway, I like the idea of using a dictionary.

fluxy/plots/flux_map.py
```diff
@@ -492,16 +487,24 @@ def plot_flux_map_over_time(
...
+    ds_chopby, time_labels = {}, {}
     for key, ds in ds_dict.items():
-        ds_chopby[key], time_labels = average_over_period(ds, dt, chop_by)
+        ds_chopby[key], time_labels[key] = average_over_period(ds, dt, chop_by)
+
+    if all([v == time_labels[key] for v in time_labels.values()]):
+        time_labels = time_labels[key]
+    else:
+        raise ValueError(
+            f"Uncoherent `time_labels` derived : {time_labels}. Most probable reason is difference between start and end dates of the datasets, slicing them to thei common period should resolve the issue."
```
hdelongueville: thei --> their

fluxy/plots/utils.py
```diff
@@ -95,8 +95,9 @@ def print_cbar_label(
     Generate a colorbar label for a dataset variable.
 
     Args:
-        ds (xr.Dataset):
-            The dataset containing the variable.
+        ds (xr.DataArray):
+            The DataArray containing the variable. The DataArray should be named with the name of the variable^M
```
hdelongueville: ^M = typo?

hdelongueville (commented):
Thanks a lot for the great work!!

hdelongueville:
Do you think you could fix this issue https://github.com/openghg/fluxy/issues/118 in this PR?
(see Daniela's solution)

fluxy/cli/annex_plot_generator.py
melodb: One option would be to implement a flag in `read_model_output` (e.g. read_std_file) which would get the filename from models_info when set to True. The option to do it automatically if only the model name tags are given is not very robust because currently we do not force the filename tags to have more than one element. Output files can be simply named "ELRIS_ch4_monthly.nc".

configs/models_info.json
alexdanjou: No indeed, I put it back

fluxy/operators/flux_align_dataset.py
alexdanjou: Didn't thought about that indeed. But my problem is also that min and max may still differ from 15 days/6 months if people don't respect the convention (`"mid of the flux interval in UTC"`), and even sometimes a couple of days if the middle of the month/year is poorly defined...

I'll try to propose something smart.

fluxy/operators/flux_align_dataset.py
alexdanjou: No I confirm that I want a Dictionnary of datasets or a dictionnary of dataarrays

fluxy/operators/flux_align_dataset.py
alexdanjou: I confirm that it should work with a dictionnary of datasets or a dictionnary of dataarrays :)

fluxy/operators/flux_map_diff.py
alexdanjou: It is to prepare the way for the uncertainty shading in those two plot functions. That's why I've also kept a datasets in all those functions, even though there is only one variable for now and a dataarray would have been easier. see discussion in issue #123

fluxy/cli/annex_plot_generator.py
alexdanjou: I had to do put this `if` condition because right now RHIME and InTEM did not put their standard files with the expected names. As soon as it is corrected (and it should be for the annexes), it won't be nessecary.

I was thinking of deleting `filenames.json` for my final push. It should be there only as a temporary option for testing (that's way I put `json_exp_to_use` as an optionnal parameter of produce_plots).

fluxy/cli/annex_plot_generator.py
alexdanjou: Because the monthly does not (didn't?) went as far as the monthly one for ch4 (and n2o?)

alexdanjou: But I agree it is a bit dodgy. I'll suggest to put something better when it will be become necessary (so not right now)

fluxy/operators/flux_align_dataset.py
alexdanjou: The best solution that I have is to simply do this slicing in every case (whihch means remove the `if` statement).

melodb: Fine for me :)

fluxy/operators/flux_map_resample.py
alexdanjou: Changed to a dictionnary
```
    time_labels = {
        "DJF": "Dec - Feb",
        "MAM": "Mar - May",
        "JJA": "Jun - Aug",
        "SON": "Sep - Nov",
    }
    return ds_avg, [time_labels[s] for s in ordered_seasons]
    
    ```
    
    I think Dec - Feb is more comprehensible than DJF

melodb (approved):
I think this can be merged. Thank you Alex for the great work! I like the AnnexConfig class very much! Maybe we can use a PARIS meeting to discuss the plotting options and make sure that no further developments are needed.

Issue #127, opened 2025-05-29
Add Corsica when selecting 'FRANCE' with plot_flux_map


hdelongueville:
When plotting the spatial flux maps for France, Corsica is missing from the plot. 

The issue comes from `get_region_coordinate`:
```
# Remove overseas territories by keeping only the largest landmass for each country
    region["geometry"] = region["geometry"].apply(
        lambda geom: extract_largest_polygon(geom)
    )
```

This bit is to remove overseas territories like, for example, Guadeloupe which is masked as `FRANCE`. 
One solution is to add an extra case specific to France with a fixed border limits.

hdelongueville:
Or, now that we have the variable `country_fraction`, we could use it to determine the boundaries of the region.

lionel42:
@hdelongueville I just noticed this issue. Wouldn't it make more sense to simply apply a geometrical clip ( https://geopandas.org/en/stable/docs/reference/api/geopandas.clip.html#geopandas-clip ) to the country shapes so that they fix the extent of the plot ?

hdelongueville:
Oh I didn't know you could do that, so useful! Yes we should definitely use this, thanks for sharing!

alexdanjou:
Same kind of issue for UK :  Northern Ireland and some Scottish islands are cut off

PR #142, opened 2025-06-10
Fix issue #127

alexdanjou:
Use geopandas.clip to restrict country extension un flux maps rather than plotting the biggest land mass (closes issue #127 )

### DIFF
--- a/fluxy/plots/utils.py
+++ b/fluxy/plots/utils.py
@@ -545,9 +545,12 @@ def get_region_coordinates(
         raise ValueError(f"No coordinates found for region '{region_name}'.")
 
     # Remove overseas territories by keeping only the largest landmass for each country
-    region["geometry"] = region["geometry"].apply(
-        lambda geom: extract_largest_polygon(geom)
-    )
+    if region["CONTINENT"].apply(lambda x : x.lower()=="europe").all():
+        region = gpd.clip(region,[-30,30,50,75])
+    else:
+        region["geometry"] = region["geometry"].apply(
+            lambda geom: extract_largest_polygon(geom)
+        )
 
     # Get the bounding box of the region of interest
     region_boundaries = region.total_bounds  # [minx, miny, maxx, maxy]
--- a/fluxy/plots/utils.py
+++ b/fluxy/plots/utils.py
@@ -464,17 +464,20 @@ def get_map_bounds(
     ds_all = list(ds_all)
     if isinstance(region, str):
         # Use the non-zero country_fraction to define the clipping region, for coherence in the country definition
-        da_mask = ds_all[0].country_fraction.sum(dim="country")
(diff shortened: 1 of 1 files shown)


hdelongueville:
I'm wondering if we should not totally remove "extract largest polygon" to avoid any confusion.

hdelongueville:
Instead of hard coding the boundaries, shouldn't we not extract the boundaries of the domain instead? (I don't know if this is too difficult to do)

hdelongueville:
Ah but we have the outer regions which can include overseas territories..

hdelongueville:
Is there a way to extract the inner region boundaries? We could use that if it is not too complicated :)

hdelongueville:
Just to check, is the country_fraction a mandatory variable in the template? If not, this might cause issue if this variable is not available in the dataset no? (For example, I know that RHIME doesn't have this variable for the NID 2024 files).

alexdanjou:
> Just to check, is the country_fraction a mandatory variable in the template? If not, this might cause issue if this variable is not available in the dataset no? (For example, I know that RHIME doesn't have this variable for the NID 2024 files).

It is mandatory now! See https://github.com/openghg/fluxy/blob/642896d822ecf3d856b17985071f86c347804705/data/templates/PAR-AVE-EYE_inversion_flux_output.cdl#L280

hdelongueville (changes requested):
It is working well for me, as long as I am not only looking at old rhime results (which do not include country_fraction). Can I suggest to add the case where if country_fraction does not exist, skip the clipping part. It will keep the overseas territories but at least the cell works, and the user can decide to zoom with the other options afterward.

alexdanjou:
> It is working well for me, as long as I am not only looking at old rhime results (which do not include country_fraction). Can I suggest to add the case where if country_fraction does not exist, skip the clipping part. It will keep the overseas territories but at least the cell works, and the user can decide to zoom with the other options afterward.

I've added an `if` statement for backward compatibility. It will however skip the clipping and we thus have the previous issue with France and its overseas territory. But I guess there is no point to put back your previous fix as it will soon be outdated.

Tell me your thoughts on this!

alexdanjou:
Test are passing on my machine. Issue with `xarray` version on github I think

hdelongueville:
Yes I agree with your comment! That is also what I had in mind. At least, you can start from the wider map and use it to determine the lat/lon borders you want and then use the region = [lat, lon] option :)

hdelongueville:
Is there a way to fix the issue?

hdelongueville (approved):
Thanks for the modifications!!

alexdanjou:
> Is there a way to fix the issue?

I think it was an issue with the new `xarray` version of github. Seems to be fixed now as tests are passing!

Issue #135, opened 2025-06-06
file config.py still contains old names


lionel42:
As @hdelongueville mentionned in #121 , after the merge, we still have some old names.

We should define what names they should be given:

* YaprioriBC -> mf_bc_prior
* Yapriori_bias -> mf_bias_posterior

And some names are missing:

* mf_bias_prior

we should fix that in a new pr

PR #144, opened 2025-06-12
Corrections to new format implementation

melodb:
This fixes issue #135 and remaining comments from Helene to #121
I have also deleted column "Old name" from README_templates.md as this was only meant to support @lionel42 in his PR.

**NOTE:**
Currently in `flux_map_model_comparison`, the sites attribute is not added to dataset "diff" even if it exists in both datasets being compared. Therefore, it will use the fallback_sites. One idea would be to plot the sites from both datasets being compared using different symbols. I think this can stay for other PR if you agree.


melodb:
Alternatively, I can make variable `add_sites: bool | list[str]`. In this case, `if isinstance(add_sites, list)`, I can set internally `fallback_sites = add_sites` and `add_sites = True`. This would avoid an additional input parameter. I can adapt the code in this way if you prefer.

hdelongueville:
1) I agree with your note.
2) I think we can keep the way you implemented it, no need for changes from my point of view. :)

hdelongueville (approved):

PR #145, opened 2025-06-13
Delete NaN variables from mf dataset

melodb:
In the CIF-EnKS dataset some variables were all NaN: `intake_height` and `stdev_mf_model`. This was leading to problems when using `.dropna('index)` in` stats_mf`. For now, this check is done only for CIF-EnKS, but this can be made general if it becomes a problem for other datasets.

### DIFF
--- a/fluxy/io.py
+++ b/fluxy/io.py
@@ -789,7 +789,7 @@ def edit_vars_and_attributes(
             # Delete NaN data variables
             for var in ds.data_vars:
                 if ds[var].isnull().all():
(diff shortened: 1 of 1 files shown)


lionel42 (changes requested):
This is not a very elegant solution. I would prefer adapting the stats mf code to remove nas only in variables you have trouble with instead of removing any variables. What do you think ?

melodb:
Thanks for the suggestion! @jthanwer in case you want to work on your Taylor plot implementation this week I think you can use this fix to handle your files for now. I will make it better next week.

Issue #140, opened 2025-06-10
Alternative country totals


ayanu:
It would be great if we can find a way to include alternative country totals, other than UNFCCC, to the country emission time series. 
These could be taken from previously published estimates or from alternative inventories. 
Should include the possibility to depict with uncertainties. 
Either individual points with error bars or additional mean line with ribbon. Will depend on the kind of result available. Single years vs. long time series. 
More than one alternative estimate should be treated. 

Functionality would help to further contextualise our results.

melodb:
@mo-aliceramsden I'm assigning you to this just in case you are able to address it while working on #111 and #82. Maybe in addition to setting folder name ("inventory") and filename tag ("UNFCCC_inventory") to inputs to `plot_country_flux` (with defaults set to "inventory" and "UNFCCC_inventory"), the folder name and filename tag input variables could also accept lists so that different inventory files can be provided in different folders. The plotting style could also be an input to the function (e.g. "bar", "scatter", "line") which would also accept a list or a dictionary (e.g. `{"inventory": "bar", "Manning_etal_2021": "line"}`). Please feel free to un-assign yourself if you feel that you won't have time for this.

melodb:
Moreover, the inventory folder name can be used to define the plot label. Maybe user-defined labels can also be defined in models_info.json using the folder name as the name tag.

PR #147, opened 2025-06-17
Inventory and baseline timestamps format update

### LINKED ISSUE #82: Inventory and baseline timestamp filenames are hard-coded
Currently, the inventory filenames are hard-coded in regions.py: `f"UNFCCC_inventory_{species}_{inventory_year}.nc"`
It would be better to set the inventory identifier as an input to the plotting function and set "UNFCCC_inventory" as default.
The same is true for the baseline timestamp files (hard-coded in select.py).

### LINKED ISSUE #111: Use ISO_A3 for "country" variable in inventory netcdf file
Currently, countries are identified by their ISO 3-letter code in the flux file. However, in the inventory file we use the full country name. Therefore, while country fluxes can be plotted without the `regions_info.json` config file (using the 3 letters code), inventory data cannot be plotted if `regions_info["country_codes"]` dict is not defined.

I think it would be more consistent to use the ISO 3-letter code in the inventory netcdf files. Moreover, it would suppress the need for regions_info which is useful for a beginner user.

This issue is related to PR #108

mo-aliceramsden:
Changes made:

- Updated inventory file format to better match flux netcdfs (closes #111)
       - Variable names are now `flux_total_inventory_country` and `stdev_flux_total_inventory_country` (to match the flux file names of `flux_total_posterior_country` and `stdev_flux_total_posterior_country`). I used the word `inventory` to differentiate between priors and these inventory files, but could change these variable names to `flux_total_prior_country` if everyone thinks that's better?
        - Country names have been changed from longnames (e.g. 'FRANCE') to 3-letter codes (e.g. 'FRA') in the inventory netcdfs. Issue #111 also suggested removing the requirement for a regions_info.json file, but this would require the user to choose a country in the notebook by writing a 3-letter code (instead of the longname) which means they would have to know the 3-letter code in advance. So for now I've left in the requirement for a regions_info.json file, and this file is still used to match the user's chosen country to the 3-letter code, which is used to extract inventory data.
        - 
- I've added options to select custom inventory and baseline timestamp filenames, leaving current filenames as default values (closes #82)

- And I've added templates for inventory and baseline timestamp files.

I'll include suggestions from #140  in a separate PR, as they aren't as urgent.


mo-aliceramsden:
I've added a new directory to the ICOS jupyterlab: `/project/paris/inverse_modelling/baseline_timestamps`. Once this branch is merged in, the old directory `intem_baseline_timestamps` should be deleted.

alexdanjou (approved):
Good for me, thanks for those needed enhancements!

fluxy/operators/flux_prepare_inventory.py
```diff
@@ -44,7 +46,8 @@ def retrieve_inventories(
 
     for year, inv_color in zip(inventory_years, inv_colors):
         ds_inv = extract_region_inventory_flux(
-            data_dir, country, species, unit, s_data, r_data, inventory_year=year
+            data_dir, country, species, unit, s_data, r_data, inventory_year=year,
+            inventory_filename=inventory_filename
```
lionel42: try to respect black formatting if possible

fluxy/operators/regions.py
```diff
@@ -206,21 +207,24 @@ def extract_region_inventory_flux(
...
         inventory_year: year of inventory to get.
-
+        inventory_filename: Name of inventory file: {inventory_filename}_{species}_{inventory_year}
     Returns:
         dataset with country selected
 
     """
+    
+    if inventory_filename is None:
+        inventory_filename = 'UNFCCC_inventory'
```
lionel42: should not be necessary as you specify the default value in the function signature

fluxy/operators/regions.py
```diff
@@ -229,10 +233,19 @@ def extract_region_inventory_flux(
...
+
+    inv_ds_all = xr.open_dataset(filepath)
+        
+    if inv_ds_all.attrs['missing_data'] != '[]':
+        logger.warning(f"Inventory is missing data: {inv_ds_all.attrs['missing_data']}")
+    
+    if 'inventory' in inv_ds_all.keys():        #for compatability with older inventory netcdfs, can be removed later
+        inv_ds = inv_ds_all['inventory']    
+    else:
+        inv_ds = inv_ds_all["flux_total_inventory_country"] #could add in sector-level read in here
```
lionel42: use a one line if else statement to get the variable name

fluxy/operators/regions.py
```diff
@@ -245,27 +258,32 @@ def extract_region_inventory_flux(
...
         logger.info(f"Converting to mass of CO2-eq using GWP = {gwp}.")
+        
     scaling_factor = get_units_conversion_factor(origin_unit, target_unit, molar_mass)
 
     inv_ds = inv_ds * scaling_factor * gwp
     inv_ds.attrs["units"] = unit
     inv_ds.attrs["year"] = inventory_year
 
-    if country in inv_ds["country"]:
+    if country in inv_ds["country"]:            #for compatability with older inventory netcdfs, can be removed later
```
lionel42: follow black format for the comment (usually put the comment on a line alone just above the line you want to comment, it is clearer)

fluxy/operators/regions.py
```diff
@@ -245,27 +258,32 @@ def extract_region_inventory_flux(
...
+       
+    country_list_update = []
+    
+    for country in country_list:            #for compatability with older inventory netcdfs, can be removed later
+        if country in inv_ds['country']:
+            country_list_update.append(country)
+        elif country in r_data['country_codes'].keys():
+            country_list_update.append(r_data['country_codes'][country])
+        else:
+            country_list_update.append(dict(map(reversed, r_data["country_codes"].items()))[country])
```
lionel42: The list comprension was a more elegent solution to this. Maybe you can create a function for the if else statement and use the function in the list comprehension ?

fluxy/operators/select.py
```diff
@@ -136,6 +137,10 @@ def slice_mf(
 
     # Get logical array with baseline timestamps
     if baseline_site is not None:
+        
+        if baseline_filename is None:
+            baseline_filename = "InTEM_baseline_timestamps"
```
lionel42: put the default value in the function signature for clairty

scripts/PARIS_inversion_results.ipynb
```diff
@@ -744,7 +748,7 @@
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "Python 3 (ipykernel)",
+   "display_name": "base",
```
lionel42: avoid to push these metadata changes in notebook, they tend to create confilcts. See also the change below

lionel42 (changes requested):

mo-aliceramsden:
Thanks for the review @lionel42, I think I've managed to address all of your suggestions. I'm new to black formatting and python type hinting, so hopefully I've managed to keep everything consistent!

lionel42 (approved):
@mo-aliceramsden Looks good to me know, thanks for your work :)

fluxy/plots/flux_timeseries.py
```diff
@@ -163,6 +163,7 @@ def plot_country_flux(
     annex_mode: bool = False,
     plot_inventory: bool = True,
     inventory_years: list[str] | None = None,
+    inventory_filename: str = 'UNFCCC_inventory',
```
melodb, line 166: If the default value for `inventory_filename` is already set here, can we remove the default from all intermediate functions (flux_prepare_inventory.py, regions.py)? Just a suggestion.

fluxy/operators/regions.py
```diff
@@ -245,27 +256,33 @@ def extract_region_inventory_flux(
...
     inv_ds = inv_ds * scaling_factor * gwp
     inv_ds.attrs["units"] = unit
     inv_ds.attrs["year"] = inventory_year
 
-    if country in inv_ds["country"]:
+    #for compatability with older inventory netcdfs, can be removed later
+    if country in inv_ds["country"]:            
         return inv_ds.sel(country=country)
+    elif r_data['country_codes'][country] in inv_ds['country']:
+        return inv_ds.sel(country=r_data['country_codes'][country])
```
melodb: Maybe here we can use the same approach used in extract_region_flux (L.43-44 in this file):
```
# Get country_codes only if regions_info exists    
country_codes = regions_info.get("country_codes", {})
# Look for the code if country_codes is defined, otherwise assume the code was given as input
country_search = country_codes.get(country, country)
if country_search in inv_ds["country"]:
    return inv_ds.sel(country=country_search) # new format
elif country in inv_ds["country"]:
    return inv_ds.sel(country=country) # old format (would only work if the user specifies the country name)
```
Or something like this.

This would delete the need for regions_info if the user specifies the code directly (when using the new format).

fluxy/operators/select.py
```diff
@@ -94,6 +94,7 @@ def slice_mf(
     end_date: str = None,
     site: str = None,
     baseline_site: str = None,
+    baseline_filename: None | str = "InTEM_baseline_timestamps",
```
melodb: If the default is a string, does it make sense to have "None" here?

fluxy/operators/regions.py
```diff
@@ -245,27 +256,33 @@ def extract_region_inventory_flux(
...
     country_list_update = [
         (
             country
-            if country in inv_ds["country"]
-            else dict(map(reversed, r_data["country_codes"].items()))[country]
-        )  # type: ignore
+            if country in inv_ds['country']
+            else r_data['country_codes'][country] if country in r_data['country_codes'].keys()
+            else dict(map(reversed, r_data["country_codes"].items()))[country] 
+        )
```
melodb: Maybe you can take a look at how it is done in extract_region_flux and see if you can avoid the mandatory use of regions_info here.

melodb (commented):
Thanks Alice! All is working fine on my side with the old files. In addition to my comments I would only suggest to add to the README file the location of the templates (sections 6 and 7).

fluxy/operators/regions.py
melodb: Just one additional comment. As you said, this requires the user to know in advance the 3 letters code which might not always be the case. However, the whole code is already prepared for that (plot_country_flux and plot_flux_map* will work if you set regions to "FRA"). So I think it would be nice to keep this feature valid even when setting `plot_inventory = True`.

mo-aliceramsden: Think I've got this working when the user inputs either full country names (e.g. 'FRANCE') or 3-letter country names (e.g. 'FRA'), so thank you for suggestions on this fix!

melodb (approved):
All working fine! Thanks for the changes and for these developments!

PR #152, opened 2025-06-24
Taylor plot

jthanwer:
Add support for comparing model statistics within Taylor diagrams with absolute or normalized data.
Normalization is performed automatically as part of the plotting function if requested.
The main function `plot_taylor_diagram` and associated class `TaylorDiagram` are implemented in the `fluxy/plots/mf_stats.py` script.
An example of how to use the new function has been added to the jupyter notebook.


fluxy/plots/mf_stats.py
```diff
@@ -1,8 +1,173 @@
...
+        ax.axis["left"].major_ticklabels.set_axis_direction("bottom")
+        ax.axis["left"].label.set_axis_direction("bottom")
+        ax.axis["left"].label.set_text(label_sd)
+
+        # Standard deviation axis (right)
+        ax.axis["right"].set_axis_direction("top")
+        ax.axis["right"].toggle(ticklabels=True, label=True)
+        ax.axis["right"].major_ticklabels.set_axis_direction("left")
+        ax.axis["right"].label.set_axis_direction("top")
+        ax.axis["right"].label.set_text(label_sd)
```
lionel42: Here you could define what you want to be specific to each ax in a dictionary and then loop over the 3 axis that you have

fluxy/plots/mf_stats.py
```diff
@@ -107,3 +272,154 @@ def plot_stats_mf(
...
+
+def plot_taylor_diagram(
+    stats: dict[Literal['prior', 'posterior'], pd.DataFrame],
+    model_colors: dict[str, str],
+    model_labels: dict[str, str],
+    model_markers: list[str] = ['o'],
+    normalize: bool = True,
+    plot_type_model: Literal['separate', 'together'] = 'separate',
+    plot_type_stat: Literal['separate', 'together'] = 'separate',
+    include: list[str] = ['prior', 'posterior'],
```
lionel42: If include is a substet of stats keys, you could use `list[Literal[...[]]`

fluxy/plots/mf_stats.py
```diff
@@ -107,3 +272,154 @@ def plot_stats_mf(
...
+
+        # Skip if stat is not in include
+        if s not in include:
+            continue
+
+        # Fetch statistical data
+        long_stat = pd.melt(stat, id_vars=['model', 'site'], value_vars=['pearson', 'sd_obs', 'sd_sim'])
+        df_pearson = long_stat[long_stat['variable'] == 'pearson'].pivot(index='site', columns='model', values='value')
+        df_sds_obs = long_stat[long_stat['variable'] == 'sd_obs'].pivot(index='site', columns='model', values='value')
+        df_sds_sim = long_stat[long_stat['variable'] == 'sd_sim'].pivot(index='site', columns='model', values='value')
```
lionel42: make a function to avoid duplicated code

fluxy/plots/mf_stats.py
```diff
@@ -107,3 +272,154 @@ def plot_stats_mf(
...
+
+            # Get the statistics for the model m
+            list_sds_obs = df_sds_obs[m].values
+            list_sds_sim = df_sds_sim[m].values
+            list_pearsons = df_pearson[m].values
+            
+            # Normalize the data if needed
+            if normalize:
+                list_sds_sim /= list_sds_obs
+                list_sds_obs /= list_sds_obs
```
lionel42: you normalize with the observations, also the simulation ?

fluxy/plots/mf_stats.py
```diff
@@ -107,3 +272,154 @@ def plot_stats_mf(
...
+                    sd_range=sd_range, 
+                    sd_unit=sd_unit,
+                    normalized=normalize,
+                )
+
+            else:
+                diag = dict_diags_pos[index_pos]
+
+            # Define labels, colors and markers
+            label = model_labels[m] + ' - ' + s
```
lionel42: use f string instead

scripts/PARIS_inversion_results.ipynb
```diff
@@ -758,7 +795,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.10.6"
+   "version": "3.12.0"
```
lionel42: don't change the python version of the notebook

scripts/taylor_plot.ipynb
```diff
@@ -0,0 +1,509 @@
...
+    "    \"\"\"\n",
+    "\n",
+    "    def __init__(\n",
+    "        self,\n",
+    "        sd_obs: float = None,\n",
+    "        fig: Figure = None,\n",
+    "        position: tuple[int, int, int] = (1, 1, 1),\n",
+    "        markersize: int = 100,\n",
+    "        sd_range: tuple[float, float] = (0, 1.5),\n",
+    "        sd_unit: str = \"ppb\",\n",
```
lionel42: This class is now in fluxy right, should not be in this notebook

scripts/taylor_plot.ipynb
```diff
@@ -0,0 +1,509 @@
...
+    "import pandas as pd\n",
+    "from typing import Literal\n",
+    "\n",
+    "\n",
+    "def plot_taylor_diagram(\n",
+    "    stats: dict[Literal['prior', 'posterior'], pd.DataFrame],\n",
+    "    model_colors: dict[str, str],\n",
+    "    model_labels: dict[str, str],\n",
+    "    model_markers: list[str] = ['o'],\n",
+    "    normalize: bool = True,\n",
```
lionel42: this also is in fluxy now, clear from the notebook

lionel42:
I think there is an issue with the notebook `taylor_plots.ipynb` did you upload it by mistake ? 

If you want to show all the different plots that you can do, it can be nice to keep it as well.

lionel42:
Forgot to mention before but you need to add a test as well ;)

fluxy/plots/mf_stats.py
```diff
@@ -107,3 +272,154 @@ def plot_stats_mf(
     )
 
     return fig
+
+
+def plot_taylor_diagram(
+    stats: dict[Literal['prior', 'posterior'], pd.DataFrame],
+    model_colors: dict[str, str],
+    model_labels: dict[str, str],
```
lionel42, line 291: Would be nice to not have to specify this and be able to you default colors and labels in the plot if they are not specified

lionel42 (changes requested):
Make sure you add a test of the `plot_taylor_diagram` before merging

fluxy/plots/mf_stats.py
```diff
@@ -1,8 +1,173 @@
...
+    """
+
+    def __init__(
+        self,
+        sd_obs: float = None,
+        fig: Figure = None,
+        position: tuple[int, int, int] = (1, 1, 1),
+        markersize: int = 100,
+        sd_range: tuple[float, float] = (0, 1.5),
+        sd_unit: str = "ppb",
```
melodb: In my view, it does not make sense to specify a default unit. I would leave it without a default.

fluxy/plots/mf_stats.py
```diff
@@ -1,8 +1,173 @@
...
+        )
+
+        # Create figure and subplot
+        if fig is None:
+            fig = plt.figure()
+        ax = fa.FloatingSubplot(fig, *position, grid_helper=gh)
+        fig.add_subplot(ax)
+
+        # Set the label for standard deviation
+        label_sd = "Normalized standard deviation [unitless]".format(sd_unit) \
```
melodb: I guess `.format(sd_unit)` is not needed here.

fluxy/plots/mf_stats.py
```diff
@@ -1,8 +1,173 @@
...
+            # Add RMSE countours
+            contours = self.add_contours(colors="0.5")
+            self.ax.clabel(contours, inline=1, fontsize=10)
+
+    def add_samples(self, sds: list[float], pearsons: list[float], label: str, *args, **kwargs):
+        """
+        Add markers representing sample points to the Taylor diagram.
+
+        Args:
+            sd_obs (list of floats):
```
melodb: Typo: rename `sd_obs` by `sds`

fluxy/plots/mf_stats.py
```diff
@@ -1,8 +1,173 @@
...
+                Standard deviations of the sample points
+            pearsons (list of floats):
+                Pearson's correlation coefficiens of the sample points
```
melodb, line 151: Add `label` to argument list?

fluxy/plots/mf_stats.py
melodb: Joël and I discussed this in the other day. I think we need to adjust this everywhere in fluxy so maybe this can be taken care of in a different PR. Currently, default colors are obtained by calling get_default_colors. This function is only called in prepare_data_to_plot when plotting the country totals. Maybe this should be expanded to all plotting functions. Moreover, once default colors are set for the first time, I think it would be nice to continue on using them for the rest of the plotting functions for consistency.

fluxy/plots/mf_stats.py
```diff
@@ -107,3 +272,154 @@ def plot_stats_mf(
...
+    stats: dict[Literal['prior', 'posterior'], pd.DataFrame],
+    model_colors: dict[str, str],
+    model_labels: dict[str, str],
+    model_markers: list[str] = ['o'],
+    normalize: bool = True,
+    plot_type_model: Literal['separate', 'together'] = 'separate',
+    plot_type_stat: Literal['separate', 'together'] = 'separate',
+    include: list[str] = ['prior', 'posterior'],
+    sd_range: tuple[float, float] = (0, 2.5),
+    sd_unit: str = 'ppb',
```
melodb: Same comment as before regarding default units.

fluxy/plots/mf_stats.py
```diff
@@ -107,3 +272,154 @@ def plot_stats_mf(
...
+        stats (dict of pd.DataFrame):
+            Dictionary containing statistics for each model and each site.
+            The keys should match the entries in the `include` parameter.
+        model_colors (dict of str):
+            Models and corresponding colours used to plot the model.
+        model_labels (dict of str):
+            Models and corresponding labels used to plot the stats.
+        model_markers (list of str):
+            Markers used to plot the stats in the `include` parameter.
+            If only one marker is given, it is used for all models.
```
melodb: Maybe say "(...), it is used for all stats in include". Otherwise it seems that we define one marker per model. However, as far as I understand, it's one marker for the prior and other for the porterior stats, right?

scripts/PARIS_inversion_results.ipynb
```diff
@@ -544,6 +544,43 @@
...
+    "    stats={'prior': stats_prior, 'posterior': stats_posterior},\n",
+    "    model_colors=model_colors,\n",
+    "    model_labels=model_labels,\n",
+    "    model_markers=['s'],\n",
+    "    normalize=True,\n",
+    "    plot_type_model='together',\n",
+    "    plot_type_stat='together',\n",
+    "    include=['prior', 'posterior'],\n",
+    "    sd_range=[0, 1.5],\n",
+    "    sd_unit='ppb'\n",
```
melodb: I would add a block with the variables to edit. In this way, the user does not need to edit the function call directly. You can also add a small explanation in front of each variable stating the different options (see other notebook cells).

Moreover, I would loop over `include` to create the different elements of `stats`. This would be useful for plotting "prior/posterior_above_BC" (see my general comment).

For example:

```
from fluxy.operators.mf import stats_mf
from fluxy.plots.mf_stats import plot_taylor_diagram

###################################
### edit variables in this block
include = ['prior','posterior']
plot_type_model = 'together'
plot_type_stat = 'together'
normalize = True
model_markers = ['s']
sd_range = [0, 1.5]
###################################

model_labels = set_model_labels(models, config_data, get_labels_from_file)

stats = {}
for stat in include:
    stats[stat] = stats_mf(ds_all_allsites, stats_type=stat)

fig = plot_taylor_diagram(
    stats=stats,
    model_colors=model_colors,
    model_labels=model_labels,
    model_markers=model_markers,
    normalize=normalize,
    plot_type_model=plot_type_model,
    plot_type_stat=plot_type_stat,
    include=include,
    sd_range=sd_range,
    sd_unit=mf_units_print,
)
```

Moreover, is `fig.show()` needed?

melodb (commented):
I have tried it with the HFC data and it is working very well. Thank you very much for this! I left some small comments in the code. In addition, I would recommend applying the black formatter to mf_stats.py for consistency with the rest of the code.

Two additional comments from my side:
1. I think it would be nice to also plot the stats above the background. This option already exists in stats_mf. Indeed, if in the call to plot_taylor_diagram, I replace "prior" by "prior_above_BC" and "posterior" by "posterior_above_BC", the code works. Only the colors do not fully work. Maybe you can adjust the code slightly to incorporate this option.

2. I think it would be interesting to distinguish the different sites in the Taylor diagram. In this way, we can easily spot the most problematic sites. This would create a huge legend when plotting 20 sites, but it can be a global legend for all subplots. We could use for example default colors for the sites, different symbols for the models, and with/without filling for prior/posterior.

Let me know what you think.

melodb:
> I think there is an issue with the notebook `taylor_plots.ipynb` did you upload it by mistake ?
> 
> If you want to show all the different plots that you can do, it can be nice to keep it as well.

In my view, this notebook can be removed. As far as I understand, it was only meant for testing, right @jthanwer?

scripts/taylor_plot.ipynb
jthanwer: The notebook was indeed included by mistake.

fluxy/plots/mf_stats.py
jthanwer: I'll take care of this in a different PR if I have time. Right now, I think it is consistent with other functions in fluxy.

fluxy/plots/mf_stats.py
jthanwer: Agree if this comment targets the class only, and not the plot_taylor_diagram function. I think we should also remove the default value for `sd_range`.

jthanwer:
The taylor_plot jupyter notebook should not have been committed. I will remove it in the next commit.

fluxy/plots/mf_stats.py
jthanwer: I don't really agree here. I would prefer this function to work (i.e., plot something) even if the user does not include these parameters. We can discuss it IRL.

fluxy/plots/mf_stats.py
jthanwer: This parameter should be called `stat_markers`, not `model_markers`, because it is the marker style applied to a specific stat (i.e. prior or posterior).

fluxy/plots/mf_stats.py
lionel42: @melodb @jthanwer  In the case of argument, I also prefer to have default values. It eases a lot the processes of using functions, because you can specify only the data needed and everything else is just adjustments you can do in a later step.
I already face issues with fluxy because so many things were needed in the config kwargs that i took me long frustrating minutes filling all the required values, just to have basic things like color and labels working, before being able to see a simple basic plot.

tests/test_plots.py
```diff
@@ -243,6 +244,29 @@ def test_plot_stats():
     )
 
 
+def test_plot_taylor_diagram():
+    ds_all_allsites = slice_mf(
+        ds_all_mf.copy(),
```
lionel42, line 249: Should not need to copy here, as the slice mf is not supposed to change the input data. (at least from what I know)
If you copied from other tests, then the other tests are wrong.

fluxy/plots/mf_stats.py
```diff
@@ -107,3 +283,228 @@ def plot_stats_mf(
     )
 
     return fig
+
+
+def plot_taylor_diagram(
+    stats: dict[Literal["prior", "posterior", "prior_above_BC", "posterior_above_BC"], pd.DataFrame],
+    model_colors: dict[str, str],
+    model_labels: dict[str, str],
```
lionel42, line 291: Following the other discussion about defautl parameters, here you can try to put empty dict as default and below in the code when you need these arguments set some default values. 

I think that would greatly improve the usability.

fluxy/plots/mf_stats.py
melodb: Oh, I see. My goal was to avoid wrong plots of being created. When doing real work, I prefer not to have a plot outputted, than to have a plot with wrong units or other wrong things that the user forgot to set up accordingly. Sometimes these type of mistakes are not immediatly spotted. But I understand that that's a bit annoying from a testing point of view. I think you can do as you prefer in this case. We can have latter a broader discussion on what's best for fluxy and adapt all functions accordingly.

melodb (approved):
I have tested it and it's working really well. Thanks for implementing this and for all the changes. I'm happy for it to be merged!

fluxy/plots/mf_stats.py
jthanwer: I think this should be part of a dedicated pull request, as we would need to modify all functions. Using a decorator to set up default values and check data would be a good idea.

tests/test_plots.py
jthanwer: I copied from other tests so I guess we should change the other tests as part of a more general pull request, or maybe this copying is relevant.

lionel42 (approved):

PR #156, opened 2025-07-02
Inlet height plotting options

mo-aliceramsden:
Updates to allow for intake height to be chosen when plotting (as well as site):

- Added `get_inlet_height` which reads inlet height info from sites_info.json. I've set this so that it reads in the heightest height available for each site, (which I think is the default for most sites more recently). This info can be used if inlet height isn't included in the platform names.
- Updated `edit_vars_and_attributes` to add inlet height to the dataset, if a model is missing this information. There's a warning message, telling users this is happening. This leads to long error messages for models with lots of sites, so maybe could be improved!
- Added option to slice by intake height as well as site, and only plot data from this height.
- Updated `mf_timeseries` plots to include height in the y axis.
- Added option to `plot_sites_timeseries` to separate site by intake height, so we can see when these intake height switches are happening.

To do:
- Update stats plots so stats could be plotted for each intake height separately. If this would be a useful feature?
- Fix tests so they work with these changes.

There's also a couple of other minor fixes to unrelated functions:
- `plot_sites_timeseries` would fail if only one model was included (at some point it uses the number of models to set x axis spacing, and this failed when len(models) = 1). So I've put a quick fix in for this.
- If any of the mf plotting functions can't find a variable named percentile_ it will instead use stdev_, and vice versa. This way we can plot a mixture of models that use only one of these variables. I've put a warning in the function output, telling the user if this is happening.

### DIFF
--- a/fluxy/io.py
+++ b/fluxy/io.py
@@ -756,6 +756,7 @@ def edit_vars_and_attributes(
                 all_inlet = np.hstack((all_inlet,
                                        np.ones(np.where(ds['number_of_identifier']==site_index)[0].shape[0])*site_inlet))
             ds = ds.assign(intake_height=("index",all_inlet))
+            logger.warning(f'intake_height not present in {model} so assuming default maximum intake heights for all sites.')
 
         # Test that the number of identifiers had valid values
         max_num_id, min_num_id = (
--- a/fluxy/io.py
+++ b/fluxy/io.py
@@ -15,7 +15,7 @@ from typing import Literal
 
 from fluxy import config
 from fluxy.operators.regions import extract_region_flux
-from fluxy.operators.select import slice_flux
+from fluxy.operators.select import slice_flux,get_intake_height,get_site_index
 from fluxy.operators.flux_align_dataset import align_time
 
 logger = logging.getLogger(__name__)
@@ -242,7 +242,8 @@ def read_model_output(
 
         # Fix variables and attributes
         ds_all[m] = edit_vars_and_attributes(
-            ds_all[m], m, period[i], file_type, config_data.get("regions_info", {})
+            ds_all[m], m, period[i], file_type, config_data.get("regions_info", {}), 
+            config_data.get("site_info", {})
         )
 
     return ds_all
@@ -532,6 +533,7 @@ def edit_vars_and_attributes(
     frequency: str,
     file_type: Literal["flux", "concentration"],
     regions_info: dict[str, str],
+    site_info: dict[str,dict],
 ) -> xr.Dataset:
     """
     Edit dataset variables and attributes.
@@ -744,6 +746,16 @@ def edit_vars_and_attributes(
             ds = ds.assign(
                 assimilation_flag=("index", np.ones(ds["index"].size, dtype=int))
             )
+
+        if 'intake_height' not in ds:
+            # Add inlet height if not present
+            all_inlet = np.array([])
+            for site in ds['platform'].values:
+                site_inlet = get_intake_height(site,site_info)
+                site_index = get_site_index(ds,site)
+                all_inlet = np.hstack((all_inlet,
+                                       np.ones(np.where(ds['number_of_identifier']==site_index)[0].shape[0])*site_inlet))
+            ds = ds.assign(intake_height=("index",all_inlet))
 
         # Test that the number of identifiers had valid values
         max_num_id, min_num_id = (
--- a/fluxy/io.py
+++ b/fluxy/io.py
@@ -747,6 +747,7 @@ def edit_vars_and_attributes(
                 assimilation_flag=("index", np.ones(ds["index"].size, dtype=int))
             )
 
+        '''
         if 'intake_height' not in ds:
             # Add inlet height if not present
             all_inlet = np.array([])
@@ -757,7 +758,7 @@ def edit_vars_and_attributes(
                                        np.ones(np.where(ds['number_of_identifier']==site_index)[0].shape[0])*site_inlet))
             ds = ds.assign(intake_height=("index",all_inlet))
             logger.warning(f'intake_height not present in {model} so assuming default maximum intake heights for all sites.')
-
+        '''
         # Test that the number of identifiers had valid values
         max_num_id, min_num_id = (
             ds["number_of_identifier"].max(),
--- a/fluxy/io.py
+++ b/fluxy/io.py
@@ -15,7 +15,7 @@ from typing import Literal
 
 from fluxy import config
 from fluxy.operators.regions import extract_region_flux
-from fluxy.operators.select import slice_flux,get_intake_height,get_site_index
+from fluxy.operators.select import slice_flux, get_intake_height, get_site_index
 from fluxy.operators.flux_align_dataset import align_time
 
 logger = logging.getLogger(__name__)
@@ -242,8 +242,12 @@ def read_model_output(
 
         # Fix variables and attributes
         ds_all[m] = edit_vars_and_attributes(
-            ds_all[m], m, period[i], file_type, config_data.get("regions_info", {}), 
-            config_data.get("site_info", {})
+            ds_all[m],
+            m,
+            period[i],
+            file_type,
+            config_data.get("regions_info", {}),
+            config_data.get("site_info", {}),
         )
 
     return ds_all
@@ -533,7 +537,7 @@ def edit_vars_and_attributes(
     frequency: str,
     file_type: Literal["flux", "concentration"],
     regions_info: dict[str, str],
-    site_info: dict[str,dict],
+    site_info: dict[str, dict],
 ) -> xr.Dataset:
     """
     Edit dataset variables and attributes.
@@ -746,18 +750,7 @@ def edit_vars_and_attributes(
             ds = ds.assign(
                 assimilation_flag=("index", np.ones(ds["index"].size, dtype=int))
             )
-        '''
-        if 'intake_height' not in ds:
-            # Add inlet height if not present
-            all_inlet = np.array([])
-            for site in ds['platform'].values:
-                site_inlet = get_intake_height(site,site_info)
-                site_index = get_site_index(ds,site)
-                all_inlet = np.hstack((all_inlet,
-                                       np.ones(np.where(ds['number_of_identifier']==site_index)[0].shape[0])*site_inlet))
-            ds = ds.assign(intake_height=("index",all_inlet))
-            logger.warning(f'intake_height not present in {model} so assuming default maximum intake heights for all sites.')
-        '''
+
         # Test that the number of identifiers had valid values
         max_num_id, min_num_id = (
             ds["number_of_identifier"].max(),
--- a/fluxy/io.py
+++ b/fluxy/io.py
@@ -746,7 +746,7 @@ def edit_vars_and_attributes(
             ds = ds.assign(
                 assimilation_flag=("index", np.ones(ds["index"].size, dtype=int))
             )
-        
+        '''
         if 'intake_height' not in ds:
             # Add inlet height if not present
             all_inlet = np.array([])
@@ -757,7 +757,7 @@ def edit_vars_and_attributes(
                                        np.ones(np.where(ds['number_of_identifier']==site_index)[0].shape[0])*site_inlet))
             ds = ds.assign(intake_height=("index",all_inlet))
             logger.warning(f'intake_height not present in {model} so assuming default maximum intake heights for all sites.')
-        
+        '''
         # Test that the number of identifiers had valid values
         max_num_id, min_num_id = (
             ds["number_of_identifier"].max(),
--- a/fluxy/io.py
+++ b/fluxy/io.py
@@ -746,8 +746,7 @@ def edit_vars_and_attributes(
             ds = ds.assign(
                 assimilation_flag=("index", np.ones(ds["index"].size, dtype=int))
             )
-
-        '''
+        
         if 'intake_height' not in ds:
             # Add inlet height if not present
             all_inlet = np.array([])
@@ -758,7 +757,7 @@ def edit_vars_and_attributes(
                                        np.ones(np.where(ds['number_of_identifier']==site_index)[0].shape[0])*site_inlet))
             ds = ds.assign(intake_height=("index",all_inlet))
             logger.warning(f'intake_height not present in {model} so assuming default maximum intake heights for all sites.')
-        '''
+        
         # Test that the number of identifiers had valid values
         max_num_id, min_num_id = (
             ds["number_of_identifier"].max(),
--- a/fluxy/operators/select.py
+++ b/fluxy/operators/select.py
@@ -102,7 +102,7 @@ def slice_mf(
     data_dir: os.PathLike | None = None,
     mf_units_print: str = None,
     keep_unassimilated: bool = False,
-    intake_height: int | None = None,
+    intake_height: float | None = None,
 ) -> dict[str, xr.Dataset]:
     """
     Slices down the mole fraction timeseries data, to within the
@@ -252,7 +252,7 @@ def slice_site(ds: xr.Dataset, site: str) -> xr.Dataset:
     return ds
 
 
-def slice_height(ds: xr.Dataset, intake_height: str) -> xr.Dataset:
+def slice_height(ds: xr.Dataset, intake_height: float) -> xr.Dataset:
     """
     Slices the dataset to only include data for a given intake height.
 
@@ -324,7 +324,7 @@ def get_unique_sites(ds_all: dict[str, xr.Dataset]) -> list[str]:
     return sites
 
 
-def get_intake_height(site: str, site_info: dict[str:dict]) -> int | None:
+def get_intake_height(site: str, site_info: dict[str, dict]) -> int | None:
     """
     Extract the inlet height from site_info.json.
     This function is used to insert the 'inlet' variable if this is missing.
@@ -340,56 +340,60 @@ def get_intake_height(site: str, site_info: dict[str:dict]) -> int | None:
             Maximum height from all networks and inlets available at the site.
     """
 
-    all_heights = []
-
-    if site in site_info:
-
-        for network in site_info[site].keys():
-            if "height" in site_info[site][network].keys():
-                all_heights += site_info[site][network]["height"]
-
-        if all_heights == []:
-            logger.warning(
-                f"No height info available for {site} in site_info.json so using 0m."
-            )
-            max_height = 0
-
-        else:
-            max_height = np.max([int(h.strip("m")) for h in all_heights])
-
-    else:
+    if site not in site_info:
         logger.warning(
             f"No height info available for {site} in site_info.json so using 0m."
         )
+
         max_height = 0
+
+        return max_height
+
+    all_heights = []
+
+    for network in site_info[site].keys():
+        if "height" in site_info[site][network].keys():
+            all_heights += site_info[site][network]["height"]
+
+    if not all_heights:
+        logger.warning(
+            f"No height info available for {site} in site_info.json so using 0m."
+        )
+
+        max_height = 0
+
+        return max_height
+
+    max_height = np.max([int(h.strip("m")) for h in all_heights])
 
     return max_height
 
+
 def get_unique_site_height_pairs(
     ds_all: dict[str, xr.Dataset],
-    siteList: list[str],
-    separate_by_height: bool = False
-    ) -> list[tuple]:
+    site_list: list[str],
+    separate_by_height: bool = False,
+) -> list[tuple]:
     """
     Finds all unique pairs of site-height from all datasets.
-    e.g. if one dataset has MHD-10, TAC-185, RGL-90 and one dataset has 
+    e.g. if one dataset has MHD-10, TAC-185, RGL-90 and one dataset has
     MHD-10, TAC-100, this function will return MHD-10, TAC-100, TAC-185 and RGL-90.
-    
+
     Args:
         ds_all:
             Dictionary of datasets with mf data from all models.
-        siteList:
+        site_list:
             List of unique sites from all models.
         separate_by_height:
             If True, includes a tuple per site height, if False, returns a tuple per
             site, with the second index set to None for all sites,
             e.g. [('MHD',None),('TAC',None)]
     Returns:
-        siteList:
+        site_list:
             Pairs of sites and heights, e.g. [('MHD',10),(TAC,100),(TAC,185)]
     """
-    
-    if separate_by_height == True:
+
+    if separate_by_height:
 
         unique_pairs = set()
 
@@ -409,13 +413,13 @@ def get_unique_site_height_pairs(
 
             unique_pairs.update(pairs)
 
-        siteList = sorted(unique_pairs)
+        site_list = sorted(unique_pairs)
 
     else:
-        siteList = list(zip(siteList, [None] * len(siteList)))
-    
-    
-    return siteList
+        site_list = list(zip(site_list, [None] * len(site_list)))
+
+    return site_list
+
 
 def clean_timeseries_missing_data(
     ds: xr.Dataset,
--- a/fluxy/operators/select.py
+++ b/fluxy/operators/select.py
@@ -102,6 +102,7 @@ def slice_mf(
     data_dir: os.PathLike | None = None,
     mf_units_print: str = None,
     keep_unassimilated: bool = False,
+    intake_height: int | None = None,
 ) -> dict[str, xr.Dataset]:
     """
     Slices down the mole fraction timeseries data, to within the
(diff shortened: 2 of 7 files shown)


fluxy/plots/mf_timeseries.py
```diff
@@ -245,9 +258,14 @@ def plot_mf_timeseries(
...
-            f'{species_info["species_print"]} {site} ({plot_units[0]})'
-        )
+        if intake_height is not None:
+            ax[iax, 0].set_ylabel(
+                f'{species_info["species_print"]} {site} {intake_height}m ({plot_units[0]})'
+            )
+        else:
+            ax[iax, 0].set_ylabel(
+                f'{species_info["species_print"]} {site} ({plot_units[0]})'
+            )
```
melodb: Maybe here you can use the if-statement just to define the extra label and leave everything eles outside. For example:

```
height_label = ''
if intake_height is not None:
      height_label = f'-{intake_height}m'
ax[iax, 0].set_ylabel({species_info["species_print"]} {site}{height_label} ({plot_units[0]}))
```

fluxy/plots/mf_timeseries.py
```diff
@@ -309,21 +328,47 @@ def plot_sites_timeseries(
...
     fig, ax = plt.subplots(1, 1, figsize=(0.7 * len(siteList), 8))
 
     assert margin < 0.5, "Margin must be smaller than 0.5"
     assert margin > 0, "Margin must be positive"
 
-    model_offset = (1 - 2 * margin) / (len(models) - 1)
+    if len(models) > 1:
+        model_offset = (1 - 2 * margin) / (len(models) - 1)
+    else:
+        model_offset = 1
```
melodb, line 363: Thank you!!! This bug was annoying me so much but I was always postponing fixing it.

fluxy/io.py
```diff
@@ -745,6 +747,17 @@ def edit_vars_and_attributes(
...
+            # Add inlet height if not present
+            all_inlet = np.array([])
+            for site in ds['platform'].values:
+                site_inlet = get_intake_height(site,site_info)
+                site_index = get_site_index(ds,site)
+                all_inlet = np.hstack((all_inlet,
+                                       np.ones(np.where(ds['number_of_identifier']==site_index)[0].shape[0])*site_inlet))
+            ds = ds.assign(intake_height=("index",all_inlet))
+            logger.warning(f'intake_height not present in {model} so assuming default maximum intake heights for all sites.')
+
```
melodb: This is working fine. However, it fails if site_info.json does not exist or if one of the sites is not present in site_info.json. In my view, we should not force the user to provide a complete (or any) site_info file. One option is to check if site_info exists before using it as well as each site. However, this makes me wonder if this bit of code is really needed. If the user does not provide the `intake_height` directly in the dataset, we will by default assign one single height per site. Therefore, the timeseries plots will only differ in the ylabel. So I don't think we gain much, specially because this height (taken as the max in site_info) might be incorrect. Taking into account that `intake_height` is marked as "optional" in the template, I would simply check if the variable exists inside `slice_height` before trying to apply the mask and raise an error if the variable does not exist.

fluxy/operators/select.py
```diff
@@ -194,11 +195,20 @@ def slice_mf(
                 logger.warning(f"Error slicing site {site} from {m}: {e}")
                 ds_all.pop(m)
                 continue
-
+            
+        # Slice according to intake height
+        if intake_height is not None:
+            try:
+                ds_all[m] = slice_height(ds_all[m],intake_height)
```
melodb: Even though you have this try/except block, maybe inside slice_height you can check if intake_height exists in the dataset and raise an error if it does not (see example in L.254-255 in this file).

scripts/PARIS_inversion_results.ipynb
```diff
@@ -262,7 +263,7 @@
     "\n",
     "ds_all_mf_sliced = slice_mf(ds_all_mf.copy(),start_date,end_date,site,baseline_site=baseline_site,\n",
     "                              baseline_filename=baseline_filename,data_dir=data_dir,\n",
-    "                              mf_units_print=mf_units_print)\n",
+    "                              mf_units_print=mf_units_print,intake_height=intake_height)\n",
```
melodb, line 266: Add flag `separare_by_height` in the call to `plot_sites_timeseries`

fluxy/plots/mf_timeseries.py
```diff
@@ -309,21 +328,47 @@ def plot_sites_timeseries(
...
 
     models = ds_all.keys()
     dt_start_date = np.datetime64(start_date)
     dt_end_date = np.datetime64(end_date)
     siteList = get_unique_sites(ds_all)
     model_labels_copy = model_labels.copy()
-
+    
+    # create list of grouped site-height pairs
+    if separate_by_height == True:
```
melodb: Maybe here you can check if intake_height exists in all datasets before undergoing the separation by height.

fluxy/plots/mf_timeseries.py
```diff
@@ -309,21 +328,47 @@ def plot_sites_timeseries(
...
+                site_ids = np.where(ds['number_of_identifier'] == get_site_index(ds, site))
+                all_heights,all_heights_id = np.unique(ds['intake_height'][site_ids],return_index=True)
+                all_heights = ds['intake_height'][site_ids][np.sort(all_heights_id)]
+                all_heights = all_heights[~np.isnan(all_heights)]
+                for h in all_heights:
+                    siteHeightList.append(f'{site}-{int(h)}')
+        siteHeightList = np.sort(np.unique(siteHeightList))
+        siteHeightList_print = [f"{i.split('-')[0]}\n{i.split('-')[1]}m" for i in siteHeightList]
+        siteList = np.array([i.split('-')[0] for i in siteHeightList])
+        heightList = np.array([float(i.split('-')[1]) for i in siteHeightList])
```
melodb: Maybe this can go to a function that returns a list of tuples with the unique site/intake_height pairs. ChatGPT suggested something like this which worked on my end:
```
unique_pairs = set()

for m, ds in ds_all_mf.items():
    # Get mapping from identifier number to platform name
    platform_names = ds['platform'].values
    number_ids = ds['number_of_identifier'].values
    intake_heights = ds['intake_height'].values

    # Map number_of_identifier to platform name
    platforms_for_obs = platform_names[number_ids]

    # Pair up platform with intake height
    pairs = zip(platforms_for_obs, intake_heights)

    unique_pairs.update(pairs)

siteList = sorted(unique_pairs)
```
Then you could loop over siteList in a way that works both for a list of str and a list of tupples. For example:
```
for i, item in enumerate(siteList):
    if isinstance(item, tuple):
        site, height = item
    else:
        site = item
        height = None
```
Or define siteList always as a tuple with `height=None` if `separate_by_height=False` or intake_height does not exist. Just suggestions to simplify this a bit.

melodb (commented):
Thanks a lot for this! I left some comments that might simplify the code a bit. See what you think.

melodb:
Regarding the stats, maybe the "get_unique_sites_intake_height" function that I suggest can be called inside `stats_mf` so that `sites_all` becomes a list of tupples. Then it would be just a matter of masking accordingly and adding "height" to the `stats_site` dictionary. Just an idea, I did not try it out.

fluxy/io.py
mo-aliceramsden: I've made the `intake_height` variable a requirement for using `slice_height` and set an error if the user tries to use the intake_height option when this variable isn't present. This does mean that we can only use these plotting options with future versions of our concentrations files, as currently no model includes the `intake_height` variable, but as you say, this makes this more transparent than assuming the max height from the info file

fluxy/operators/select.py
```diff
@@ -194,11 +195,20 @@ def slice_mf(
                 logger.warning(f"Error slicing site {site} from {m}: {e}")
                 ds_all.pop(m)
                 continue
-
+            
+        # Slice according to intake height
+        if intake_height is not None:
+            #try:
```
hdelongueville: Is keeping this try - except commented section (lines 201, 203-206) necessary or can it be removed?

fluxy/operators/select.py
```diff
@@ -292,6 +327,36 @@ def get_unique_sites(ds_all: dict[str, xr.Dataset]) -> list[str]:
...
+    This assumes use of the heighest height from all available at the specified site.
+    
+    Args:
+        site (str):
+            3-letter site code.
+        site_info (dict of dict):
+            Data extracted from site_info.json.
+    Returns:
+        max_height (int):
+            Maximum height from all networks and inlets available ast s
```
hdelongueville: typo at the end of the sentence?

fluxy/plots/mf_timeseries.py
```diff
@@ -189,7 +190,19 @@ def plot_mf_timeseries(
...
+                    if 'percentile' in unc_var:
+                        unc_var_in = unc_var
+                        unc_var = unc_var.replace('percentile','stdev')
+
+                    elif 'stdev' in unc_var:
+                        unc_var_in = unc_var
+                        unc_var = unc_var.replace('stdev','percentile')
+
+                    if unc_var not in ds_plot.keys():
+                        KeyError(f"Variables {unc_var_in} and {unc_var} not found in {m}.")
```
hdelongueville: Missing **raise** KeyError(...)

fluxy/plots/mf_timeseries.py
```diff
@@ -189,7 +190,19 @@ def plot_mf_timeseries(
...
+                    else:
+                        print(logger.warning(f"Variable {unc_var_in} not found in {m} so reading uncert from {unc_var}."))
```
hdelongueville: Could you remove the unnecessary **print()** around the logger call?

hdelongueville (changes requested):
Thanks a lot for the work! I've made some comments. In another PR, Lionel suggested me to ask Copilot for a review, which ended to be super useful to spot small bugs. I recommend!
Once this PR is ready to be merged, I recommend applying black formatter before merging :)

fluxy/plots/mf_timeseries.py
mo-aliceramsden: Thanks for this suggestion, that's very helpful! I've added that simplification in to the code

mo-aliceramsden:
> Regarding the stats, maybe the "get_unique_sites_intake_height" function that I suggest can be called inside `stats_mf` so that `sites_all` becomes a list of tupples. Then it would be just a matter of masking accordingly and adding "height" to the `stats_site` dictionary. Just an idea, I did not try it out.

Good idea! I'll make a note to work on this next week (as I'm on leave for the rest of the week) and use your recommendation to help with the stats plot updates. But if this PR needs to be merged in before then, please feel free to merge this in and I can work on the stats plot update in a separate PR.

mo-aliceramsden:
> Thanks a lot for the work! I've made some comments. In another PR, Lionel suggested me to ask Copilot for a review, which ended to be super useful to spot small bugs. I recommend! Once this PR is ready to be merged, I recommend applying black formatter before merging :)

Thanks for your review and for reminding me about the black formatter! I've run that over the two files where I made any changes (and will hopefully remember to run for any future changes I make...)

fluxy/plots/mf_timeseries.py
```diff
@@ -317,13 +348,42 @@ def plot_sites_timeseries(
...
+            platform_for_obs = platform_names[number_ids]
+
+            pairs = zip(platform_for_obs, intake_heights)
+
+            unique_pairs.update(pairs)
+
+        siteList = sorted(unique_pairs)
+
+    else:
+        siteList = zip(siteList, [None] * len(siteList))
```
melodb: I would move all this code inside a new function e.g. get_unique_site_height_pairs(ds_all, separate_by_height). You can put this function in operators/select.py
Moreover, inside the `else`, you need to convert the zip into a list so that the code works.

fluxy/plots/mf_timeseries.py
```diff
@@ -317,13 +348,42 @@ def plot_sites_timeseries(
...
     assert margin < 0.5, "Margin must be smaller than 0.5"
     assert margin > 0, "Margin must be positive"
 
-    model_offset = (1 - 2 * margin) / (len(models) - 1)
+    if len(models) > 1:
+        model_offset = (1 - 2 * margin) / (len(models) - 1)
+    else:
+        model_offset = 1
 
     for site_iter, site in enumerate(siteList):
```
melodb: Maybe you can change the for-loop to:
`for site_iter, (site,height) in enumerate(siteList):`
and use `site` and `height` directly in the code instead of `site[0]` and `site[1]`

fluxy/plots/mf_timeseries.py
```diff
@@ -363,8 +425,11 @@ def plot_sites_timeseries(
         dt_start_date - np.timedelta64(1, "D"), dt_end_date + np.timedelta64(1, "D")
     )
 
-    ax.set_xticks(np.arange(siteList.size))
-    ax.set_xticklabels(siteList)
+    ax.set_xticks(np.arange(len(siteList)))
+    if separate_by_height == True:
+        ax.set_xticklabels([f"{s[0]}\n{int(s[1])}m" for s in siteList])
```
melodb: Here you can also consider using `for (site,height) in siteList`

scripts/PARIS_inversion_results.ipynb
```diff
@@ -248,6 +248,8 @@
     "### edit variables in this block\n",
     "species = 'hfc134a' \n",
     "site = 'MHD'\n",
+    "intake_height = None\n",
+    "separate_by_height = False      #If True, plots \n",
```
melodb: Move `separate_by_height` to the next notebook cell. As far as I understand it is only relevant for `plot_sites_timeseries`

melodb (commented):
Thanks for this once again. I have added some small comments. After the fix the tests pass succesfully and the code can be merged from my side.

melodb:
I have added @lionel42 in case he has something to add.

mo-aliceramsden:
> Thanks for this once again. I have added some small comments. After the fix the tests pass succesfully and the code can be merged from my side.

Thanks for taking another look at this and for recommending some good fixes (this process is definitely helping to improve my python!) I've got the tests passing now, so will merge this in (and make a separate issue about the stats plotting by site-height) once I've had confirmation from the others that this can be merged.

melodb (approved):

fluxy/io.py
```diff
@@ -744,7 +746,18 @@ def edit_vars_and_attributes(
...
+            # Add inlet height if not present
+            all_inlet = np.array([])
+            for site in ds['platform'].values:
+                site_inlet = get_intake_height(site,site_info)
+                site_index = get_site_index(ds,site)
+                all_inlet = np.hstack((all_inlet,
+                                       np.ones(np.where(ds['number_of_identifier']==site_index)[0].shape[0])*site_inlet))
+            ds = ds.assign(intake_height=("index",all_inlet))
+            logger.warning(f'intake_height not present in {model} so assuming default maximum intake heights for all sites.')
+        '''
```
lionel42: If you add commented code, there is no real need. Better would be to delete this commented code

fluxy/operators/select.py
```diff
@@ -102,6 +102,7 @@ def slice_mf(
     data_dir: os.PathLike | None = None,
     mf_units_print: str = None,
     keep_unassimilated: bool = False,
+    intake_height: int | None = None,
```
lionel42: Why is it integer and not float ?

fluxy/operators/select.py
```diff
@@ -250,6 +252,35 @@ def slice_site(ds: xr.Dataset, site: str) -> xr.Dataset:
...
+
+    Args:
+        ds (xarray dataset):
+            Dataset with mf data of a given model.
+        intake_height (str):
+            intake_height of interest.
+    Returns:
+        ds (xarray dataset):
+            Dataset with mf data of a given model, sliced to only include data for the given intake_height.
+    """
```
lionel42: Also here why is the inlet height a string and not a float ?

fluxy/operators/select.py
```diff
@@ -293,6 +324,99 @@ def get_unique_sites(ds_all: dict[str, xr.Dataset]) -> list[str]:
...
+
+    all_heights = []
+
+    if site in site_info:
+
+        for network in site_info[site].keys():
+            if "height" in site_info[site][network].keys():
+                all_heights += site_info[site][network]["height"]
+
+        if all_heights == []:
```
lionel42: It is no usual to compare to the empty list, better use `if not all_heights`

fluxy/operators/select.py
```diff
@@ -293,6 +324,99 @@ def get_unique_sites(ds_all: dict[str, xr.Dataset]) -> list[str]:
...
+        else:
+            max_height = np.max([int(h.strip("m")) for h in all_heights])
+
+    else:
+        logger.warning(
+            f"No height info available for {site} in site_info.json so using 0m."
+        )
+        max_height = 0
+
+    return max_height
```
lionel42: Usually when testing bad conditions. I prefer the following.

```
if condition_not_met:
    return value
# proceed with the code
if second_condition:
    ... 

```
This usually help the code to be more flat and easier to read.

fluxy/operators/select.py
```diff
@@ -293,6 +324,99 @@ def get_unique_sites(ds_all: dict[str, xr.Dataset]) -> list[str]:
...
+        separate_by_height:
+            If True, includes a tuple per site height, if False, returns a tuple per
+            site, with the second index set to None for all sites,
+            e.g. [('MHD',None),('TAC',None)]
+    Returns:
+        siteList:
+            Pairs of sites and heights, e.g. [('MHD',10),(TAC,100),(TAC,185)]
+    """
+    
+    if separate_by_height == True:
```
lionel42: When you test a boolean, you can just do `if separate_by_height:`

fluxy/operators/select.py
```diff
@@ -293,6 +324,99 @@ def get_unique_sites(ds_all: dict[str, xr.Dataset]) -> list[str]:
...
+            number_ids = ds["number_of_identifier"].values
+            intake_heights = ds["intake_height"].values
+
+            platform_for_obs = platform_names[number_ids]
+
+            pairs = zip(platform_for_obs, intake_heights)
+
+            unique_pairs.update(pairs)
+
+        siteList = sorted(unique_pairs)
```
lionel42: Use snake_case_naming of variables

fluxy/plots/mf_timeseries.py
```diff
@@ -189,7 +191,22 @@ def plot_mf_timeseries(
...
+                        unc_var = unc_var.replace("stdev", "percentile")
+
+                    if unc_var not in ds_plot.keys():
+                        raise KeyError(
+                            f"Variables {unc_var_in} and {unc_var} not found in {m}."
+                        )
+                    else:
+                        logger.warning(
+                            f"Variable {unc_var_in} not found in {m} so reading uncert from {unc_var}."
+                        )
```
lionel42, line 209: If you raise anyway you can remove the else statement and unindent the logger warning

fluxy/plots/mf_timeseries.py
```diff
@@ -245,9 +262,15 @@ def plot_mf_timeseries(
             )
 
         # Set timeseries y-axis label and legend
+
+        height_label = ""
+        if intake_height is not None:
+            height_label = f"-{intake_height}m"
```
lionel42, line 268: use a one line if statement

fluxy/plots/mf_timeseries.py
```diff
@@ -363,8 +403,11 @@ def plot_sites_timeseries(
         dt_start_date - np.timedelta64(1, "D"), dt_end_date + np.timedelta64(1, "D")
     )
 
-    ax.set_xticks(np.arange(siteList.size))
-    ax.set_xticklabels(siteList)
+    ax.set_xticks(np.arange(len(siteList)))
+    if separate_by_height == True:
+        ax.set_xticklabels([f"{s}\n{int(h)}m" for (s,h) in siteList])
+    else:
+        ax.set_xticklabels([s for (s,h) in siteList])
```
lionel42: Better to create the labels variable in the if statement and calling the set_tickslabels method after the if statement

lionel42 (changes requested):
Nice update, since I have seen you want to improve your python code, I have made several comments how to improve the code.


We push to have the code formatted with the `black` formatter.
Can you try it on the python files ? 

`pip install black`

and then format the file by running 

`black /path/to/fluxy/io.py`  


Also I don't understand (and did not find in the discussion) what the type of the inlet height should be (str or float ?) 



Do you want to try to add some unit test file for the inlet height also ? 
I think that would be useful to check that the code runs.

fluxy/operators/select.py
mo-aliceramsden: Thanks for this tip! This does make things a lot easier to read without all the if/else statements!

mo-aliceramsden:
> Nice update, since I have seen you want to improve your python code, I have made several comments how to improve the code.
> 
> We push to have the code formatted with the `black` formatter. Can you try it on the python files ?
> 
> `pip install black`
> 
> and then format the file by running
> 
> `black /path/to/fluxy/io.py`
> 
> Also I don't understand (and did not find in the discussion) what the type of the inlet height should be (str or float ?)
> 
> Do you want to try to add some unit test file for the inlet height also ? I think that would be useful to check that the code runs.

Thanks for your review, all your comments have been very helpful! And thanks for checking on the type of the `inlet_height` variable, I've fixed that as float throughout now.

I've add a basic test for the slice_height function, which requires the `inlet_height` variable to be present in a concentrations netcdf, so I've added a new format InTEM data file which contains that variable, and is used in this test. When other models add in these variables to their files, they can be added to the test too.

lionel42:
@mo-aliceramsden Thanks for adding the tests.

If you look at the size of the file in the test folder, you will notice that they are much smaller than the ones you updated. 
We reduce the size on purpose so that it works faster with git.
There is a file called reduce size in the script folder.

Can you try using it and reducing the test file size ?

Or maybe we could simply add the inlet height to one of the existing netcdf test file ? 

@melodb @mo-aliceramsden I will let you decide the best option

mo-aliceramsden:
> @mo-aliceramsden Thanks for adding the tests.
> 
> If you look at the size of the file in the test folder, you will notice that they are much smaller than the ones you updated. We reduce the size on purpose so that it works faster with git. There is a file called reduce size in the script folder.
> 
> Can you try using it and reducing the test file size ?
> 
> Or maybe we could simply add the inlet height to one of the existing netcdf test file ?
> 
> @melodb @mo-aliceramsden I will let you decide the best option

Thanks for pointing me to the script that reduces the file size, that's very useful. I've used that to reduce the file size significantly, and the tests still pass

fluxy/plots/mf_timeseries.py
```diff
@@ -363,8 +403,12 @@ def plot_sites_timeseries(
         dt_start_date - np.timedelta64(1, "D"), dt_end_date + np.timedelta64(1, "D")
     )
 
-    ax.set_xticks(np.arange(siteList.size))
-    ax.set_xticklabels(siteList)
+    ax.set_xticks(np.arange(len(site_list)))
+    if separate_by_height:
+        xticklabels = [f"{s}\n{int(h)}m" for (s, h) in site_list]
+    else:
+        xticklabels = [s for (s, h) in site_list]
```
lionel42: Now that I see it like this, you could move the if else directly in the list comprehension.

fluxy/test_utils/models.py
```diff
@@ -30,3 +33,22 @@ def get_loaded_models(
...
+
+    ds_all_mf = read_model_output(
+        data_dir,
+        file_type,
+        species="hfc134a",
+        models=test_models_new,
+        config_data=config_data,
+    )
+
+    return ds_all_mf
```
lionel42: This creates a duplicated function. Maybe you could add the list of models as a kwarg in the existing function ?

fluxy/test_utils/models.py
```diff
@@ -11,6 +11,9 @@
     "ELRIS_NAME_EUROPE_EDGAR",
 ]
 
+test_models_new = [
+    "InTEM_NAME_EUROPE_EDGAR"
+]
```
lionel42: New is not very nice name, maybe "with_inlet" ? 

Also I am not sure why you did not simply add add the model to the existing list of model ?

lionel42:
@mo-aliceramsden I just posted a few last comments and then should me good from my side.

fluxy/plots/mf_timeseries.py
```diff
@@ -309,23 +339,31 @@ def plot_sites_timeseries(
...
             Dictionary with model lables.
+        separate_by_height (bool):
+            If True, separates obs by intake height and by site.
     """
 
     models = ds_all.keys()
     dt_start_date = np.datetime64(start_date)
     dt_end_date = np.datetime64(end_date)
-    siteList = get_unique_sites(ds_all)
+    site_list = get_unique_sites(ds_all)
```
melodb: Small suggestion: maybe you can move `site_list = get_unique_sites(ds_all)` inside function `get_unique_site_height_pairs` instead of passing it as input. Variable `site_list` will be updated anyway inside the function.

fluxy/test_utils/models.py
mo-aliceramsden: Thanks for this suggestion, I've removed that duplicated function and renamed the list of models with the `inlet_height` variable, so these can be read in as needed for this test.

melodb:
@lionel42 and @hdelongueville, can this be merged?

lionel42 (approved):

lionel42:
> @lionel42 and @hdelongueville, can this be merged?

Good for me I had a quick look again and it seems very good

hdelongueville (approved):

Issue #87, opened 2025-03-26
Merge Similar Flux Functions


hdelongueville:
There are several functions and files related to flux time series and flux maps that share similar functionality. To reduce redundancy, consolidate or merge these similar functions into a single 'flux' function.

- [ ] Merge and optimize `fluxy.operators.flux_map_resample.py` and `fluxy.operators.flux_timeseries_resample.py` to a single file `fluxy.operators.flux_resample.py`. (e.g. using .groupby instead of .resample?)
- [x] Split `fluxy.operators.flux_prepare_map_data.prepare_data` to a new function `fluxy.operators.align_dataset.align_model` and a new function in `fluxy.operators.flux_combine.py`
- [x] Rename `fluxy.operators.flux_prepare_map_data.py` to `fluxy.operators.flux_map_diff.py`

hdelongueville:
`flux_combine.py` has two similar functions `combine_dataset` and `combine_map_dataset`. `combine_dataset`, which is currently flux timeseries specific, should be modified to be flexible enough to be merged with `combine_map_dataset`.

PR #157, opened 2025-07-09
Update flux map to plot only active sites and include flux uncertainties

### LINKED ISSUE #153: Use sites list from concentration files in flux maps
Currently, the list of sites to be plotted on the flux maps is extracted from the `sites` attribute in the flux NetCDF file via the `get_sites_coordinates` function. If the sites attribute is missing, an alternative list from another NetCDF file is used, when available.

Since the sites attribute is not mandatory, the `add_sites=True` option only works when using InTEM, which includes the sites attribute. To improve flexibility, a new option allows `get_sites_coordinates` to accept a `fallback_sites` list.

An additional potential enhancement would be to extract site information from the concentration NetCDF, for which this info is mandatory. This file is read during the read_model_output call. The site information could then be stored as an attribute and passed forward, with `slice_flux` adjusted to include only sites that were "active" during the specified time period.

hdelongueville:
**Summary**
This PR updates the flux mapping workflow to:

- Include only active observation sites in the map
- Resample flux datasets rather than averaging
- Prepare for displaying flux error reduction in maps --> THIS PART HAS BEEN ABORTED BECAUSE THIS METRIC HAS BEEN FOUND UNSUITABLE

**Changes**

1. Add `sites` variable to flux dataset
- The `read_model_output` function can now also read the concentration file when `file_type="flux"` and `add_sites_to_flux=True`.
- This adds a new sites variable to the flux dataset, used for plotting active sites.
- `add_sites_to_flux` default is `False` to avoid unnecessary file reads for country-level plots.
- New function added: `fluxy.io.add_sites_var`

2. Switch from averaging to resampling for flux maps

- Changed time/model "averaging" of fluxes to proper resampling.
- The `sites` variable is now grouped (not averaged) to include all active sites across models/periods.
- Uncertainty resampling now mirrors the method used in `flux_timeseries_resample.calculate_resampled_uncertainty`, enabling future uncertainty map plots.
- Reorganised `flux_map.py` to remove any calculation from the plotting loop.
- New functions added:
`fluxy.plots.utils.get_active_sites_coordinates`
`fluxy.operators.flux_map_diff.make_diff_ds`
`fluxy.operators.flux_map_resample.group_sites`
`fluxy.operators.flux_map_resample.calculate_resampled_uncertainties`
`fluxy.operators.flux_map_resample.calculate_resampled_flux`
`fluxy.operators.flux_map_resample.calculate_resampled_dataset`

- [x] Close #153 

3. Plot map uncertainties --> ABORTED

4. Cleanup
- [x] Depracate `fluxy.plots.utils.get_sites_coordinates`
- [x] Depracate `fluxy.plots.operators.flux_map_resample.get_flux_mean`
- [x] Apply black formatter
- [x] Update Notebook
- [x] Update Annexes
- [x] Update tests


hdelongueville:
I still need to 
- remove depracated `fluxy.plots.utils.get_sites_coordinates` and `fluxy.plots.operators.flux_map_resample.calculate_flux_mean`
- apply black formatter 
- update the annexes with `add_sites_to_flux`

I was planning to do this after your review.

Also, note that the restructuring for plotting the error reduction is no longer strictly necessary. However, I kept some of the changes because they result in cleaner code. At least now fluxy is better prepared for future error map implementations.

hdelongueville:
fluxy/tests/test_io.py is failing at test_read_mf but I don't know why as I haven't changed anything to this part of the code. 

Could it be related to this comment?
"This test fails sometimes when runned wil all the other tests because of a xarray cache problem"

If so, how can I overcome this?

lionel42:
> fluxy/tests/test_io.py is failing at test_read_mf but I don't know why as I haven't changed anything to this part of the code.
> 
> Could it be related to this comment? "This test fails sometimes when runned wil all the other tests because of a xarray cache problem"
> 
> If so, how can I overcome this?

Yes definitely, i have changed the netcdf reader to one that is thread safe, now you should not have this wierd issue anymore.

fluxy/operators/flux_map_resample.py
```diff
@@ -5,67 +5,188 @@
...
+#         )
+
+#         if season not in seasonal_means.season.values:
+#             raise ValueError(f"Season '{season}' not found in the dataset.")
+
+#         ds_output = seasonal_means.sel(season=season)
+#     ds_output.attrs["start_date"] = data.time.values.min()
+#     ds_output.attrs["end_date"] = data.time.values.max()
+
+#     return ds_output
```
lionel42: Just one comment, because i saw you removed the test for this function.

It is not a good idea to remove a function. As I understand, you removed it because you implemented the same capabilities in another function.
If people start to write scripts with the `calculate_mean_flux` their scripts won't work any more

In this case I would reimplement the `calculate_mean_flux` function by calling the new function you created (and keep the test ;) ).

lionel42:
@hdelongueville  Can you also request a copilot review ? In my experience it helps quite a lot to have a first review at the code.

fluxy/operators/flux_map_resample.py
hdelongueville: That's a very good point, thanks!

hdelongueville: I've also renamed some functions, should I make a note of the changes?
e.g. `average_over_months` --> `resample_over_months` bcs this function is not only doing averaging anymore and thought the old name was misleading.

lionel42: Good question 

in case the function has a confusing name and we want to remove it, we should deprecate it.
this means that anyone using the function will get a deprecation warning message:

You can implement something like this.
```
def new_function(x, y):
    return x + y

def old_function(x, y):
    warnings.warn(
        "'old_function' is deprecated...",
        DeprecationWarning,
        stacklevel=2  # <- points to the user's call site
    )

    return new_function(x, y)
```

hdelongueville: That's good info, thanks!

hdelongueville:
> @hdelongueville Can you also request a copilot review ? In my experience it helps quite a lot to have a first review at the code.

I think I need Copilot Premium to do that, which I don't have unfortunately.

lionel42:
> > @hdelongueville Can you also request a copilot review ? In my experience it helps quite a lot to have a first review at the code.
> 
> I think I need Copilot Premium to do that, which I don't have unfortunately.

I did it for you ;) 
In case you are interested, they give free licences for academics. 

You can go though the comments and see which ones make sense to you

hdelongueville:
Thanks! I think the PR is ready now. :)

fluxy/operators/flux_map_resample.py
```diff
@@ -25,30 +29,151 @@ def get_flux_mean(
...
+
+    Returns:
+        xr.DataArray: 
+            Flux averaged over each group.
+    """
+    da = flux.groupby(groups).mean(dim="time", keep_attrs=True)
+
+    return da
+
+def calculate_resampled_uncertainties(
```
hdelongueville, line 66: As brought in a discussion with Alex, I should add the option to resample the uncertainties by calculating the mean. Similarly to what is done for country fluxes, with the option resample_uncert_correlation = False #If True, uses mean uncertainty during resampling, if False, recalculates uncertainty assuming no correlation.

hdelongueville: Ideally, we'd have to join flux_timeseries_resample and flux_map_resample (see https://github.com/openghg/fluxy/issues/87) because some functions are doing the same thing but with different format inputs. That's for a futur PR..

fluxy/io.py
```diff
@@ -792,3 +802,91 @@ def edit_vars_and_attributes(
...
+            Dictionary with settings read from json file.
+
+    Returns:
+        ds_flux (xarray dataset):
+            xarray dataset with a new 'sites' variable added if the concentration
+            file exists. If not, returns the original ds_flux unchanged.
+    """
+
+    # Derive the path for the concentration file
+    filepath_conc = filepath_flux.with_name(filepath_flux.stem + '_concentrations.nc')
```
lionel42: I think it is possible in the format that  the flux file ends with `_flux`  or with nothing.
This works only if the flux file ends with nothing right ?

fluxy/io.py
```diff
@@ -792,3 +802,91 @@ def edit_vars_and_attributes(
...
+        valid_times = ds_conc["time"].where(mask, drop=True)
+
+        if frequency == "yearly":
+            mf_keys = valid_times.dt.year.values
+        elif frequency == "monthly":
+            mf_keys = list(zip(valid_times.dt.year.values, valid_times.dt.month.values))
+            mf_keys = np.array(mf_keys, dtype=[('year', 'i4'), ('month', 'i4')])
+
+        # Mark time steps in flux where observations from this site exist
+        sites.sel(platform=site)[:] = np.isin(flux_keys, mf_keys).astype(int)
```
lionel42: Don't use sel when assigning values. Use loc or where.
(see warning there: https://docs.xarray.dev/en/latest/generated/xarray.DataArray.sel.html)

tests/test_io.py
```diff
@@ -17,8 +17,10 @@ def test_read_flux(model):
 
     period = "yearly"
 
+    add_sites_to_flux = True
+
     ds_all_flux = read_model_output(
-        data_dir, "flux", species, [model], config_data, period=period
+        data_dir, "flux", species, [model], config_data, period=period, add_sites_to_flux=add_sites_to_flux
```
lionel42: Would be nice to keep the test also with `add_sites_to_flux = False`

tests/test_plots.py
```diff
@@ -52,7 +53,7 @@
     )
 else:
     ds_all_flux = read_model_output(
-        data_dir, "flux", species, models, config_data, period=period
+        data_dir, "flux", species, models, config_data, period=period, add_sites_to_flux=add_sites_to_flux
     )
```
lionel42: Also here, keep the test with `add_sites_to_flux = False`.
Usually when you use tests like this, you don't modify the previous tests, but simply add new ones with the new features.
The goal is to preserve backward compatibility.

scripts/PARIS_inversion_results.ipynb
```diff
@@ -810,7 +811,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.10.6"
+   "version": "3.13.1"
```
lionel42: avoid changing notebooks metatdata

scripts/PARIS_inversion_results.ipynb
```diff
@@ -74,7 +74,7 @@
     "presentation_mode = False\n",
     "\n",
     "### Group the models of interest in meaningful experiment names \n",
-    "experiments = {'std_run'       :['InTEM_NAME_EUROPE_EDGAR_std','ELRIS_NAME_EUROPE_EDGAR_std','RHIME_NAME_EUROPE_EDGAR_std'],\n",
+    "experiments = {'std_run'       :['InTEM_NAME_EUROPE_EDGAR', 'ELRIS_NAME_EUROPE_EDGAR', 'RHIME_NAME_EUROPE_EDGAR'],\n",
```
lionel42: Why change this ? We had an issue discussing this problem #160 , maybe a separate PR would be more appropriate

fluxy/test_utils/models.py
```diff
@@ -27,6 +27,7 @@ def get_loaded_models(
         species="hfc134a",
         models=test_models,
         config_data=config_data,
+        add_sites_to_flux = True,
```
lionel42: Same comment as for the test files, better not change exisiting tests

fluxy/operators/flux_combine.py
```diff
@@ -37,24 +37,42 @@ def combine_dataset(
...
+            p0 = v.isel(percentile=0).min(dim="model", keep_attrs=True)
+            p1 = v.isel(percentile=1).max(dim="model", keep_attrs=True)
+            ds_combined[var] = xr.concat([p0, p1], dim="percentile", combine_attrs="override")
+
+        elif var == "sites" and dims == {"time", "platform"}:
+            ds_combined[var] = v.any(dim="model", keep_attrs=True).astype(int)
+
+        elif "time" in dims:
+            ds_combined[var] = v.mean(dim="model", keep_attrs=True)
+
```
lionel42: In this version, it seems that some variables might not be given in the output. (no else condition)

lionel42: Also you could reduce some duplicate code by specifying the kwargs only once

fluxy/operators/flux_map_resample.py
lionel42, line 1: This resample file has many function which are very similar (resample over a time period).

It would be nice to simplify them in a way that we keep only the one function in the file which decide which of the others to call.

However this seems to be a lot of work that should be done in a separate PR.

lionel42 (changes requested):
Few changes needed for the code.

Make sure you keep the tests with the old parameters and add new ones, instead of modifying the old version

fluxy/io.py
hdelongueville: Yes it only works if the flux file ends with nothing.
I don't think the filename convention allows for flux files to end with `_flux` though.

hdelongueville: @melodb do you know if flux filename ending with _flux is allowed?

melodb: No. Even though I don't like it, currently the flux file should end with nothing and the concentration file should end with "_concentration".

tests/test_plots.py
hdelongueville: Does it mean I need to duplicate the test and add the new option?
Can I not use @pytest.mark.parametrize("add_sites_to_flux", [True, False]) or best behaviour is to "duplicate"?

lionel42: You are right, parametrize would be a very elegant option.

hdelongueville: Actually, this test_plots.py file is a bit painful to work with pytest with all the global variables in it. Using pytest would need to modify most of the functions..

I'd like to set something like
@pytest.mark.parametrize("add_sites_to_flux, add_sites", [(True, True),(False, False)])
but the issue is that add_sites_to_flux needs to be called in read_model_output which is outside a function. 

Do you know if there is a simple way to do this?
 
If not, I'll just define ds_all_flux_with_sites = read_model_output(
        data_dir, "flux", species, models, config_data, period=period, add_sites_to_flux=True
    )

fluxy/operators/flux_combine.py
hdelongueville: > In this version, it seems that some variables might not be given in the output. (no else condition)

Indeed, I don't include variables which don't have a time dimension. I've added a logger.info message to specify if a variable has been excluded when combining datasets over time.

hdelongueville: > Also you could reduce some duplicate code by specifying the kwargs only once

Thanks for the advice! I’m not in the habit of using kwargs, but I’ll try to use them more often.

scripts/PARIS_inversion_results.ipynb
hdelongueville: This change happened without me noticing.
Is there a way to avoid changing notebooks metadata "unintentionally"?

fluxy/plots/flux_map.py
```diff
@@ -118,48 +124,62 @@ def plot_flux_map(
...
+        try:
+            sites_info = (
+                get_active_sites_coordinates(ds, config_data, fallback_sites) if add_sites else ""
+            )
+        except Exception as e:
+            raise RuntimeError(
+                "Failed to get active sites coordinates. "
+                "Check that `add_sites_to_flux` is True in `read_model_output` "
+                "or that a `fallback_sites` list is provided in `plot_flux_map`."
+            ) from e
```
melodb: Is this try/except block needed? It seems to me that the main problems are already checked inside get_active_sites_coordinates

melodb (approved):
I left only a small comment. Thank you for all this effort! I confess that I don't really follow all the post-processing needed for the flux maps, but it seems very organized :)
I tested the notebook and it is working fine.

fluxy/plots/flux_map.py
hdelongueville: Yes, you're totally right! I've removed it.

scripts/PARIS_inversion_results.ipynb
lionel42: Yes, this is done automatically in vscode by the python extension.

Usually when you add the files to the commit, you can see which changes were done and revert the changes before committing. 
Maybe there is an option to force vscode to not change this, but I don't know it.

tests/test_plots.py
lionel42: My bad, I think I haven't read properly where this is applied.

Since this is a change in the global variables, I would keep things as they are and simply add new tests. 

* one test in io for the add_sites_to_flux option
* one new test for the plot, where you can load the data directly in the function 
* revert the original test code to what it was, so we don't change existing tests

I know that this might create duplicates and very similar code, but for unit tests it is generally not a problem.

hdelongueville: Thanks for your reply! I just have one comment/question. In the original test, there is the option add_sites = True which will now fail. I could add a fallback list of sites so that it works almost like before, or I could set add_sites = False, but that would mean slightly changing the test.

lionel42: that is the goal of the test, you have to make it pass without changing the inputs ;) 

Think that the behaviour should be the same as the one before you implemented the new option

Issue #146, opened 2025-06-16
Sectorial decomposition of fluxes


lionel42:
In many applications it is important to split the fluxes in different sectors (ex GNFR sectors) 

Do we already have a convention in how to do this ? 

My idea would be to simply add a sector dimension and expand the variables that require sectorialization to this dimension.

Another cleaner option, for not having to mess with variables having different possibilities for the dimensions would be to create new variables adding a sectorial suffix.

melodb:
That option is already considered in the new format. The name tag "total" (e.g. flux_total_prior) can be replaced by a sector name (e.g. agriculture, fossil). I did not add this to README_templates because fluxy is not yet ready to deal with this. @mo-aliceramsden is working on an implementation which would scale the posterior fluxes according to the prior sector fluxes. Maybe we can wait for her developments to start with.

lionel42:
so this means you want to create one variable for each sector ? 
Why not using a dimension for sectors and avoid the multiplication of variables ?

Also do you plan on allowing only specific sectors any name ?
This might create issues with the naming convention.

melodb:
That's a very fair point. According to the template, one must specify one variable per sector. Maybe @ayanu can clarify the reasons behind this choice.

Moreover, we have not agreed on specific sector names yet. That's something we can easily align within the PARIS project, but won't be applicable to all. I guess this would be a problem regardless of using one variable per sector or creating a sector dimension.

stiglers-eponym:
The sectorial decomposition is of high interest for us. From my experience I want to emphasize the importance of providing uncertainty correlations for the sectors. I cannot see how to include this information in the current data format. A dimension for sectors could make it easier to include such information.

Beyond the correlation estimates, we also estimated averaging kernel matrices for the sector estimates that turned out to be important for the interpretation of the results. If the data format will be extended to include the sectors, maybe one could check that the format will allow for an extension to include averaging kernel matrices in the future. Again, when using a sector dimension, this should not be difficult.

melodb:
We have discussed this with @ayanu and for now we would prefer to keep the template as is. Nevertheless, inside fluxy, we can reorganize the variables in a better way by including a sector dimension.

[...]

PR #159, opened 2025-07-10
Sector post processing to create sector flux timeseries and spatial maps

mo-aliceramsden:
**Sector-level inversion output created by scaling total fluxes by sector flux proportions from a prior (e.g. EDGAR).**

The changes below all assume we're keeping the current flux output format, with each sector as a separate variable, e.g. `flux_agriculture_prior`. There is discussion over including `sector` as a dimension #146, but this is undecided, so for now I've set this up to work with the current file format.

The framework for this is mostly complete, but there's a few things left to finish. I'm on holiday 11th-18th July, so I can add in any suggested updates/changes when I'm back.

Changes made:
- Added a sector flux proportions file (inside the `sector_proportions` directory on the ICOS jupyterhub) and a template for this. The example version of this file just contains EDGAR data, but I've added a variable that allows you to select different versions of this file (so we can use other inputs from priors, inventories etc.)
- New functions in `flux_scale_by_sector_proportions` to scale `flux_total_prior` and `flux_total_posterior` by the sector proportions, to produce sector spatial fluxes. And then sum fluxes over countries to create country sector flux totals. This is basic so far and doesn't include any scaling of uncertainties, or inclusion of uncertainty covariance in the summing of spatial fluxes. I'm not sure how best to do this, as we don't have any uncertainty info on the spatial sector fluxes.
- `cell_area` is needed to recalculate country totals from the sector spatial fluxes, but this variable isn't present in all models' output, so I've added a EUROPE-specific example `cell_area` file, with a template for this too.
- Updated `regions` extraction functions, so sector-level fluxes can be extracted.
- Updated `flux_timeseries` and `flux_map` plotting functions so one sector from all models can be plotted at the same time. 
- Added stacked bar chart plot of all sector fluxes, from each model.

What I need to add:
- Inventory sector-level data, to include in flux timeseries and stacked bar chart plots of all sectors. We currently have this info for the UK and the Netherlands, I just need to process it into the correct format.
- Fix tests to work with sector options.

Things to discuss (then I can add these updates to the PR):
- How to calculate sector-level flux uncertainties, for both spatial and country fluxes?
- Is the stacked bar chart plot the best way to show the timeseries of country sector fluxes for all models?
- Do we want to include spatial fluxes at the `inversion_grid` resolution? Currently, the code doesn't do this (it only uses sector info at the resolution of the prior), so I've put in some warnings if the user tries to plot as sector other than `total` with an `inversion_grid` variable.


lionel42:
for me the most critical point which we discussed a bit in #146 is the extensibility of this to other sectorial decomposition.
Currently we need a flux variable for each sector and the name of the sector has to be included in the variable name.

I am not a huge fan of this solution, since as you mention I would prefer to add a sector dimension to the existing variables, so that we don't need to change our variables definitions.

If we still want to have a variable for each sector, we need to add functionalities to read an arbitrary number of sectors name, or add a nc variable that contains all sectors.

lionel42:
I had a quick look to the code. In the new functions there is a lot of duplicated code, which would be nice to clean. I am happy to do a more detailed code review once this is done.

fluxy/plots/flux_timeseries.py
```diff
@@ -244,12 +244,7 @@ def plot_country_flux(
         units = {ds.posterior.units for ds in ds_all.values()}
     else:
         units = {ds["flux_total_posterior_country"].units for ds in ds_all.values()}
-    if len(units) == 1:
-        unit = list(units)[0]
-    else:
-        raise ValueError(
-            f"Inconsistency in the units from the different datasets : {units} are present. Only one is expected."
-        )
+    unit = list(units)[0]
```
melodb: Is there a reason to remove this check? As far as I understand, it was meant to avoid plotting datasets with different units. This can happen if the user sets `country_flux_units_print=None` and the datsets don't have consistent units.

fluxy/plots/flux_timeseries.py
```diff
@@ -178,6 +177,7 @@ def plot_country_flux(
     plot_resample_and_original: bool = False,
     return_res: bool = False,
     rolling_mean: bool = False,
+    sector: str | None = None
```
alexdanjou: I think that it would be better to put default as "total" (for consistency with other functions and also to pass test tests.test_plots.plot_sites_timeseries). The function is currently not working when sector is set to None

data/templates/PAR-AVE-EYE_sector_proportions.cdl
```diff
@@ -0,0 +1,53 @@
...
+		flux_total:units = "g m-2 -s" ;
+		flux_total:long_name = "total flux from all sources" ;
+	double lat(lat) ;
+		lat:_FillValue = NaN ;
+	double lon(lon) ;
+		lon:_FillValue = NaN ;
+	int64 time(time) ;
+		time:units = "days since 1980-01-01 00:00:00" ;
+		time:calendar = "proleptic_gregorian" ;
+	float flux_proportion_total(time, lat, lon) ;
```
alexdanjou: I don't understand what is inside this variable. Could you be more descriptive?

data/templates/PAR-AVE-EYE_sector_proportions.cdl
```diff
@@ -0,0 +1,53 @@
...
+		flux_proportion_total:_FillValue = NaNf ;
+		flux_proportion_total:longname = "total flux as a proportion of total flux" ;
+	float flux_waste(time, lat, lon) ;
+		flux_waste:_FillValue = NaNf ;
+		string flux_waste:EDGAR_sectors = "SWD_LDF","SWD_INC","WWT" ;
+	float flux_proportion_waste(time, lat, lon) ;
```
alexdanjou: Is this variable necessary? If it's `flux_waste/flux_total` I think it should be removed for simplicity. The proportion could them be calculated in the fluxy. Same for other sectors

fluxy/plots/flux_timeseries.py
```diff
@@ -483,3 +479,279 @@ def plot_country_flux(
...
+    fig, axes = plt.subplots(
+        n_rows,
+        n_cols,
+        sharex=True,
+        constrained_layout=True,
+        figsize=(n_cols * 6, n_rows * 4),
+    )
+
+    row_count = 0
+    col_count = 0
```
alexdanjou: variables `row_count` and `col_count` are not used.

fluxy/plots/flux_timeseries.py
```diff
@@ -483,3 +479,279 @@ def plot_country_flux(
...
+
+    country_equivalent = {
+                "NW_EU2": "NW EUROPE",
+                "CW_EU": "CENTRAL W EUROPE",
+                "NW_EU_CONTINENT": "NW CONTINENTAL EUROPE",
+            }
+    print_country = country_equivalent.get(plot_region, plot_region)
+
+    # Create figure
+    if plot_inventory_or_prior == 'inventory':
```
alexdanjou: Could that be in determine subplots_arrangement for consistency and conciseness?

fluxy/plots/flux_timeseries.py
```diff
@@ -483,3 +479,279 @@ def plot_country_flux(
...
+                    ds_to_plot[m_extract].prior,
+                    label=sector.title(),
+                    color=sector_colors[sector],
+                    bottom = total_s_comp,
+                    alpha=0.7,
+                    width=width
+                    )
+                total_s_comp += ds_to_plot[m_extract].prior
+                                    
+        if resample != None:
```
alexdanjou, line 678: Those stuff (650-666) tackles the same issues as lines 411-433 as far as I understand. A function set_time_ticks common to both `plot_country_flux` and `plot_country_sector_flux_bar` that take the time points and return xticks and xtickslabels would be interesting. I will avoid duplicated code and also make those two plot consistent in term of display.

I can help on that (and subplots_arrangement) if you agree with my comments but don't want to do it :)

alexdanjou (changes requested):
Good job. I think more bits of code should be shared between plot_country_flux and plot_country_sector_flux_bar. I can help on that if you don't have time

fluxy/plots/flux_timeseries.py
alexdanjou: Same for below with ax_comp

data/templates/PAR-AVE-EYE_sector_proportions.cdl
mo-aliceramsden: Very good point! I've removed the 'flux_proportion_sector' variables from this file and moved the proportion calculation into fluxy.

fluxy/plots/flux_timeseries.py
mo-aliceramsden: Good idea! This isn't vital for getting this PR into development, so I've created an issue which explains this improvement, then if one of us has some time in the future, we can add that update in.

lionel42 (changes requested):
Since there are many new feature, we should also add some unit tests for new plot functions.

scripts/PARIS_inversion_results.ipynb
```diff
@@ -762,7 +836,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.10.6"
+   "version": "3.1.undefined"
   }
```
melodb: Remove version changes in notebook if possible.

scripts/PARIS_inversion_results.ipynb
```diff
@@ -135,6 +139,14 @@
     "    for m in models:\n",
     "        ds_all_flux_scaled[m] = slice_flux({m:ds_all_flux[m]},config_data,start_date,end_date,species=species,country_flux_units_print=country_flux_units_print)[m]\n",
     "\n",
+    "if sector_file is not None:\n",
+    "    for m in models:\n",
+    "        ds_all_flux_scaled[m] = scale_by_sector_proportions(data_dir=data_dir,ds=ds_all_flux_scaled[m],\n",
```
melodb: Can you move the for-loop inside the function and input a dictionary of datasets instead? Just to simplify the notebook.

fluxy/operators/flux_scale_by_sector_proportions.py
```diff
@@ -0,0 +1,191 @@
...
+        ds (xarray dataset):
+            Input ds, with added flux_sector_prior and flux_sector_posterior variables.
+            If regions is not None, also contains region/country sector total variables.
+    """
+    
+    if create_region_sector_totals == True:
+        
+        if 'country_fraction' not in ds:
+            raise ValueError(
+                f"country_fraction variable not present in dataset for {ds['inversion_system']} "
```
melodb: Is `ds['inversion_system']` a mistake? In other comment of mine, I have suggested to make this function accept a dictionary of datasets instead of one dataset at a time. If you decide in this way, you will have access to the dictionary keys which you can use here.

scripts/PARIS_inversion_results.ipynb
```diff
@@ -600,7 +658,17 @@
     "for m in models:\n",
     "    ds_all_flux_scaled[m] = slice_flux({m:ds_all_flux[m]},config_data,start_date,end_date,species=species,flux_units_print=flux_units_print)[m]\n",
     "\n",
-    "model_labels = set_model_labels(models,config_data,get_labels_from_file)"
+    "model_labels = set_model_labels(models,config_data,get_labels_from_file)\n",
+    "\n",
+    "if sector_file is not None:\n",
+    "\n",
+    "for m in models:\n",
+    "    ds_all_flux_scaled[m] = scale_by_sector_proportions(data_dir=data_dir,ds=ds_all_flux_scaled[m],\n",
```
melodb, line 715: Fix indentation.

scripts/PARIS_inversion_results.ipynb
```diff
@@ -121,6 +122,9 @@
     "start_date = '2018-01-01' #inclusive. Option to set as list of dates, e.g. ['2018-01-01','2019-01-01'] which is required for total fgases if one model is missing obs for a year\n",
     "end_date = '2024-01-01' #not inclusive. Option to set as list of dates, e.g. ['2023-01-01','2022-01-01'] which is required for total fgases if one model is missing obs for a year\n",
     "get_labels_from_file = False # If False, uses default labels. If True, uses labels in models_info.json.\n",
+    "sector_file = None #  #if not None, adds sector level results, by scaling model totals by the sector proportions in this file, default: 'EUROPE_EDGAR'\n",
+    "sectors = ['total']      #sectors to add to model ds, using sector_proportions_file, options for:  ['agriculture','waste','energy','industry'] \n",
```
melodb: Is it useful to manually specify the sectors to include? In my view, it would require less typing if we could just select the sector_file and the sectors were taken automatically from the variables in sector_file.

configs/EUROPE_cell_area.nc
melodb, line 1: Why are the units of cell_area set to 1? Should it be "m2"?

fluxy/operators/flux_scale_by_sector_proportions.py
```diff
@@ -0,0 +1,191 @@
...
+            configs_dir = parent_dir / "configs"
+            if 'domain' in ds.attrs:
+                domain = ds.attrs["domain"]
+            else:
+                domain = sector_file.split('_')[0]
+                logger.warning(f'No domain info in dataset attributes, so reading domain from sector_filename: {domain}')
+            with xr.open_dataset(os.path.join(configs_dir,f"{domain}_cell_area.nc")) as f:
+                cell_area = f['cell_area'].values
+                
+        dict_regions: dict[str,str] = config_data['regions_info'].get("regions",{})
```
melodb: Variable dict_regions is not used.

fluxy/operators/flux_scale_by_sector_proportions.py
```diff
@@ -0,0 +1,191 @@
...
+                
+        country_codes = config_data['regions_info'].get("country_codes", {})
+        country_search = [country_codes.get(r, r) for r in regions]
+                
+        #units for spatial flux to regional flux totals
+        molar_mass = config_data["species_info"][species]["molar_mass"]
+        s_in_year = 60*60*24*365
```
melodb: s_in_year not used.

fluxy/operators/flux_scale_by_sector_proportions.py
```diff
@@ -0,0 +1,191 @@
...
+        scaling_factor = get_units_conversion_factor('mol s-1', country_flux_units_print, molar_mass)
```
melodb: Is this scaling_factor used anywhere? It seems to be recalculated in L.145

fluxy/operators/flux_scale_by_sector_proportions.py
```diff
@@ -0,0 +1,191 @@
...
+            Region names. Sector region/country totals are only calculated for
+            these areas.
+        country_flux_units_print (str):
+            Units for country flux, e.g. 'Gg yr-1'
+        config_data (dict):
+            config_data (dict of dict):
+            Dictionary with settings read from json file.
+            Use json filenames as keys.
+        sector_file (str):
+            Start of sector file name, e.g. 'EUROPE_EDGAR'
```
melodb, line 47: Maybe add that this file should be inside folder "sector_flux"?

fluxy/operators/flux_scale_by_sector_proportions.py
```diff
@@ -0,0 +1,191 @@
...
+    sector_prop_path = os.path.join(data_dir,'sector_flux',
+                                       f'{sector_file}_{species}_yearly_flux_sectors.nc')
+
+    sector_prop = {}
+    
+    with xr.open_dataset(sector_prop_path) as f:
+        sector_time = f.time.values
+        for s in sectors:
+            #sector_prop[s] = f[f'flux_proportion_{s}']
+            sector_prop[s] = f[f'flux_{s}']/f[f'flux_total']
```
melodb: "f" is not needed in `"[f'flux_total']`. Moreover, I would remove the commented line above.

fluxy/operators/flux_scale_by_sector_proportions.py
```diff
@@ -0,0 +1,191 @@
...
+            
+    if create_region_sector_totals == True:
+        
+        # create a dictionary of {region_name:[available_region_masks]}  
+        country_all = {}
+
+        for country in country_search:
+            # if a grouped region flux total and grouped country mask is available e.g. InTEM: {BEL-LUX-NLD:[BEL-LUX-NLD]}
+            if country in ds["country_fraction"].country and country in ds['country'].values:
```
melodb: Does it make sense to check the existence of "country" in both `ds["country_fraction"].country` and `ds['country'].values`? Taking into account our data structute, isn't `ds["country_fraction"].country` = `ds['country'].values` by definition?

fluxy/operators/flux_scale_by_sector_proportions.py
```diff
@@ -0,0 +1,191 @@
...
+                country_all[country] = [country]
+                country_id = np.where(ds["country_fraction"].country.values == country)[0]
+                r_fraction = ds['country_fraction'].values[country_id,:,:]
```
melodb: Doesn't `r_fraction = ds['country_fraction'].sel(country = country)` yield directly what you want?

fluxy/operators/flux_scale_by_sector_proportions.py
```diff
@@ -0,0 +1,191 @@
...
+                
+                # if a grouped region flux total but no grouped country mask is available, e.g. RHIME: {BEL-LUX-NLD:[BEL,LUX,NLD]}
+                if np.isnan(r_fraction).all() == True:
+                    logger.warning(
+                        f"{country} country_fraction is not present in ds. "+
+                        f"Considering sum of individual countries: {country_all}."
+                        )
```
melodb: Is the output of "country_all" needed? I get the current status of the variable which is a bit confusing:

```
WARNING:fluxy.operators.flux_scale_by_sector_proportions:BEL-LUX-NLD country_fraction is not present in ds. Considering sum of individual countries: {'DEU': ['DEU'], 'GBR': ['GBR']}.
WARNING:fluxy.operators.flux_scale_by_sector_proportions:BEL-DEU-FRA-GBR-IRL-LUX-NLD country_fraction is not present in ds. Considering sum of individual countries: {'DEU': ['DEU'], 'GBR': ['GBR'], 'BEL': ['BEL'], 'LUX': ['LUX'], 'NLD': ['NLD']}.
```

fluxy/operators/flux_scale_by_sector_proportions.py
```diff
@@ -0,0 +1,191 @@
...
+            else:
+                # if no region flux total and no country mask is available, e.g. ELRIS: {BEL:BEL,LUX:LUX,NLD:NLD}
+                logger.warning(
+                    f"{country} country_fraction is not present in ds. "+
+                    f"Considering sum of individual countries: {country_all}."
+                    )
+                for c_add in country.split('-'):
+                    country_all[c_add] = [c_add]
+            
+    for s_id,s in enumerate(sectors):
```
melodb: s_id is not used so maybe replace by `for s in sectors:`

fluxy/operators/flux_scale_by_sector_proportions.py
```diff
@@ -0,0 +1,191 @@
...
+            # if no early enough dates in sector proportion file, use earliest available
+            elif t.astype('datetime64[Y]').astype('datetime64[ns]') < sector_time[0]:
+                t_index = 0
+            # if no late enough dates in sector proportion file, use latest available
+            elif t.astype('datetime64[Y]').astype('datetime64[ns]') > sector_time[-1]:
+                t_index = -1
+            scaling_factor = sector_prop[s].values[t_index,:,:]
+
+            flux_sector_prior_out[i,:,:] = ds[f'flux_total_prior'].values[i,:,:]*scaling_factor
+            flux_sector_post_out[i,:,:] = ds[f'flux_total_posterior'].values[i,:,:]*scaling_factor
```
melodb: I'm not an expert in xarray but I have the feeling we could do these same operations with xarray and avoid the need for ordered indexes. Maybe @alexdanjou can give some suggestions.

fluxy/operators/flux_scale_by_sector_proportions.py
melodb: Can you move this bit of code upper and join the two if-statements `if create_region_sector_totals == True:`?

fluxy/plots/flux_timeseries.py
```diff
@@ -483,3 +479,272 @@ def plot_country_flux(
...
+                                data_dir: str | None = None,
+                                fix_y_axes: bool = False,
+                                resample: str | list[str] | None = None,
+                                resample_uncert_correlation: bool = False,
+                                rolling_mean: bool = False,
+                                sectors: list[str] = ['agriculture','waste','energy','industry']
+                            ) -> Figure | list:
+    """
+    Timeseries plot of prior and posterior country fluxes, from list of
+    areas in plot_regions.
```
melodb: Update function description

fluxy/plots/flux_timeseries.py
```diff
@@ -483,3 +479,272 @@ def plot_country_flux(
...
+        return_res: Wheter or not including a dictionnary with the results as output
+        rolling_mean : If True, calculates a rolling mean (xx years) for each of the data to plot.
+    Returns:
+        fig: A plot per country/region.
+        res_dict : If return_res, return also a dictionnary containaing the plotted results
+    """
+    sector_colors = {'agriculture':'darkgreen',
+                    'waste':'purple',
+                    'industry':'darkblue',
+                    'energy':'dodgerblue'}
```
melodb: Can these hard-coded colors be moved to config.py?

fluxy/plots/flux_timeseries.py
```diff
@@ -483,3 +479,272 @@ def plot_country_flux(
...
+                "NW_EU2": "NW EUROPE",
+                "CW_EU": "CENTRAL W EUROPE",
+                "NW_EU_CONTINENT": "NW CONTINENTAL EUROPE",
+            }
+
+    country_equivalent = {
+                "NW_EU2": "NW EUROPE",
+                "CW_EU": "CENTRAL W EUROPE",
+                "NW_EU_CONTINENT": "NW CONTINENTAL EUROPE",
+            }
```
melodb: Repeated code.

fluxy/plots/flux_timeseries.py
```diff
@@ -483,3 +479,272 @@ def plot_country_flux(
...
+    
+    if plot_inventory_or_prior == 'inventory':
+        n_plots = len(ds_all.keys())+1
+    elif plot_inventory_or_prior == 'prior':
+        n_plots = len(ds_all.keys())*2
+        
+    n_cols, n_rows = determine_subplots_arrangement(n_plots)
+    
+    units = {ds["flux_total_posterior_country"].units for ds in ds_all.values()}
+    unit = list(units)[0]
```
melodb: Same comment as before regarding checking for units consistency along models.

fluxy/plots/flux_timeseries.py
```diff
@@ -483,3 +479,272 @@ def plot_country_flux(
...
+
+    fig, axes = plt.subplots(
+        n_rows,
+        n_cols,
+        sharex=True,
+        constrained_layout=True,
+        figsize=(n_cols * 6, n_rows * 4),
+    )
+    
+    max_cf = 0
```
melodb: Repeated code?

fluxy/plots/flux_timeseries.py
```diff
@@ -483,3 +479,272 @@ def plot_country_flux(
...
+        end_date: Start dates of the data to plot (used to slice inventory data).
+        annex_mode: If True, replace the labels with more concise versions for National Inventory Report Annexes.
+        scale_co2eq: If True, adapt y-axis label to CO2-eq.
+        plot_inventory: If True, plots inventory flux estimates as bars in each plot.
+        inventory_years: List of inventory data from different years to include. If None, only plots the most recent inventory data.
+        inventory_filename: Name of inventory file: {inventory_filename}_{species}_{inventory_year}
+        data_dir: Path to top data directory, used to read inventory data files.
+        fix_y_axes: If True, uses a consistent y axis for all plots.
+        add_prior: If True, plots prior as dashed lines.
+        add_prior_unc: If True, plots prior uncertainty as shaded area.
```
melodb: Remove args that don't exist in this function.

melodb (commented):
Thanks a lot for working on this! I left some comments mainly regarding some coding details. Regarding function `plot_country_sector_flux_bar`, I tried it out but I did not look at it in much detail. Options resample and rolling_mean are working fine. I just noticed that the y_label is missing when `plot_inventory_or_prior='prior'`. I will take a second look once you address the comments. Thanks a lot once again!

fluxy/operators/flux_scale_by_sector_proportions.py
alexdanjou: Indeed, you can reproduce what you do in your time loop with xarray.DataArray.interp and using fill_value and bounds_error in the kwargs. I'll propose something tomorrow.

scripts/PARIS_inversion_results.ipynb
alexdanjou: I agree with that

scripts/PARIS_inversion_results.ipynb
```diff
@@ -121,6 +122,9 @@
     "start_date = '2018-01-01' #inclusive. Option to set as list of dates, e.g. ['2018-01-01','2019-01-01'] which is required for total fgases if one model is missing obs for a year\n",
     "end_date = '2024-01-01' #not inclusive. Option to set as list of dates, e.g. ['2023-01-01','2022-01-01'] which is required for total fgases if one model is missing obs for a year\n",
     "get_labels_from_file = False # If False, uses default labels. If True, uses labels in models_info.json.\n",
+    "sector_file = None #  #if not None, adds sector level results, by scaling model totals by the sector proportions in this file, default: 'EUROPE_EDGAR'\n",
```
alexdanjou: Shouldn't those three lines be :
```
sector_file = 'EUROPE_EDGAR' #  #if not None, adds sector level results, by scaling model totals by the sector proportions in this file, default: 'EUROPE_EDGAR'
sectors = ['agriculture','waste','energy','industry']      #sectors to add to model ds, using sector_proportions_file, options for:  ['agriculture','waste','energy','industry'] 
create_region_sector_totals = True      #if True, uses country_fraction variable to sum spatial sector fluxes to region sector fluxes
```
to be coherent with the cell under `#### Timeseries of sector country fluxes - as bar charts` ?

alexdanjou: I've pushed this in my commit

fluxy/operators/flux_scale_by_sector_proportions.py
mo-aliceramsden: Thank you for your updates to this Alex, this is so much better now!

In the original function, there was a section that ensured that country sector fluxes were only calculated for the specified list of regions, because calculating this for all countries in the dataset was very slow (taking up to 90 seconds to run). This is much faster when only calculating sector fluxes for the 1-4 regions that will be plotted (only about 20 seconds). So I've added this section back in. There's still a default option available to calculate country sector fluxes for all countries in the datasets, if we need that at any point.

mo-aliceramsden:
> Since there are many new feature, we should also add some unit tests for new plot functions.

Thanks for this recommendation, this is a good idea. I've added some basic tests for the creation of the new sector flux variables and for the bar chart plots.

melodb:
Maybe you would like to reduce the size of the test files before merging.

mo-aliceramsden:
> Maybe you would like to reduce the size of the test files before merging.

Thanks for this reminder! I've reduced the size of the ch4 test files and added a reduced size 'EUROPE_cell_area.nc' file which is now required for the tests to run. All tests are passing for me now.

hdelongueville (approved):
Thanks for all the hard work!

fluxy/plots/flux_timeseries.py
```diff
@@ -261,13 +262,16 @@ def plot_country_flux(
...
         units = {ds["flux_total_posterior_country"].units for ds in ds_all.values()}
+
     if len(units) == 1:
         unit = list(units)[0]
     else:
         raise ValueError(
             f"Inconsistency in the units from the different datasets : {units} are present. Only one is expected."
         )
 
+    unit = list(units)[0]
```
melodb: This is not needed because `unit` is already defiend inside the if.

melodb (approved):

melodb:
As discussed, numpy to xarray optimization will stay for another PR. Moreover, in the future we can also make an attempt to use parts of #162 in sector plotting.

melodb:
@alexdanjou and @lionel42, can this be merged? I think it would be nice to merge this and the eddy branch soon so that we can move on

alexdanjou (approved):
Thanks for this important work!

PR #162, opened 2025-07-25
Adding eddy flux type

lionel42:
* adding a new type of data for eddy flux measurment 
* updated code to read the data 
* added a config for sectorial parameters 
* adding plots 
* modifying some functions used for mf to make them work also for ecflux
* adding tests


lionel42:
@melodb  I don't know if you have time, but if you want to make the review, I am very happy

.gitignore
```diff
@@ -3,4 +3,5 @@
 __pycache__/
 **/__pycache__
 
-data/ne_data/
\ No newline at end of file
+data/ne_data/
+data/my_data/
```
melodb: If folder my_data is created, maybe rename it to something else.

data/templates/README_templates.md
```diff
@@ -95,3 +95,54 @@ The most important variables are described below. Please refer to the cdl files
 | stdev_mf_model                  | mol mol-1 | index      | Model uncertainty of simulated mole fraction
 
 (3) fluxy also accepts ppm, ppb and ppt. However, please make sure that all variables have the same units.
+
+## 3. Eddy covariance flux file
```
melodb, line 99: Maybe you can also add a cdl file to folder templates.

data/tests/ecflux/sectors.yml
```diff
@@ -0,0 +1,94 @@
...
+  Agriculture: "#ff7f00"
+  Fugitives: "#a65628"
+  Shipping: "#FFB5B8"
+
+molar_masses:
+  CO2: 44.01e-3
+  CH4: 16.04e-3
+  N2O: 44.013e-3
+  NOx: 46.005e-3
+  CO: 28.01e-3
```
melodb: Should you get this info from species_info for consistency within the code? Or cross-check if the numbers are the same?

fluxy/types.py
```diff
@@ -0,0 +1,33 @@
...
+
+def file_pattern(file_type: DataTypes) -> str:
+    """Returns the ending pattern for the given file type."""
+    if file_type == DataTypes.FLUX:
+        # Default file type
+        return ".nc"
+    elif file_type == DataTypes.CONCENTRATION:
+        return "_concentrations.nc"
+    else:
+        return f"{file_type.value}.nc"
```
melodb: Should you add "_" before {file_type.value}?

fluxy/io.py
```diff
@@ -87,14 +112,26 @@ def read_config_files(
...
+    # List of files to be read
+    files = itertools.chain(*(configs_dir.glob(f"*{ext}") for ext in read_func.keys()))
 
     # Read json files
     data_dict = {}
-    for file in json_files:
-        data = read_json(file)
+    for file in files:
+        if file.suffix not in read_func:
+            raise ValueError(f"Unsupported config file extension: {file.suffix} in file {file.name}")
```
melodb: Why is this if needed? As far as I understand, in L.127 `files` will correspond only to the files ending in .json/yaml/yml

fluxy/io.py
```diff
@@ -211,33 +248,31 @@ def read_model_output(
...
             xarray dataset read directly from each model's mole fraction netCDF.
     """
+    file_type = DataTypes(file_type)
+
+    if period is None and file_type in (DataTypes.FLUX, DataTypes.CONCENTRATION):
+        # Default period
+        period = "yearly"
 
     if isinstance(period, str):
         period = [period] * len(models)
```
melodb: Maybe here `if isinstance(perior,str)` can be replaced by `if isinstance(perior,str | None)`. In this way, even when `period=None`, it is still transformed into a list and you can keep the if-statement below as well as the usage of period[i]

fluxy/io.py
```diff
@@ -71,6 +71,31 @@ def read_json(filepath: os.PathLike) -> dict[str, dict]:
...
+
+    Args:
+        filepath (str or Path):
+            Path to yaml file including filename.
+    Returns:
+        yaml_data (dictionary of dictionaries):
+            Dictionary with data read from filepath.
+    """
+
+    import yaml
```
melodb: Shouldn't we move this import statement up?

fluxy/io.py
```diff
@@ -87,14 +112,26 @@ def read_config_files(
         # Get location of json files
```
melodb: Update comments refering to "json" files (see L.129) and function/returns description.

fluxy/operators/mf.py
```diff
@@ -145,69 +147,40 @@ def stats_mf(
     type_options = ["prior", "posterior", "prior_above_BC", "posterior_above_BC"]
     assert stats_type in type_options, f"'{stats_type}' is not in {type_options}"
```
melodb, line 140: Update function description.

fluxy/operators/stats.py
```diff
@@ -0,0 +1,125 @@
...
+            stats_site |= {
+                f"q{i:02d}_{simobs}": np.quantile(
+                    sim if simobs == "sim" else obs, i / 100
+                )
+                for i in [5, 25, 75, 95]
+                for simobs in ["sim", "obs"]
+            }
+            stats_site |= {
+                f"{func}_{simobs}": getattr(np, func)(sim if simobs == "sim" else obs)
+                for func in ["max", "min", "mean", "median", "std"]
```
melodb, line 136: Option "std" is already covered above: sd_sim, sd_obs.

melodb (commented):
This is quite a big piece of work! Thanks for adding the eddy flux analysis into fluxy. I have revised the functions that already existed in fluxy. I will take a look at the remaining files later (I still have 10 to go). In addition to my comments, I would propose:
- add umol to units_scale in config.py in case you need to undergo units conversion
- add attribute long_name in your netcdf files so that the netcdf file itself is a bit more descriptive

melodb:
Also thanks for cleaning up the stats function :)

fluxy/operators/stats.py
lionel42: True, but I don't like the sd naming (not consistent with numpy and so on). i am happy to keep both if we can.

.gitignore
lionel42: It is not created, it is more a "convenience" folder where I have put my data.

I would be happy to keep it in the gitignore

data/tests/ecflux/sectors.yml
lionel42: I did it. Just for NOx we have the issue that the there is no real definition of the molar mass.
Also for CO and NOx the gwp is not that easy since they are unstable substances, I left them out.

fluxy/types.py
lionel42: Fixed

fluxy/io.py
lionel42: fixed

fluxy/io.py
lionel42: Since this is an external dependency and I did not want to break everyone's setup, I prefered to keep it here.

i will move it up. I guess we are not that many users that this create troubles

fluxy/operators/stats.py
melodb: Two options:
- remove `sd_sim` and `sd_obs` from above and rename `sd_res` to `std_res`
- rename `sd_sim, sd_obs, sd_res` to `std_sim, std_obs, std_res` and delete option "std" in L.109

In my view, one of these options would be better than having repeated statistics with different names.

lionel42:
@melodb  Thanks for the review, can you have a look to the new code ? Maybe particularly to the doc of the stat function.

fluxy/operators/stats.py
lionel42: Then I would keep only the std. Should I change in the code where this is used ?

fluxy/io.py
```diff
@@ -211,33 +248,31 @@ def read_model_output(
...
-    if isinstance(period, str):
+    if period is None and file_type in (DataTypes.FLUX, DataTypes.CONCENTRATION):
+        # Default period
+        period = "yearly"
+
+    if isinstance(period, str | None):
         period = [period] * len(models)
 
-    if len(period) != len(models):
+    if period is not None and len(period) != len(models):
```
melodb: if condition `period is not None` is not needed anymore.

fluxy/operators/stats.py
melodb: Yes, please, if that's ok for you. Might be only here, in the Taylor plot and the notebook.

fluxy/plots/mf_timeseries.py
```diff
@@ -20,23 +24,52 @@
...
+            kwargs["include"] if "include" in kwargs else default_include,
+        )
+        if "include" in kwargs:
+            kwargs.pop("include")
+    else:
+        # Only kwargs
+        if "include" not in kwargs:
+            kwargs["include"] = default_include
+
+    return plot_timeseries(*args, **kwargs)
```
melodb, line 54: In this type of situations should we print a warning stating that function plot_mf_timeseries is deprecated and that plot_timeseries will be used instead? Just to slowly make the user move to the new function so that in the new release we can delete these legacy functions. I can see that they are useful not to break anyone's code, but I would not like to see the code full of them.

fluxy/plots/mf_timeseries.py
```diff
@@ -191,40 +244,48 @@ def plot_mf_timeseries(
...
-            m,
-            vars_to_plot,
-            diff_include,
-            model_color,
-            presentation_mode,
-            annotate_coords,
-            annotate_index=i,
-            plot_type=plot_type,
-        )
+        if histogram_type != "none":
```
melodb: Shouldn't we also check if `histogram_type = None` when setting the subplot arrangement? Currently, the use of 2 columns is hard-coded in L.152.

fluxy/plots/mf_timeseries.py
```diff
@@ -470,26 +551,26 @@ def plot_histogram(
...
                 xy=annotate_coords[index],
                 xycoords="axes fraction",
-                color=model_color[config.mf_color_index[var]],
+                color=model_color[config.mf_color_index.get(var, 0)],
             )
 
     # Write number of obs
     if plot_type == "separate":
-        n_obs = ds["mf_observed"].count().values
+        n_obs = ds["time"].count().values
```
melodb: I'm afraid this won't work when converting from the old format to the new format. In the old format, variable assimilation_flag does not exist so we set it equal to 1 for all timestamps. However, even though `time` has always a value, some values of mf_observed/etc are NaN. Maybe function io:edit_vars_and_attributes can be updated so that assimilation_flag is set to zero when mf_observed in NaN.

fluxy/operators/ecflux.py
```diff
@@ -0,0 +1,111 @@
+import logging
+from fluxy.types import DsAll
+from fluxy.operators.utils import apply_to_dict_or_single
+import xarray as xr
+
+
+@apply_to_dict_or_single
+def filter_ecflux(
+    ds: xr.Dataset,
```
melodb: Should it be `ds: xr.Dataset | DsAll` to be in agreement with the parameters description? (I understand that at the end the function is actually called with a single dataset at the time via the decorator)

fluxy/operators/ecflux.py
```diff
@@ -0,0 +1,111 @@
...
+    min_friction_velocity :
+        Minimum friction velocity to consider a flux valid, by default 0.2 m/s
+        The friction velocity threshold helps to avoid low turbulence situation
+
+    Returns
+    -------
+    xr.Dataset or DsAll
+        Filtered dataset(s) with the same structure as input.
+    """
+    logger = logging.getLogger(__name__)
```
melodb, line 49: So far we have been defining the logger after the import statements, outside the fucntions. Should we make it consistent?

fluxy/operators/ecflux.py
```diff
@@ -0,0 +1,111 @@
...
+        logger.warning(
+            "No `friction_velocity` found in the dataset. "
+            "Skipping friction velocity filtering."
+        )
+    else:
+        mask &= ds["friction_velocity"] >= min_friction_velocity
+
+    ds_out = ds.where(mask, drop=True)
+
+    print_func = print if verbose else logger.info
```
melodb: Why would you prefer `print()` over `logger.info`? Isn't the latter enough?

fluxy/plots/correlation.py
```diff
@@ -0,0 +1,184 @@
...
+    figsize: tuple[float, float] = (10, 10),
+):
+    """Plot correlation between two timeseries.
+
+    Args:
+        ds_all (dict): Dictionary of xarray datasets for each model.
+        variable (str | tuple):
+            Variable(s) to plot. If `oppose='variables'`,
+            2 variables must be provided as a tuple.
+        include (VariableType): Variables to include/oppose in the plot.
```
melodb: Were you planning to use `include` to be in alignment with `plot_timeseries`? If not, delete it from Args list.

fluxy/plots/correlation.py
```diff
@@ -0,0 +1,184 @@
...
+LimsTuple = tuple[float, float]
+
+
+def plot_correlation(
+    ds_all: dict[str, xr.Dataset],
+    variable: str | tuple[str, str],
+    species: str | None = None,
+    site: str | None = None,
+    oppose: Literal["variables", "models"] = "models",
+    style: Literal["scatter", "density"] = "scatter",
```
melodb, line 25: Add description of style

fluxy/plots/correlation.py
```diff
@@ -0,0 +1,184 @@
...
+
+    if oppose == "variables":
+        if not isinstance(variable, (list, tuple)) or len(variable) != 2:
+            raise ValueError(
+                "When oppose='variables', variable must be a list or tuple of two variable names."
+            )
+        if len(models) != 1:
+            raise ValueError("When oppose='variables', only one model can be provided.")
+    elif oppose == "models":
+        if not isinstance(variable, str):
```
melodb, line 82: In this way, it will break if `variable` is a list or tuple with one single element. Should this be avoided?

fluxy/plots/correlation.py
```diff
@@ -0,0 +1,184 @@
...
+                "When oppose='models', exactly two models must be provided."
+            )
+
+        # Check common time axis
+        ds_x, ds_y = ds_all[models[0]], ds_all[models[1]]
+        is_single_platform = lambda ds: len(ds["platform"]) == 1
+        if not (is_single_platform(ds_x) and is_single_platform(ds_y)):
+            raise ValueError(
+                "When oppose='models', both datasets must have a single platform (site)."
+            )
```
melodb: Can't the user choose the site instead? In this way we can probably use it with our datasets.

fluxy/plots/correlation.py
```diff
@@ -0,0 +1,184 @@
...
+    if oppose == "variables":
+        ds = ds_all[models[0]]
+        x, y = ds[variable[0]].values, ds[variable[1]].values
+        # Remove the nans
+        mask = ~np.isnan(x) & ~np.isnan(y)
+        x, y = x[mask], y[mask]
+        x_label, y_label = variable[0], variable[1]
+    elif oppose == "models":
+        x, y = da_x.values, da_y.values
+        x_label, y_label = models[0], models[1]
```
melodb, line 123: You could move L.80-94 inside this if-statement to separate the verification of the inputs from the data processing.

fluxy/plots/correlation.py
```diff
@@ -0,0 +1,184 @@
...
+    else:
+        raise ValueError("Oppose must be either 'variables' or 'models'.")
```
melodb, line 125: I guess this else is not needed because it would have failed in the previous if-statement.

fluxy/plots/correlation.py
```diff
@@ -0,0 +1,184 @@
...
+            If a single tuple is provided, it is used for both axes.
+            If two tuples are provided, they are used for x and y axes respectively.
+        figsize (tuple[float, float], optional):
+            Size of the figure. Defaults to (10, 10).
+
+    Returns:
+        matplotlib.figure.Figure: The figure object containing the plot.
+    """
+
+    models = list(ds_all.keys())
```
melodb: Instead of assessing everything based on `ds_all.keys()` why not adding a variable like `models_to_plot` (equivalent to `models_to_subtract` in `compute_mf_difference`). In this way the dictionary can have whatever number of models and the user choses what to plot.

fluxy/plots/correlation.py
```diff
@@ -0,0 +1,184 @@
...
+            x_lims, y_lims = lims[0], lims[1]
+        case (lim_min, lim_max), _:
+            # Use the provided limits for both x and y
+            x_lims, y_lims = lims, lims
+        case _, _:
+            raise ValueError(
+                "lims must be a tuple of two tuples or a single tuple."
+                f" Got {lims=} instead."
+            )
+
```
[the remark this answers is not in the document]
melodb: I like its suggestion :)

fluxy/plots/correlation.py
```diff
@@ -0,0 +1,184 @@
...
+            alpha=0.6,
+            s=scatter_size,
+        )
+    elif style == "density":
+        # Calculate the 2D histogram
+        hist, xedges, yedges = np.histogram2d(x, y, bins=100)
+        # Create a meshgrid for the edges
+        X, Y = np.meshgrid(xedges[:-1], yedges[:-1])
+        # Plot the density as a contour plot
+        ax.contourf(X, Y, hist.T, levels=20, cmap="viridis", alpha=0.7)
```
melodb: Maybe change the color scale to something which is white when there is no data. The purple background is a bit ugly.

fluxy/plots/stats.py
```diff
@@ -0,0 +1,126 @@
...
+                        this_x - width / 2,
+                        this_x + width / 2,
+                        color="black",
+                        linewidth=1,
+                    )
+
+        case _:
+            raise ValueError(f"Unknown plot type: {plot_type}")
+
+    ax.set_title(f"Statistics for {species}" if species else "Statistics")
```
melodb: USe species_print in species_info instead

fluxy/plots/stats.py
```diff
@@ -0,0 +1,126 @@
...
+    ax.set_ylabel(variable)
```
melodb: In my view, the info in the y_label should be taken from the data. The user should not be able to choose whatever they want as a label,

fluxy/plots/stats.py
```diff
@@ -0,0 +1,126 @@
...
+import matplotlib.pyplot as plt
+import pandas as pd
+
+
+class StatsPlotTypes(Enum):
+    MEAN_AND_STD = "mean_and_std"
+    BOX_PLOT = "box_plot"
+
+
+def plot_stats(
```
melodb, line 17: Could this plotting options be somehow integrated in `plot_stats_mf` by setting `stats_to_plot=["mean_and_std", "box_plot"]`? I feel that it's pretty much the same data structure. Alternatively, I would try at least to make the style of the plots similar (otherwise they seem to be provided by completely different post-processing codes).

scripts/ec_flux/plot_results.ipynb
```diff
@@ -0,0 +1,371 @@
...
+    "\n",
+    "import matplotlib.pyplot as plt\n",
+    "import xarray as xr\n",
+    "plt.style.use('default')\n",
+    "\n",
+    "\n",
+    "species = 'CO2'\n",
+    "\n",
+    "data_dir = data_dir.parent / 'my_data' / \"eddy_pro\" \n",
+    "config_dir = Path.cwd() / 'configs'\n",
```
melodb: There is no species_info file in this directory. Either move a copy here or move sectors.yml to the main config folder (in this case, maybe rename it to sectors_info_ecflux).

scripts/ec_flux/plot_results.ipynb
```diff
@@ -0,0 +1,371 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "a928d9d9",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "\n",
+    "fig = plot_timeseries(\n",
```
melodb: Maybe add config_data as input to plot_timeseries so that the ylabel shows the species

melodb (commented):
Some more files done. Only the sector functions are missing now. I have made some comments. Moreover:
- as I said in my previous general comment, I would recommend:
   - Add umol to units in config
   - Add longname to the netcdf files
- In my view there is no need to put the EDDY folder inside ecflux. We can treat it as any other model.
- similarly to example_basics, I would rename plot_results to example_basics_ecflux and change data_dir so that it uses the data in data/tests.
- Change "CO2" to "co2" in folders and filenames
- Provocative question: why not changing the yaml file into a json file? (I'm sorry to only raise this after you have changed all the code). There is no problem in supporting both file structures, but I find it a bit more clean if fluxy relies only on one. In this way the user does not need to jump between the rules of one and the other. Consider this with all your love and understanding.

fluxy/operators/sectors.py
```diff
@@ -0,0 +1,81 @@
...
+        return sector_groups
+
+    # Deep copy the dict of dict to avoid modifying the original
+    sector_groups = {k: v.copy() for k, v in sectors_config["sectors_groups"].items()}
+
+    if (
+        "sector_groups_per_species" in sectors_config
+        and species in sectors_config["sector_groups_per_species"]
+    ):
+        # Provide additional sector groups for specific speciess
```
melodb: Typo: "speciess"
Moreover, I had to try the code below to understand what it was doing. Do I understand correctly that in the particular case of the example file you are re-grouping the sectors that will be summed as "Others"? If that's the case maybe you can modify the comment to something like "Modify sector groups for specific species".

fluxy/operators/sectors.py
```diff
@@ -0,0 +1,81 @@
...
+        additional_groups = sectors_config["sector_groups_per_species"][species]
+        # Sometimes the values are the keys of the other dict, so we need to update
```
melodb: Why "sometimes"? Maybe you can add a description of the input file variables to README.md.

fluxy/operators/sectors.py
```diff
@@ -0,0 +1,81 @@
...
+
+def sectors_group_from_config_or_dict(
+    sector_groups: dict[str, list[str]] | None = None,
+    sectors_config: dict[str, str] | None = None,
+    species: str | None = None,
+):
+    """
+    Decorator to group sectors based on a configuration or a dictionary.
+    If `sector_groups` is provided, it will be used to group the sectors.
+    If `sectors_config` is provided, it will be used to map old sectors to new ones.
```
melodb: Add Args and Returns to function description.

fluxy/operators/sectors.py
```diff
@@ -0,0 +1,81 @@
...
+                sector_groups[key].extend(new_values)
+
+    return sector_groups
+
+
+@apply_to_dict_or_single
+def group_sectors(
+    ds: xr.Dataset,
+    sector_groups: dict[str, list[str]] | None = None,
+    sectors_config: dict[str, str] | None = None,
```
melodb, line 59: Add function description

fluxy/plots/mf_timeseries.py
lionel42: I don't know, I feel like not having to give the variable input, but using the function directly is a good improvement. 
And would be consistent with the choice of letting other functions specific for mf (like stats_mf)

fluxy/plots/mf_timeseries.py
lionel42: I added a None option, remove the second column when not necessary and put a test

fluxy/plots/mf_timeseries.py
lionel42: I think here we have another issue, because if the variables to plot is greater than one, we need to print a number of obs that can in theory differ for the different variables. 
For now I decide to print this only in the case of just one variable, and I will use the mask not nan of the variable that is created above already.

fluxy/operators/ecflux.py
lionel42: Good question, I will put your suggestion, because in the future, we might build a documentation automatically that reads the types hints directly.

fluxy/operators/ecflux.py
lionel42: Here I am happy ot keep inside, because then we know the warning come from the filtering

fluxy/operators/ecflux.py
lionel42: Not sure why I needed to print, I put only the logger.info for simiplicity

fluxy/plots/correlation.py
lionel42: I made it possible to get single element list

fluxy/plots/correlation.py
lionel42: I adapted to slice site

fluxy/plots/correlation.py
lionel42: Since the first if statement and the second are doing different things, I prefer to keep them separate for clarity. I added comment to distinguish what is done in each step

fluxy/plots/correlation.py
lionel42: Can still be useful when someone implement another oppose value in the future. I will prefer to keep it.

fluxy/plots/stats.py
lionel42: Issue is that in the dataframe we don't have any info on the variable.

What would you suggest to put as y label otherwise ?
Should I just rename this parameter to y_label ?

fluxy/plots/stats.py
lionel42: I didn't intend to create similar function. In my view the plot stat mf function plots what is relevant to mole fraction data.

The plot stat function is more general for any variables.

If we want to modify the plot stat mf function to use the plot stat function internally, I would do it in a separate PR.

scripts/ec_flux/plot_results.ipynb
lionel42: I agree to move, but renaming to ecflux is not a good idea, we can make a sector config general, which works for all kinds of data

lionel42:
@melodb  Thanks for the review, just had some issue with the general comments
> 
>       
>       * Add umol to units in config
>       * Add longname to the netcdf files

Sorry I wont do because of lazyness (I have to create a long name for like 20 variables)

> 
>     * In my view there is no need to put the EDDY folder inside ecflux. We can treat it as any other model.

Actually I would be happy to separate the data in different types, currently the 3 models for inversions are in the main folder. I think it would make sense to put them also in a subfolder. Because in the future if we add more data, we would be happy to have it more organized in subfolders.


> 
>     * Provocative question: why not changing the yaml file into a json file? (I'm sorry to only raise this after you have changed all the code). There is no problem in supporting both file structures, but I find it a bit more clean if fluxy relies only on one. In this way the user does not need to jump between the rules of one and the other. Consider this with all your love and understanding.

I feel very provoked haha, yaml is just better than json. Easier formatting, comments allowed, less "useless" special characters. 
I wouldn't be against moving all json to yamls (in a separate PR) but if I can use yaml for my things I would be very happy.

fluxy/plots/mf_timeseries.py
```diff
@@ -68,7 +68,7 @@ def plot_timeseries(
     y_lim: None | tuple[float | None, float | None] = None,
     n_bins: int = 30,
     time_freq_min: FrequencyType = None,
-    histogram_type: Literal["hist", "violin", "none"] = "hist",
+    histogram_type: Literal["hist", "violin", "none"] | None = "hist",
```
melodb, line 73: Do we need both `None` and "none"? Maybe "none" can be deleted?

fluxy/plots/mf_timeseries.py
```diff
@@ -558,8 +558,11 @@ def plot_histogram(
             )
 
     # Write number of obs
-    if plot_type == "separate":
-        n_obs = ds["time"].count().values
+    if plot_type == "separate" and len(hist_to_plot) == 1:
+        var = list(vars_to_plot)[0]
+        values = ds[var].values
+        mask_not_nan = ~np.isnan(values)
+        n_obs = np.sum(mask_not_nan)
```
melodb, line 605: Maybe it was not clear from the code, but our goal with n_obs was to show the number of assimilated observations in the inversion during the time window/site being plotted. Not the number of datapoints of whatever variables being plotted. Therefore, it does not matter how many variables are being plotted.
If the dataset was filtered based on the assimilation flag before arriving to this function (which should be the case), ds["time"].count().values should be equal to ds["mf_observed"].count().values or to any other variable. Everything should yield the same result but we applied count() to mf_observed because that's really the variable we are interested in.

Is the number of assimilated observations also a useful info for the analysis of ec-flux results? If yes, I would keep the code as it was and simply fix the assimilation flag in the conversion from old to new format (I can take care of it if you want).

fluxy/plots/correlation.py
```diff
@@ -83,11 +84,15 @@ def plot_correlation(
...
-            )
+        if site is not None:
+            ds_x = slice_site(ds_x, site)
+            ds_y = slice_site(ds_y, site)
+        else:
+            is_single_platform = lambda ds: len(ds["platform"]) == 1
+            if not (is_single_platform(ds_x) and is_single_platform(ds_y)):
+                raise ValueError(
+                    "When oppose='models', both datasets must have a single platform (site)."
+                )
```
melodb, line 101: I just realized that maybe you would like not to have slicing inside the plotting function and rather oblidge the user to call a slicing function before hand. With the current version of the code, `plot_correlation` works nicely with the PARIS dataset if I set `site=<site_name>`. I can use either the non sliced dataset (ds_all_mf) or the sliced one (ds_all_mf_sliced). Even for the sliced dataset, I need to specify site=<site_code> because even though I slice for one single site the platform variable remains unchanged. In plot_timeseries we used `len(np.unique(ds_plot["number_of_identifier"]))` to check if only one site exists.

In summary: either leave it like this - it's working fine. Or remove slice_site and check for the values in number_of_identifier instead; if there is data for more than one site print a warning pointing the used to slice_site as we do in the plot_timeseries function.

fluxy/plots/stats.py
melodb: The `species` you already have as an input to the function, so that you can take from `species_print`. It can always be set to something stupid but at least is something. Then regarding the units, the only thing that I can remember is to add a colum "units" to the data frame when creating it in `stats_observed_vs_simulated`. There this info can be taken from the dataset. The units would not make sense for the unitless quantities but I guess that's fine.

fluxy/plots/stats.py
melodb: I think we can leave it as a separate function as you said. In case we want to adapt the style in the future that can be done in a different PR.

configs/sectors.yml
```diff
@@ -0,0 +1,94 @@
...
+  Shipping:
+    - GNFR_G
+  Others:
+    - GNFR_E
+    - GNFR_R
+    - GNFR_H
+    - GNFR_I
+
+sector_groups_per_species:
+  CH4:
```
melodb, line 38: Change species to lower case

configs/sectors.yml
```diff
@@ -0,0 +1,94 @@
...
+  Road Transport: "#FBC15E"
+  Waste: "#377eb8"
+  Respiration: "#FFB5B8"
+  Others: "#E24A33"
+  Livestock: "#FBC15E"
+  Agriculture: "#ff7f00"
+  Fugitives: "#a65628"
+  Shipping: "#FFB5B8"
+
+molar_masses:
```
melodb: Delete molar_masses

data/tests/ecflux/sectors.yml
melodb, line 1: Is this file still needed or can it be deleted?

fluxy/plots/ec_flux/sectorial_stack.py
```diff
@@ -0,0 +1,175 @@
...
+def raise_var_dims(var_name, ds, expected_dims):
+    if ds[var_name].dims != expected_dims:
+        raise ValueError(
+            f"Variable {var_name} must have dimensions {expected_dims}. "
+            f"Current dimensions: {ds[var_name].dims}"
+        )
+
+
+def plot_stacked(
+    ds: xr.Dataset,
```
melodb, line 26: Add function description

fluxy/plots/ec_flux/sectorial_stack.py
```diff
@@ -0,0 +1,175 @@
...
+            "count": serie_obs_groupped.count(),
+        },
+        axis=1,
+    )
+
+    fig, ax = plt.subplots(figsize=(12, 6))
+    ef_kwargs = {
+        "color": "black",
+        "marker": "x",
+        "label": f"{model} flux\nmeasurements",
```
melodb: The variable in named "model", but then it's used to label the observations. Consider renaming the variable.

fluxy/plots/ec_flux/sectorial_stack.py
```diff
@@ -0,0 +1,175 @@
...
+    # add the text to the existing legend
+    handles, labels = handles[::-1], labels[::-1]
+    handles.append(plt.scatter([], [], color="black", marker="$12$"))
+
+    labels.append("Count of valid\ncomparisons")
+    ax.legend(handles, labels, loc="center left", bbox_to_anchor=(1, 0.5))
+
+    ax.set_ylim(y_lims)
+
+    ax.set_ylabel(f"{substance} Flux " " [ µmol m$^{-2}$ s$^{-1}$ ]")
```
melodb: Maybe replace "substance" by "species" for consistency. Moreover, it would be nice to add the site to the ylabel

fluxy/plots/ec_flux/sectorial_stack.py
```diff
@@ -0,0 +1,175 @@
...
+        "%m": "Month of the year",
+        "%Y_%m": "Year and month",
+        "%m_%H": "Month and hour of the day",
+        "%H_%M": "Hour and minute of the day",
+    }
+    x_label = x_labels[group_format]
+    ax.set_xlabel(x_label)
+    if group_format == "%Y_%m":
+        # Rotate the x labels
+        ax.set_xticklabels(ax.get_xticklabels(), rotation=90, ha="right")
```
melodb, line 181: I noticed the the x-axis is a bit wierd in the example plot: the ticks are all distance by the same amount, but some hours are missing (e.g. 3, 6, 9, ...). It's a bit misleading.

fluxy/plots/mf_timeseries.py
```diff
@@ -274,13 +339,16 @@ def plot_mf_timeseries(
...
-    else:
-        for ax0 in ax[:, 0]:
-            ax0.set_ylim(y_lim)
+        y_lim = [min_mf - 0.02 * min_mf, max_mf + 0.05 * max_mf]
+
+    # Set all the axes to the same y-axis limits
+    for iax, ax0 in enumerate(ax[:, 0]):
+        ax0.set_ylim(y_lim)
+
+        if ncols == 2:
```
melodb: You need to check in this if-statement if `diff_include == None` or an empty list. Otherwise, Obs-variable will be plotted and the x-axis should be left free (it will be approximately centered around 0)

data/tests/ecflux/EDDY/CO2/EDDY_HARDAU_CO2_eddy_flux.nc
melodb, line 1: Delete "CO2" folder

melodb (commented):
Hopefully this is my last set of comments. All files reviewed now. Thanks for adding the eddy covariance flux analysis to fluxy!

fluxy/operators/sectors.py
lionel42: I added a small comment directly in the example yaml file

fluxy/plots/mf_timeseries.py
lionel42: `None` is clearer than `"none"`, I have left `"none"` for backward compatibility

fluxy/plots/mf_timeseries.py
lionel42: This is actually an issue, because the function can now plot any timeseries (simulated or observed) .

I would be happy to simply show the number of datapoints. So maybe we can just rename this to `N` ? or `N_samples` ? or better ? 

Or do you really need to be backward compatible with `mf observed` ? In that case we can adapt the code to read the assimilation flag.

fluxy/plots/correlation.py
lionel42: I think I prefer to leave it like this. 

I know the philosophy is in general to slice before the functions, but another philosphy is to make it easier to the user to get the plot done and this inner slicing is helping doing that.

data/tests/ecflux/sectors.yml
lionel42: I will delete it

fluxy/plots/ec_flux/sectorial_stack.py
lionel42: This variable was not very useful, I removed it

fluxy/plots/ec_flux/sectorial_stack.py
lionel42: Yes, i have some issues with the x ticks. 
I want to see if i find a better solution, so I created a separate issue.

https://github.com/openghg/fluxy/issues/177

fluxy/plots/mf_timeseries.py
lionel42: Not sure I understand.

You mean if diff_include is None, we don't want to set the y_lim ?

fluxy/plots/mf_timeseries.py
melodb: I'm fine with simply showing the number of datapoints and calling it `N`. I just don't like the restriction that it will only shows up if we plot one single variable. The only suggestion that I have is to check if the number of data points is the same for all ploted variables. In that case, `N` can be shown.

fluxy/plots/mf_timeseries.py
melodb: Sorry, I was referring to the if statement `if ncols == 2:` (L.349-351). You set x_lim equal to y_lim. However, if `diff_include == None` or `diff_include == []`, we don't want to set x_lim.

melodb (approved):
Thank you once again for all of this and for your patience during the review. I replied directly to the unresolved issues. I would be happy for it to be merged afterwards.

fluxy/plots/mf_timeseries.py
lionel42: Good, I understood and fixed

fluxy/plots/mf_timeseries.py
lionel42: i have renamed this to N, maybe we can come with another PR later to improve that using another variable to specify how to show this info

lionel42:
@melodb  I have merged and tried, and it works well. Can you try if it works on your side as well ?

PR #165, opened 2025-08-05
Fix usage of subdirectories inside /path/to/data/<inverseModel>/<species>/

melodb:
The first element of the filename is expected to be the inverse model name (e.g. "ELRIS" in "ELRIS_NAME_EUROPE_EDGAR"). Therefore, in the code, the inverse model name was obtained with `split("_")[0]`. This info is used to know the folder name where the data is stored: `/path/to/data/<inverseModel>/<species>/`

This was a problem when creating sub-directories inside `<inverseModel>/<species>/` and trying to access the data with `models = ["subdirectory/ELRIS_NAME_EUROPE_EDGAR"]`. The code would only work if `subdirectory` had the format `<inverseModel>_<etc_etc>`

Therefore, the code was modified so that `split("_")[0]` is applied to the basename.
This allows the use of sub-directories with any name (e.g. "HFC_paper", or "Intercomparison_paper").

### DIFF
--- a/fluxy/config.py
+++ b/fluxy/config.py
@@ -134,7 +134,9 @@ def set_model_colors(models: list[str]) -> dict[str, list]:
 
     Args:
         models (list of str):
-            Keys specifying model names, e.g. ['intem','elris']
+            Model name tags specifying model runs,
+            i.e. '<inversionModel>_<optional_identifying_tags>', preceded by subdirectory if applicable,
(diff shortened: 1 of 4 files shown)


hdelongueville:
That's brilliant and so useful!!

hdelongueville (approved):

lionel42 (approved):

Issue #160, opened 2025-07-10
PARIS_inversion_results.ipynb not working out-of-the-box


mgalkowski:
After running a clean download, the `PARIS_inversion_results.ipynb` is not running in full. Blocks throw errors complaining about missing data.

Appears to be related with some new-format data files missing in `/data/tests/` (e.g. InTEM dir only contains old format data).

lionel42:
Hi @mgalkowski , 
could you upload here the full traceback of the errors ?

mgalkowski:
Hi @lionel42,

Here's what I see when trying to execute the second block of the `PARIS_inversion_results.ipynb`

```python
WARNING:fluxy.io:Cannot find flux file: /work/bm1400/b301033/projects/fluxy/data/InTEM/hfc134a/InTEM_NAME_EUROPE_EDGAR_intem_obs_intem_baseline_optimized_hfc134a_yearly.nc.
WARNING:fluxy.io:Cannot find flux file: /work/bm1400/b301033/projects/fluxy/data/ELRIS/hfc134a/ELRIS_NAME_EUROPE_EDGAR_elris_obs_elris_baseline_optimized_HFC_134a_yearly.nc.
WARNING:fluxy.io:Cannot find flux file: /work/bm1400/b301033/projects/fluxy/data/RHIME/hfc134a/RHIME_NAME_EUROPE_EDGAR_rhime_obs_rhime_baseline_optimized_hfc134a_yearly.nc.

---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
Cell In[2], line 29
     26     ds_all_flux = read_model_output(data_dir,"flux",species,models,config_data,period=period)
     28     for m in models:
---> 29         ds_all_flux_scaled[m] = slice_flux({m:ds_all_flux[m]},config_data,start_date,end_date,species=species,country_flux_units_print=country_flux_units_print)[m]
     31 model_colors = set_model_colors(models)
     32 model_labels = set_model_labels(models,config_data,get_labels_from_file)

KeyError: 'InTEM_NAME_EUROPE_EDGAR_std'
```

As I've mentioned above, in `fluxy/data/tests/InTEM/hfc134a`, I only see two files:
InTEM_NAME_EUROPE_EDGAR_old_format_hfc134a_yearly_concentrations.nc  
InTEM_NAME_EUROPE_EDGAR_old_format_hfc134a_yearly.nc

So clearly the model expects other files to be present as well.

_[updated for formatting purposes]_

[...]

PR #172, opened 2025-08-08
New example notebook

lionel42:
In #160 , we discussed the issue that the example of fluxy was not working directly and this could be annoying for a new user who wants to try out fluxy.

We decided to create another new notebook, which would be the basic example of what fluxy can do.


lionel42:
@melodb As we discussed, here is the PR for a basic example. What do you think ?

melodb (approved):
Thanks for this. It looks great! I noticed a bug in the new format ELRIS file so I have replaced it. Ready to merge from my side.

Issue #171, opened 2025-08-08
ANNEX: seasonal spatial plot


melodb:
When plotting the seasonal difference, the 4 plots are quite far away from each other. Could this be optimized? Moreover, the ticks of the middle plots are a bit on top of the labels "Jun-Aug" and "Sep-Nov". Would it be possible to remove the ticks from the interior plots and keep them only in the outside plots? @lionel42 asked for this improvement quite some time ago but I have continuously postponed it.

Here is the current plot:

<img width="2353" height="1640" alt="Image" src="https://github.com/user-attachments/assets/cfeb3c8a-bb6d-499a-99f0-e1d25079c14d" />

And this is how I think it should be:

<img width="2155" height="1804" alt="Image" src="https://github.com/user-attachments/assets/1c54adcd-9472-4b43-8bc3-b80a0111573d" />

hdelongueville:
To double check: do you now if we only want this special display only for seasonal - model combined?
Currently, this display is called if n_rows * n_cols == 4:, which seems not to be what we want. 
If this is indeed not correct, I'll change the condition to be more specific.

melodb:
In my opinion, these changes can be applied to all spatial maps. That is, remove the x/y numbers of internal axis (but keep the ticks) and make sure the white spacing is well balanced in the plot.

Regarding specifically to your question: I think you can keep the condition. I'm not aware of situations for which n_rows * n_cols == 4 and we would like a different display.

hdelongueville:
I just had a situation like that where I am looking at two models for two years. The function failed in that case.

hdelongueville:
For the white spacing, it is very tricky because of the size of the region to plot varies a lot. I'll give it another try, but if anyone's got an idea on how to improve this, I'm interested.

melodb:
> For the white spacing, it is very tricky because of the size of the region to plot varies a lot. I'll give it another try, but if anyone's got an idea on how to improve this, I'm interested.

Are you able to get the aspect ratio of the crop before setting the figure size? If yes, you could then make sure the total figure size follows more a less the same aspect ratio.

melodb:
Well, the same aspect ratio only if its a squared figure. But it should be easy to make the math for different nrows/ncolums.

hdelongueville:
Ok, I have found a way to remove the extra blank space, using layout='compressed' in plot.subplots().

[...]

PR #173, opened 2025-08-08
Annex improvements

melodb:
### Main fluxy developments:
- Option to plot standard run was implemented in `get_filename` (flag `read_standard_run`). Similarly to `json_exp_to_use`, the standard runs can now be specified in `models_info.json`. Simply add a dictionary for the model name (e.g. "InTEM_longrun"), otherwise it will take the standard run from key "default". (see config file to reproduce the annexes in JupyterHub - NID2025/config_for_fluxy)
- As a consequence, `read_flux_total_fgases` was updated. In particular, dictionaries `default_overwrite` and `valid_experiments` don't exist anymore. Standard fluxy errors will be raised if a file is missing.

### Annex plot generator:
- The code was updated taking into account the implementation of `read_standard_run`
- start/end dates were moved to the config file
- Some changes were made so that exceptions (like N2O for NID2025) can be easily set
- Hard-coded settings were adjusted to yield the NID2025 plots

### To do for the annexes (not for this PR):
- Check inventory data: values differ from 2025 submission @mo-aliceramsden
- #171 Fix spatial maps so that they look better @hdelongueville
- Small final fixes @melodb:
   - Delete "(resampled)" from legend when annex_mode = True
   - Make uncetainty shading lighter
   - Discuss crop of the UK (it does not include IRE) - we might need to define all crops manually to include neighboring countries
   - Update config file to yield NID2026 plots (remove special case of N2O)

### Mistery:
There are no PFC-318 inversion results between 2008 and 2010 for InTEM_longrun. However, in the old plots, the total PFC plot started in 2008. I don't know how that happened. The new plots start in 2011 as expected (at least in my view).


fluxy/cli/config_annex_plot.py
```diff
@@ -189,7 +189,7 @@ class AnnexConfig:
         start_date_fgases (dict): start date to use for every species of the selected region
     """
     ### Path to results directory
-    data_dir = "/project/paris/inverse_modelling/"
+    data_dir = "/project/bmda/PARIS_jupyter/NID2025/"
```
hdelongueville, line 195: Is this a specific directory for the annexes? why not using /project/paris/NID2025/?

fluxy/cli/config_annex_plot.py
```diff
@@ -251,31 +265,44 @@ class AnnexConfig:
...
+            plot_separate=[True, False, False, False],
+            plot_combined=[False, True, True, True],
+        ),
+        "n2o": dict(
+            plot_separate=[True, False, False],
+            plot_combined=[False, True, True],
+        ),
+    }
+
+    ### Settings for spatial maps (list or dictionary if different between species)
```
hdelongueville, line 304: Should we add a warning message that RHIME is missing, to be sure we don't forget using it for NID2026?

hdelongueville (approved):
Thanks a lot for this tedious work! I've added two minor comments.

hdelongueville:
Is fluxlim_percentiles in config_annex_plot still used? If not (which I think is the case), should we remove this section?

fluxy/cli/config_annex_plot.py
melodb: Yes, sorry, it's my personal folder. I will change it to /project/paris/NID2025/

fluxy/cli/config_annex_plot.py
melodb: I have added some prints. They are however only notes for ourselves that can be deleted when setting up NID2026.

melodb:
> Is fluxlim_percentiles in config_annex_plot still used? If not (which I think is the case), should we remove this section?

You're right. I have deleted the dictionary ;)

melodb:
@lionel42 would you mind taking a look at io.py? There is no need for you to revise the annex part - that's very paris specific.

fluxy/cli/config_annex_plot.py
```diff
@@ -214,18 +81,33 @@ class AnnexConfig:
...
-        "InTEM_longrun",
-        "InTEM",
-        "ELRIS",
-        "RHIME",
-    ]  # NOTE: only options are basic model names w/ and w/o longrun
-
-    # for annual species
+    # for monthly species (list or dictionary if different between species)
+    # no RHIME N2O results in NID2025
+    print("WARNING: Excluding RHIME from N2O country fluxes!")
```
lionel42: use logger.warning

scripts/PARIS_inversion_results.ipynb
```diff
@@ -810,7 +813,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.10.6"
+   "version": "3.11.7"
```
lionel42: might create conflicts with other branches if you change the version

fluxy/io.py
```diff
@@ -115,6 +115,7 @@ def get_filename(
     file_pattern: str,
     config_data: dict[str, dict],
     data_dir: os.PathLike | str,
+    read_standard_run: bool,
 ) -> Path:
```
lionel42, line 119: add documentation for this new parameter

lionel42 (approved):
I am not very familiar with the cli code, so I had only a brief overview.

About the io, it looks good

alexdanjou:
About excluding the RHIME data for N2O, this is only for 2025 annexes, it won't be the case for the 2026.

Maybe we should rename `config_annex_plot` to `config_annex_plot_{year}` for tracability over the years.

If you decide to let this for the next round of update, no pb, I'll open an issue to keep in mind :)

melodb:
> About excluding the RHIME data for N2O, this is only for 2025 annexes, it won't be the case for the 2026.
> 
> Maybe we should rename `config_annex_plot` to `config_annex_plot_{year}` for tracability over the years.
> 
> If you decide to let this for the next round of update, no pb, I'll open an issue to keep in mind :)

At the moment I'm saving the config files for NID2025 in JupyterHub (see "/project/paris/NID2025/config_for_fluxy"). In this way we have data and config files in the same location. Renaming the file itself is a bit tricky because then we would have to adapt the import statement in annex_plot_generator. Once this is merged and the TODO list mentioned above is done, I will open a new PR to update the config file for NID2026.

melodb:
@alexdanjou I will merge the branch now if that's ok for you

melodb:
> > Is fluxlim_percentiles in config_annex_plot still used? If not (which I think is the case), should we remove this section?
> 
> You're right. I have deleted the dictionary ;)

After all it was still used so I have added it back in. We can rethink it's usage for NID2026

Issue #170, opened 2025-08-08
ANNEX: rolling mean


melodb:
At the moment, rolling_mean is applied to all datasets. However, in the F-gases plots of the annexes, we want to apply it to all models expect InTEM_longrun. Could you change the code so that rolling_mean can also be a list of bool as before?

alexdanjou:
if 3 plots are combined and the 3 are applied a rolling_mean, what is the expected behavior for the combined plot? To have a rolling_mean applied as well?

melodb:
I don't remember the order of the rolling_mean/combined operations in the code. But I would say that the rolling mean should not be applied twice.

Issue #151, opened 2025-06-23
Combined uncertainties not matching separated uncertainties when rolling_mean = True


hdelongueville:
When `rolling_mean = True` and `plot_combined = True`, the combined uncertainties are not matching the min and max uncertainties of the separated uncertainties (see figure 1). Is it because the rolling_mean is applied on the combined_uncertainties? Should we calculate the combined_uncertainties after having applied the rolling_mean to the separated flux instead? 
As you can see in figure 2 where `rolling_mean = False`, the uncertainties are matching then. 

Do we want to keep it as is or do we want to recalculated combined uncertainties when `rolling_mean = True`?

Figure 1 (`rolling_mean = True`)
![Image](https://github.com/user-attachments/assets/fcfdc31f-dd4b-49fb-afb8-16dcbabc0c1a)

Figure 2 (`rolling_mean = False`)
![Image](https://github.com/user-attachments/assets/612549b3-4468-4be6-9376-bcc655001b8b)

melodb:
We need to address this issue because it is leading do different uncertainty bands in the annexes. I think we should make it consistent like in Figure 2. @alexdanjou do you think you could easily solve this?

PR #175, opened 2025-08-14
Iss170 rolling mean

alexdanjou:
Fix issue #170 and #151


melodb (approved):
Thanks for this! I have tested your implementation with the annex data and I got the correct values.

PR #176, opened 2025-08-15
fix calculating prior uncertainty for regions from countries

stiglers-eponym:
fluxy.operators.regions.extract_region_flux sums up covariances of countries to get the variance of a region. In that function, the selection of countries was not applied consistently, leading to wrong prior uncertainties.


alexdanjou (approved):
Well spotted! Could you also change `ds` to `ds_region` lines 87, 99 and 108 (the `if` conditions). Won't change anything but will make stuff more consistent :)

Thanks !

alexdanjou:
Great thanks! I've merged your changes

PR #179, opened 2025-08-20
Bug fix for adding separate BEL and LUX fluxes, so this works with both new and old InTEM files

mo-aliceramsden:
Code previously had hardcoded fixes for InTEMs `countryname` and `countrynumber` variables. These fixes aren't required with InTEM's new-format netcdfs, so these fixes have been made optional (based on presence of these variables). This fix was also required for the part of the code that added separate BEL and LUX fluxes to InTEM's data (as InTEM only outputs BEL-LUX fluxes). The code can now read in both new and old format InTEM netcdfs.


melodb (approved):
Thanks for this fix!

Issue #79, opened 2025-02-17
Improve notebook structure and functionality based on ACRG feedback


hdelongueville:
Here are some suggestions to improve the notebook, based on discussions with the ACRG modellers at the University of Bristol:

- [x] **CO2-eq:** Specify the source of GWP values in the notebook and verify HFO values.
- [x] **Domain specification:** Move it to the notebook or a dedicated JSON file.
- [ ] #185
- [ ] **Jupyter Notebook functions**: Pass a dictionary instead of separate parameters.
- [ ] **Save plots**: Add a global save option at the top to avoid multiple save plot cells.
- [x] #197
- [x] **NetCDF templates**: Should we add them to the repository?

melodb:
Thanks Helene for putting everything here. Regarding the 2nd point (Domain specification: Move it to the notebook or a dedicated JSON file.), we already have variable "domain" defined in models_info.json. Is that it or does this comment refer to something else?

hdelongueville:
The ACRG modellers were suggesting to move it from `models_info.json` as it is quite different from the rest of the content.

melodb:
Ok, I get it. Maybe we can move it to the first cell of the notebook.
Regarding the comment to pass a dictionary instead of separate parameters, does it mean to define something like:
`plot_settings = {plot_inventory: False, inventory_years: None, ...}`
And to input plot_settings to the function?

hdelongueville:
Yes, we could use the ** unpacking operator to pass a dictionary as keyword arguments to a function. 
For example, 
def my_function(a, b, c):
    print(f"a: {a}, b: {b}, c: {c}")

params = {"a": 1, "b": 2, "c": 3}

my_function(**params)

melodb:
That's great! Let's do that!
Regarding the domain, maybe a good alternative is to move it to a new json file (e.g. regions_info.json) together with the country/region definitions and cities lat/lon which are currently in config.py

hdelongueville:
I like that idea!

melodb:
> I like that idea!

I'm creating the json file and making the necessary changes in the code.

PR #184, opened 2025-09-03
Update site_info using openghg_defs 2025-09-03 status

### LINKED ISSUE #164: site_info.json may be out of date, but is it used?
Someone spotted that the GSN inlet is wrongly listed as 10m in ```site_info.json``` (should be 17m). 

This looks like the file from the ACRG repo, which may be a bit out of date now? Are we sure this data is reliable?

In any case, I couldn't see anywhere where this file is referenced in the code, only in the README.md. Is it used? If not, should it be removed?

hdelongueville:
Update `fluxy/configs/site_info.json` using `openghg_defs/data/site_info.json` latest status from 2025-09-03.

### DIFF
--- a/configs/site_info.json
+++ b/configs/site_info.json
@@ -1,3597 +1,4193 @@
 {
-    "AAO": {
-      "NOAA": {
-        "height_station_masl": 230.0,
-        "latitude": 40.05,
-        "long_name": "Airborne Aerosol Observatory, Bondville, Illinois, United States",
-        "longitude": -88.37
-          }
-      },
-    "ABP": {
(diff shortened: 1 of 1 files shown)


alexdanjou:
Could we add openghg_defs as a needed library in a requirements.txt? That way we don't have to copy paste it when we update it

hdelongueville:
Yes, that could be an option which has already been suggested by Rachel during a meeting: https://github.com/openghg/fluxy/issues/79. It would require some changes in fluxy so I'd suggest doing that in a futur PR.

melodb (approved):

Issue #178, opened 2025-08-19
plot_inversion_grid_flux = True  only affects posterior and not prior in maps


alexdanjou:
The diff posteior - prior is done with `"flux_{sector}_posterior_inversion_grid"` but `"flux_{sector}_prior"`. I think it's not what should be done

melodb:
Thanks for spotting this. I have completely forgotten. In theory, if `variable flux_{sector}_prior_inversion_grid` exists, that should be shown in the top plot and the diff should be computed from `flux_{sector}_posterior_inversion_grid - flux_{sector}_prior_inversion_grid`. However, the prior_inversion_grid variable does not exist in all datasets (at least not in ELRIS). Maybe an exception can be implemented where it trys to use it and if it does not exist reverts to the non inversion grid variable.

alexdanjou:
Ok! I'll implement it

PR #186, opened 2025-09-05
fix issue #178

alexdanjou:
Fix issue #178


alexdanjou:
If we set `plot_inversion_grid_flux = True` in `plot_flux_map` and one model hasn't the variable `"flux_prior_inversion_grid"` but has the variable `"flux_posterior_inversion_grid"`, the prior for this model will be plotted on the prior grid, the posterior on the inversion grid, and the diff on the prior grid. 
For all the other models (if they have both `"flux_prior_inversion_grid"` and `"flux_posterior_inversion_grid"`), the plots will be made on the inversion grid.

So it can result in only one plot being on a different grid and unconsistency in the displayed grid in a column. But doing otherwise would require way more adaptations (for something that must be corrected soonish (?) by ELRIS maybe).

I think that, although not being ideal, this is better than the current state where the diff is made on different grid.

fluxy/operators/flux_map_diff.py
```diff
@@ -34,24 +37,42 @@ def define_var_plot(
...
             ds_output[var_p] = ds[f"flux_{sector}_posterior"] - ds[
                 f"flux_{sector}_posterior"
             ].mean(dim="time")
+
         elif var_p == "posterior_prior_diff_inversion_grid":
-            ds_output[var_p] = (
-                ds[f"flux_{sector}_posterior_inversion_grid"]
-                - ds[f"flux_{sector}_prior"]
-            )
+            if f"flux_{sector}_prior_inversion_grid" in ds: # fix because ELRIS don't have (yet) prior on inversion grid
```
hdelongueville: I think this caution can't stay even after ELRIS adds prior on inversion grid. So maybe change the comment to something like "# ensure both prior and posterior are on the same grid."

fluxy/operators/flux_map_diff.py
```diff
@@ -34,24 +37,42 @@ def define_var_plot(
...
+
         else:
-            if var_p not in ds:
-                raise ValueError(f"'{var_p}' not found in dataset(s)")
-            ds_output[var_p] = ds[var_p]
-            unit_var = var_p
+            if var_p in ds:
+                var_p_bis = var_p
+            else:
+                if "prior_inversion_grid" in var_p: # fix because ELRIS don't have (yet) prior on inversion grid
```
hdelongueville: Same as previous comment, change to something like "# use prior grid if inversion grid is missing"

fluxy/plots/flux_map.py
```diff
@@ -112,6 +115,10 @@ def plot_flux_map(
             f"Currently, you cannot plot sectors other than 'total' using the inversion_grid variable. "
             + "Set plot_inversion_grid_flux to False to plot other sectors."
         )
+    if plot_inversion_grid_flux:
+        logger.warning(
```
hdelongueville: Is this additional warning necessary? If one dataset is not on inversion grid, there will be a warning anyway.

fluxy/plots/flux_map.py
```diff
@@ -347,6 +352,10 @@ def plot_flux_map_model_comparison(
             f"Currently, you cannot plot sectors other than 'total' using the inversion_grid variable. "
             + "Choose a non inversion_grid variable to plot other sectors."
         )
+    if "inversion_grid" in var:
```
hdelongueville: same comment as before

hdelongueville (approved):

alexdanjou:
@melodb Are you ok with that ? If yes I'll merge :)

> If we set `plot_inversion_grid_flux = True` in `plot_flux_map` and one model hasn't the variable `"flux_prior_inversion_grid"` but has the variable `"flux_posterior_inversion_grid"`, the prior for this model will be plotted on the prior grid, the posterior on the inversion grid, and the diff on the prior grid. For all the other models (if they have both `"flux_prior_inversion_grid"` and `"flux_posterior_inversion_grid"`), the plots will be made on the inversion grid.
> 
> So it can result in only one plot being on a different grid and unconsistency in the displayed grid in a column. But doing otherwise would require way more adaptations (for something that must be corrected soonish (?) by ELRIS maybe).
> 
> I think that, although not being ideal, this is better than the current state where the diff is made on different grid.

melodb (approved):
All good! Thanks a lot

PR #187, opened 2025-09-05
171 annex seasonal spatial plot

### LINKED ISSUE #171: ANNEX: seasonal spatial plot

hdelongueville:
Optimize map plots size to avoid extra white space + remove duplicated tick's labels for a cleaner look.

### DIFF
--- a/fluxy/plots/flux_map.py
+++ b/fluxy/plots/flux_map.py
@@ -175,7 +175,7 @@ def plot_flux_map(
     figsize = define_map_figsize(
         map_bounds, n_rows, n_cols, fixed_value=3 * n_rows, fixed_dimension="height"
     )
-    fig, ax = plt.subplots(n_rows, n_cols, figsize=figsize, constrained_layout=True)
+    fig, ax = plt.subplots(n_rows, n_cols, figsize=figsize, layout='compressed')
 
     for col, (model, ds) in enumerate(ds_dict.items()):
         lon, lat = ds.longitude, ds.latitude
@@ -389,8 +389,11 @@ def plot_flux_map_model_comparison(
     # Initialize figure
     n_rows = 1
     n_cols = 3
+    figsize = define_map_figsize(
+            map_bounds, n_rows, n_cols, fixed_value=3*n_cols, fixed_dimension="width"
+        )
     fig, ax = plt.subplots(
-        n_rows, n_cols, constrained_layout=True, figsize=(n_cols * 5, 9)
+        n_rows, n_cols, figsize=figsize, layout='compressed'
     )
     for col, (model, ds) in enumerate(ds_dict.items()):
         ax_i = ax[col]
--- a/fluxy/plots/flux_map.py
+++ b/fluxy/plots/flux_map.py
@@ -396,7 +396,7 @@ def plot_flux_map_model_comparison(
     n_rows = 1
     n_cols = 3
     figsize = define_map_figsize(
-            map_bounds, n_rows, n_cols, fixed_value=3*n_cols, fixed_dimension="width"
+            map_bounds, n_rows, n_cols, fixed_value=5*n_cols, fixed_dimension="width"
         )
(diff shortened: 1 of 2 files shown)


fluxy/plots/flux_map.py
```diff
@@ -609,12 +623,18 @@ def plot_flux_map_over_time(
...
+            map_bounds, 2, 2, fixed_value=7, fixed_dimension="width"
+        )
+        fig, ax = plt.subplots(2, 2, figsize=figsize, layout='compressed')
         ax = ax.flatten()
     else:
-        fig, ax = plt.subplots(n_rows, n_cols, figsize=(n_cols * 4, n_rows * 3))
+        figsize = define_map_figsize(
+            map_bounds, n_rows, n_cols, fixed_value=3*n_cols, fixed_dimension="width"
+        )
+        fig, ax = plt.subplots(n_rows, n_cols, figsize=figsize, layout='compressed')
```
melodb: Just a small code improvement suggestion: maybe inside `if (is_single_season)`/`else`, you can simply re-define the number of rows/collums (e.g. fig_rows = ...,  fig_cols = ..., fixed_value = ...) and then call `define_map_figsize` and `plt.subplots` outside the if-statements using these variables as inputs.

melodb (commented):
Thanks for working on this. I just left a small comment in the code.
Regarding the plots layout, there are some issues:
- in `plot_flux_map_model_comparison`, the color bar at the bottom is on top of the tick labels.
- in `plot_flux_map_over_time`, when the figure has multiple rows but only one column (e.g. 3 models, chop_by = year, dt = max number of years in dataset), the color bar is placed at the bottom subplot instead of spanning all columns.
- also in `plot_flux_map_over_time`, when the figure has only one subplot (1 row and 1 column), (e.g. 1 model, chop_by = year, dt = max number of years in dataset), the code breaks with error message: "UnboundLocalError: cannot access local variable 'target_ax' where it is not associated with a value".
- Moreover, in the same case described above (1 model), the model name does not show up on the left.

fluxy/plots/flux_map.py
hdelongueville: Oh yes you're right! I meant to do that and then forgot. Thanks for spotting that!

hdelongueville:
Thanks for spotting all those issues!
For "in plot_flux_map_model_comparison, the color bar at the bottom is on top of the tick labels.", do you know which region you plotted? Because I don't have that issue for UK.

melodb:
> Thanks for spotting all those issues! For "in plot_flux_map_model_comparison, the color bar at the bottom is on top of the tick labels.", do you know which region you plotted? Because I don't have that issue for UK.

I have plotted it for France (I used the raw notebook):
image

I get the same with HFC-134a. I think it depends on the aspect ratio of the image because I have the same problem for Norway, Hungary and Switzerland.

hdelongueville:
Hopefully, the layout should be better now. 
@melodb Can I ask you to replot your figures to be sure it is working now?

melodb (approved):
All working fine now!
However, I just noticed that the sites plotting is not working fine. I get the warning:

WARNING:fluxy.plots.utils:No 'sites' found in dataset. Please ensure 'add_sites_to_flux' is True in 'read_model_output' or that a 'fallback_sites' list is provided in plot_flux_map.

I leave to you the decision to solve it now or in another PR.

hdelongueville:
Yes, I noticed! I'm wondering if this happened during the merging of another PR..
I'll fix this in another PR.

Issue #183, opened 2025-09-01
Optimisation of sector flux calculation and plotting


mo-aliceramsden:
A few updates could be made to the sector flux plotting:

- Optimisation of sector flux calcuation, by using xarray instead of numpy function (whilst keeping option to only calculate country sector flux for a short list of countries, as this speeds up the calculation considerably)
- Combination of current sector plotting functions functions with those in #162 
- Include uncertainty in posterior sector fluxes (after discussion on how to calculate this)

PR #189, opened 2025-09-05
Update fluxy/operators/flux_scale_by_sector_proportions.py so it uses xarray and not numpy to calculate new variables

alexdanjou:
Solves one point of issue https://github.com/openghg/fluxy/issues/183

### DIFF
--- a/fluxy/operators/flux_scale_by_sector_proportions.py
+++ b/fluxy/operators/flux_scale_by_sector_proportions.py
@@ -9,6 +9,79 @@ from fluxy.operators.flux_align_dataset import align_lat_lon
 logger = logging.getLogger(__name__)
 
 
+def open_and_align_sector_dataset(
+    sector_prop_path: str, ds_ref: xr.Dataset
+) -> xr.Dataset:
+    """
+    Open dataset containing prior sector fluxes and align it temporaly an spatially on reference dataset.
+
+    Args:
+        sector_prop_path: path to dataset containing prior sector fluxes
+        ds_ref: datset used as reference to align along
+
+    Returns:
+        ds_sectors: dataset with prior sector fluxes align on reference dataset
+    """
+
+    with xr.open_dataset(sector_prop_path) as f:
+        # timely align sector dataset on main dataset
+        freq_ds = ds_ref.attrs["frequency"]
+        if xr.infer_freq(f.time) not in ["YS-JAN", "AS-JAN"] or freq_ds not in [
+            "monthly",
+            "yearly",
+        ]:
+            raise ValueError(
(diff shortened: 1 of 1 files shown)


fluxy/operators/flux_scale_by_sector_proportions.py
```diff
@@ -9,6 +9,79 @@
 logger = logging.getLogger(__name__)
 
 
+def open_and_align_sector_dataset(
+    sector_prop_path: str, ds_ref: xr.Dataset
+) -> xr.Dataset:
+    """
+    Open dataset containing prior sector fluxes and align it temporaly an spatially on reference dataset.
```
hdelongueville: typo "and"

fluxy/operators/flux_scale_by_sector_proportions.py
```diff
@@ -9,6 +9,79 @@
...
+
+    with xr.open_dataset(sector_prop_path) as f:
+        # timely align sector dataset on main dataset
+        freq_ds = ds_ref.attrs["frequency"]
+        if xr.infer_freq(f.time) not in ["YS-JAN", "AS-JAN"] or freq_ds not in [
+            "monthly",
+            "yearly",
+        ]:
+            raise ValueError(
+                "This part of the code as not be tested for a sector file with frequency not equal to 'YS-JAN' or dataset frequency not one of 'monthly'/'yearly'. In the current implementation, the sector file frequency is suppose to be bugger or equal to the dataset frequency and start before or at the same time as the dataset."
```
hdelongueville: suggestion:
"This part of the code has not been tested with a sector file with frequency different from 'YS-JAN' nor a dataset's frequency different from 'monthly'/'yearly'. In the current implementation, the sector file's frequency is suppose to be bugger or equal to the dataset's frequency and to start before or at the same time as the dataset."

fluxy/operators/flux_scale_by_sector_proportions.py
```diff
@@ -9,6 +9,79 @@
...
+        _, ds_sectors = align_lat_lon([ds_ref, ds_sectors], "longitude")
+
+    return ds_sectors
+
+
+def create_cell_area(
+    ds_ref: xr.Dataset, cell_area_test_file: bool, sector_file: str
+) -> xr.DataArray:
+    """
+    Open dataset containing prior sector fluxes and align it temporaly an spatially on reference dataset.
```
hdelongueville: typo "and"

fluxy/operators/flux_scale_by_sector_proportions.py
```diff
@@ -9,6 +9,79 @@
...
+
+    Args:
+        ds_ref: path to dataset containing prior sector fluxes
```
hdelongueville: is it? isn't ds_ref the reference dataset?

hdelongueville (approved):
Thanks! 
I've just commented some typos which I think are from a previous PR :)

fluxy/operators/flux_scale_by_sector_proportions.py
alexdanjou: Oups indeed

Issue #148, opened 2025-06-19
`region` parameter in `flux_map` functions don't work for region that are not country (and also for France) in some work environment


alexdanjou:
It is working fine on ICOS Jupyter Lab and my local, but not on the JASMIN Jupyter Lab... So might be a geopandas (or whatever) version issue.

hdelongueville:
I have weird stuff happening on my side. When I open a new session, it works fine. Then at one point it doesn't work anymore but I am not able to reproduce the issue. I haven't figure it out what is making it failing.

melodb:
Is this issue solved?

hdelongueville:
I don't think so because I'm still having this issue sometimes :/

PR #192, opened 2025-09-08
Fix issue 148

alexdanjou:
Fix issue #148 
Change the clipping region to the limits of where country_fraction is non zero for the country or group of country. It was previously clipped to where country_fraction was different of zero (no selection of the country), which resulted in selecting the whole domain (I think it was a mistake).


alexdanjou:
This needs the country_fraction variable in the first dataset of the dataset list.

Is it always the case? If not I'll improve that to check every dataset and just keep the first one that has this variable.

hdelongueville (approved):

PR #200, opened 2025-09-11
temporary fix for repeated TAC timestamps bug in InTEM files

mo-aliceramsden:
This is a short term fix to allow for concentration comparisons when there are bugs in InTEM's concentration files (repeated TAC timestamps)

### DIFF
--- a/fluxy/io.py
+++ b/fluxy/io.py
@@ -875,6 +875,12 @@ def edit_vars_and_attributes(
             # Fill intake_height and stdev_mf_model with fake values
             ds["intake_height"][:] = 0
             ds["stdev_mf_model"][:] = 0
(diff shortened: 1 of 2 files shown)


melodb:
I have made a suggestion. Check if you agree. In this way it does not stay in the middle of the code. And if we forget to delete it it's not a big issue.

mo-aliceramsden:
> I have made a suggestion. Check if you agree. In this way it does not stay in the middle of the code. And if we forget to delete it it's not a big issue.

Yes your implementation is much better, you're right that will be easier to remember that the temporary fix has been added if it's included there. I'm happy for that to be merged in now, if you are. Thanks :)

melodb (approved):

PR #201, opened 2025-09-11
Change time selection from using `.where` to create an index coordinates and use `.sel` on it.

alexdanjou:
Using `.where` was causing performace issues with InTEM files. For ch4 when `site=None`, `slice_mf` goes from 20mn with `.where` to 6mn with `.sel`. There is still a time issue with the scaling of the variables (the left 6mn), but this is a problem with InTEM files rather than fluxy (@mo-aliceramsden @moPeterAndrews ). 

With all other models, `slice_mf` only takes a fraction of seconds, whether with use `.where` or `.sel`.


fluxy/operators/select.py
```diff
@@ -76,7 +76,7 @@ def slice_flux(
             logger.warning(
                 f"No {m} fluxes found between {start_date[im]} and {end_date[im]}."
             )
-            ds_all_sliced.pop(m)
+            ds_all.pop(m)
```
melodb, line 79: Why did you change this? Is it a mistake from solving the conflicts?

melodb (approved):
I left only a small comment regarding a fix in select. I have tried the code and it's working fine for me.

melodb:
But what's actually the problem with InTEM files? I don't really get it

fluxy/operators/select.py
alexdanjou: Yes, it seems like a mistake from conflict solving, good catch!

alexdanjou:
The pb disappeared with PR #200 so I'm closing this PR

PR #202, opened 2025-09-11
Iterate over all models to get the clip region rather than just checking the first model

alexdanjou:
Also solves issue #197

### DIFF
--- a/fluxy/plots/utils.py
+++ b/fluxy/plots/utils.py
@@ -78,7 +78,7 @@ def add_colorbar(fig, ax, im, extend, label, n_cbar, idx_cbar, colorbar_type="ro
         ax_dim = np.array(ax).ndim
 
         if ax_dim == 1:
-            if nrows ==2 and ncols ==2: # single_season case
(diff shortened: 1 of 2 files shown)


hdelongueville:
From a group discussion:
- first check the country_fraction
- if it does not exists for any of the dataset, use the geopandas library
+ if country_fraction exists but the region/country is not in country_fraction --> use the geopandas library

Would it be possible to add the two last conditions?

alexdanjou:
PR is ready. However, when it is based on country_fraction the window will change compared as before for country or regions with  maritime area (as country_fraction is not just land, compared to geopandas).

ex with NW_EU:
image

melodb (approved):
Thanks Alex for these improvements! I have tried it for UK, FRA, BENELUX, DEU, NW_EU2 and honestly I think it looks fine. We can always specify lon/lat if we want a specific zoom-in.

melodb:
I took the freedom to add a small commit to your branch. I have just added a check if `len(region) == 4` when it is a list/tuple.

PR #203, opened 2025-09-15
Remove "total" from labels

melodb:
I find a bit annoying to have the word "total" in all the labels. Would it be ok to suppress it?


mo-aliceramsden (approved):
Yes this is a great idea, apologies for not thinking of it when I put the sector options in originally!

PR #205, opened 2025-09-15
Optimize annex_plot_generator

melodb:
- Move helper functions to utils_annex_plot
- Replace some settings by species-specific dictionaries
- Refactor annex_plot_generator to loop over all species

These modifications reduce considerably the amount of repeated code.

### DIFF
--- a/fluxy/cli/config_annex_plot.py
+++ b/fluxy/cli/config_annex_plot.py
@@ -229,24 +229,14 @@ class AnnexConfig:
 
     ### Settings for country fluxes
     ## Model definitions (list or dict["<period>": list(), "<species>": list()] if different between species)
-    models_country_flux = {
-        "monthly": [
-            "InTEM_NAME",
-            "InTEM_FLEXPART",
-            "ELRIS_NAME",
-            "ELRIS_FLEXPART",
-            "RHIME_NAME",
-            "RHIME_FLEXPART",
-        ],
-        "yearly": [
-            "InTEM_NAME",
-            "InTEM_FLEXPART",
-            "ELRIS_NAME",
-            "ELRIS_FLEXPART",
-            "RHIME_NAME",
-            "RHIME_FLEXPART",
-        ],
-    }
+    models_country_flux = [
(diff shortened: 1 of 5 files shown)


fluxy/cli/config_annex_plot.py
```diff
@@ -226,26 +228,25 @@ class AnnexConfig:
...
-    models_yearly_species = [
-        "InTEM_NAME",
-        "InTEM_FLEXPART",
-        "ELRIS_NAME",
-        "ELRIS_FLEXPART",
-        "RHIME_NAME",
-        "RHIME_FLEXPART",
-    ]
+    ## Model definitions (list or dict["<period>": list(), "<species>": list()] if different between species)
+    models_country_flux = {
```
alexdanjou: In get_species_specific_settings, you return directly `models_country_flux` when it's a list (if I understood correctly). I guess you don't have to put a dict here (as the values are the same) and you can put directly the list, no?

fluxy/cli/config_annex_plot.py
```diff
@@ -267,26 +268,31 @@ class AnnexConfig:
...
+            resample=None,
+            rolling_mean=False,
+        )
+    }
 
-    ### Settings for spatial maps (list or dictionary if different between species)
+    ### Settings for spatial maps
+    ## Model definitions (list or dict["<period>": list(), "<species>": list()] if different between species)
+    # NOTE: it is assumed that models_spatial_maps exist in models_monthly/yearly_species
     models_spatial_maps = [
```
alexdanjou: `models_spatial_maps = models_country_flux` may be easier to understand what's happening? (ie we redefine only if there is a diff)

alexdanjou (approved):
Looks good! the new utils_annex_plot is great. I unfortunately still cannot test it (my ICOS Jupyter is still down).

fluxy/cli/config_annex_plot.py
melodb: Yes, you're right. I forgot about that particularity :p

PR #208, opened 2025-09-19
Reformat plot_country_flux and add option to aggregate by month

### LINKED ISSUE #207: Plot average seasonal timeseries of country totals
In order to better assess the seasonality of CH4 and N2O results, it would be interesting to plot the average country totals over all Jan/Fev/... for each model between sliced start_date and end_date. This plot would have the different months in the x-axis.
Either we add this option to plot_country_flux or we create a new plotting function. It would be nice to have also the options to plot separate/combined curves.

This type of plotting is also implemented for the spatial maps when setting `chop_by = [1,2,3,4,5,6,7,8,9,10,11,12]`. Maybe some functions used for data processing can re-used.

alexdanjou:
resolves issue #207 .

### DIFF
--- a/fluxy/plots/flux_timeseries.py
+++ b/fluxy/plots/flux_timeseries.py
@@ -3,9 +3,10 @@ import logging
 import numpy as np
 import xarray as xr
 import matplotlib.pyplot as plt
+from typing import Tuple
 
+from matplotlib.axes import Axes
 from matplotlib.figure import Figure
-from matplotlib.ticker import NullFormatter
 from matplotlib.dates import YearLocator, MonthLocator
 from matplotlib import __version__ as mplt_version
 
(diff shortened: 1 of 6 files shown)


alexdanjou:
I've pushed by mistake on devel before creating that branch.. Hopefully it gets merge soon and we don't have to do stuff and devel.

Main changes can be seen in first commit..

alexdanjou:
yes I confirm most of the changes are already on devel, I'm really bad with github :'(

fluxy/plots/flux_timeseries.py
```diff
@@ -20,6 +21,54 @@
...
+
+    return plot_regions
+
+
+def get_unit(ds_all: dict[str, xr.Dataset], species: str) -> str:
+    """
+    Determine unit of posterior estimations from datasets. If incoherencies between datasets, an error is raised.
+    Args:
+        ds_all: dictionnary of datasets from which read the units.
+        species: species of interest, used as "all_hfc"/"all_pfc" use a different name for the prior variable.
```
melodb: This part of species description "used as all_hfc/all_pfc use a different name for the prior variable" is not very clear to me. Even by looking at the code, I'm not sure what you mean.

fluxy/plots/flux_timeseries.py
```diff
@@ -81,9 +160,19 @@ def prepare_data_to_plot(
         expected_size=len(ds_all_region.keys()),
     )
 
-    model_colors = model_colors.copy()
-    # Assign default color list and label to input dataset
+    # Timely aggregate the data when necessary
+    if aggreg_month:
+        if any(resample):
+            raise ValueError(
+                f"`resample` and `aggreg_month` cannot be both set to True. Please correct."
```
melodb: Maybe instead of "Please correct" one can say "Please set only one of them to True."

fluxy/plots/flux_timeseries.py
```diff
@@ -169,6 +258,371 @@ def prepare_data_to_plot(
...
+
+def add_prior_plot(
+    ax: Axes, ds_region: xr.Dataset, annex_mode: bool, add_prior_unc: bool
+) -> dict[str, dict]:
+    """
+    Plot the variable prior of the dataset ds_region on the axis.
+    Axes:
+        ax: axes on which to plot
+        ds_region: dataset containing prior data
+        annex_bool: if True, the linewidth is made slighty bigger (1.5) and the transparency alpha is higher (1.0 -no transparency) than when False (resp. 1.0 and 0.7).
```
melodb: Maybe this comment can be made clearer? At least I had to look at the code to understand. Suggestion: "if True, the linewidth is made slighty bigger (1.5) and no transparency is applied to the prior uncertainty. If False, linewidth and alpha are set to standard values (1.0 and 0.7, respectively)." However, I don't understand why no transparency in the prior uncertianty would even be desirable. We have never plot it for the annexes.

fluxy/plots/flux_timeseries.py
```diff
@@ -169,6 +258,371 @@ def prepare_data_to_plot(
...
+        f"{s_data.get(species, {}).get('species_print', species)} {sector.title() if sector != 'total' else ''}"
+        f" ({unit.replace('2','$_{{2}}$').replace('-1','$^{{-1}}$')})"
+    )
+
+
+def set_xlims_and_ticks(
+    ax: Axes, yearly_freq: bool, res_dict: dict[str, dict], aggreg_month: bool
+):
+    """
+    Set x limits, ticks and ticks labels of matplotlib axes. Optimize them by looking at if theye monthly, yearly, or monthly aggregated, and covered time range.
```
melodb: Fix typo: "if theye" --> if they are

melodb (approved):
Thanks for this and for all the cleaning. I left very small comments. I did not try it because I'm in the middle of some plotting and I did not want to break my devel. But I will give it a try once this is merged.

fluxy/plots/flux_timeseries.py
alexdanjou: It is indeed not making any sense... I removed the parameter species. The function will now check that "posterior" or "flux_total_posterior_country" is present in every dataset.

fluxy/plots/flux_timeseries.py
alexdanjou: Indeed. I'll change it. And I also mixed everything in the despcription... When `annex_mode=True`, `linewidth=1.0` and `alpha=0.7`. So it's when we're not in annex mode that the prior uncertainty is not transparent. 
I'll change the description and let `linewidth` and `alpha` values as they are. But I'm also doubtful about `alpha=1.0`. Anyway we never set `add_prior_unc` to True..

PR #209, opened 2025-09-22
Fix bug in pr #208

alexdanjou:


fluxy/plots/flux_timeseries.py
```diff
@@ -757,6 +757,10 @@ def plot_country_flux(
 
         # plot inventory
         if plot_inventory:
+            if aggreg_month:
+                raise ValueError(
+                    "`plot_inventory` is not yet supported for monthly aggregate plots (`aggreg_month=True`)."
```
melodb: Maybe instead of raising an error and forcing the user to actively set plot_inventory = False, we could simply print a warning and set plot_inventory = False internally.

scripts/PARIS_inversion_results.ipynb
```diff
@@ -113,11 +113,11 @@
     "\n",
     "###################################\n",
     "### edit variables in this block\n",
-    "species = 'hfc134a' #options for individual species, or 'all_hfc' or 'all_pfc'\n",
+    "species = 'ch4' #options for individual species, or 'all_hfc' or 'all_pfc'\n",
```
melodb: I think this change in the notebook in unnecessary.

scripts/PARIS_inversion_results.ipynb
```diff
@@ -113,11 +113,11 @@
...
     "models = experiments['std_run'] #if using 'all_hfc' or 'all_pfc' use basic model names  (e.g. ['InTEM','ELRIS','RHIME']) and standard filenames (listed in config_data[\"models_info\"][\"standard_run\"][\"default\"]) will be used for each species. \n",
     "                                #To use non standard filenames, do <model>_<config_key> : e.g. with ['ELRIS_elris-n2o'], the filenames will be looked in config_data[\"models_info\"][\"standard_run\"][\"elris-n2o\"].\n",
     "regions = ['GERMANY','UK','BENELUX','NW_EU2']\n",
-    "period = 'yearly' #inversion period, must be a string or a list of the same length as models, e.g. ['monthly','yearly']\n",
+    "period = 'monthly' #inversion period, must be a string or a list of the same length as models, e.g. ['monthly','yearly']\n",
```
melodb: I think this change in the notebook in unnecessary.

scripts/PARIS_inversion_results.ipynb
```diff
@@ -880,7 +920,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.10.6"
+   "version": "3.12.2"
```
melodb: I would revert this change as well.

scripts/PARIS_inversion_results.ipynb
```diff
@@ -209,6 +209,46 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "###################################\n",
+    "### edit variables in this block\n",
+    "aggreg_month=True\n",
+    "###################################\n",
```
melodb: Is it really needed to create a new cell just for option aggreg_month = True? In my view this makes the notebook longer without much gain. Moreover, the user might want to use some of the option in the upper cell. They are not fully independent.

melodb (commented):
Thanks for fixing this! I left some small comments.

melodb (approved):

PR #216, opened 2025-09-30
mf monthly aggregation

alexdanjou:
This is the sort of plot this PR can create:
image
For now, I made it only available for "mf_xxx" variables. The uncertainty plotted is the 0.159-0.841 percentile range of the variable.

What else do we want?

1. 'posterior_above_BC'/'prior_above_BC' ?
2. 'observed_above_BC' ?
3. 'observed_posterior_diff'/'observed_prior_diff' ?
4. ...

@ag12733 @alisonredington @ayanu if you have suggestions. (I don't know Alistair username so couldn't put it here)


fluxy/plots/mf_timeseries.py
```diff
@@ -268,8 +295,8 @@ def plot_timeseries(
                 # Define uncertainty band
                 flag_fill_between = False
                 if unc_var.split("_")[0] == "percentile":
-                    y1 = ds_plot[unc_var][0, :].values
-                    y2 = ds_plot[unc_var][1, :].values
+                    y1 = ds_plot[unc_var].sel(percentile=0.159).values
+                    y2 = ds_plot[unc_var].sel(percentile=0.841).values
```
melodb: This change must be reverted. Percentile values might differ from 0.159/0.841. They are defined in the dataset in variable `percentile`. When `aggreg_month=True`, percentile variables in `ds_plot` should keep the same format so that `[0,:]` and `[1,:]` work as well.

fluxy/plots/mf_timeseries.py
```diff
@@ -179,25 +180,51 @@ def plot_timeseries(
...
+                if var.split("_")[0] != "mf":
+                    raise NotImplementedError(
+                        "`aggreg_month` disabled this for variable that are not mole fractions (i.e. names not starting with `mf`)."
+                    )
+                ds_plot = (
+                    ds_plot_org[[var]]
+                    .groupby("time.month")
+                    .mean()
+                    .rename({"month": "time"})
+                )
```
melodb: Variable `ds_plot` is defined inside the for-loop over vars to plot. Therefore, it does not conserve them. Would it be better to define a function (e.g. `prepare_mf_data_to_plot`) and call it outside this for-loop? In my view, `groupby("time.month").mean().rename({"month": "time"})` can be applied to all mf/stdev variables (only `percentile_*` excluded). Then, your quantile approach can be used to compute `percentile_{var}` for all variables (e.g. both `percentile_mf_observed` and `percentile_stdev_mf_total`). Other variables, like `mf_observed_above_bc_posterior`, can be computed inside this function (maybe for both aggreg_month = True/False; and maybe only if this variable is specified in include - if we make it general it might mess up with @lionel42's usage of this function with the eddy flux data).

fluxy/plots/mf_timeseries.py
```diff
@@ -301,7 +328,7 @@ def plot_timeseries(
         if ncols == 2:
             plot_histogram(
                 ax[iax, 1],
-                ds_plot,
+                ds_plot_org,
```
melodb: If all mf/std variables are grouped by month and defined in ds_plot, ds_plot can be used here instead of ds_plot_org. Maybe the histogram will be a bit useless, but at least the data would be consistent.

fluxy/plots/mf_timeseries.py
```diff
@@ -357,10 +384,10 @@ def plot_timeseries(
             for l in leg.legendHandles:
                 l.set_linewidth(5.0)
 
-        if len(ds_plot["time"]) <= 1:
+        if len(ds_plot_org["time"]) <= 1:
             continue
-        start_date = ds_plot["time"].values.min()
-        end_date = ds_plot["time"].values.max()
+        start_date = ds_plot_org["time"].values.min()
+        end_date = ds_plot_org["time"].values.max()
```
melodb: xlims is currently not working for aggreg_month = True. Maybe a new function can be defined (e.g. add_mf_xlims_and_ticks) and called here.

melodb (commented):
Thanks for working on this so quickly. I left a couple of comments. In addition, I think the figure length should be reduced when aggreg_month=True. Otherwise, it's hard to see a seasonal variation. Moreover, if aggreg_month=True, maybe a title should be added to the plot with the start/end dates.

fluxy/plots/mf_timeseries.py
alexdanjou: Ok, thanks for the quick reply! Is the variable always called percentile? if so I'll put `.isel(percentile=0)`. It will be better so we don't have to bother with dimension being swap (which is happening when `groupby` is used when aggregating by month).

alexdanjou:
New look for the plots:
image

alexdanjou:
Sorry I ran again the formatter on all the files... I've just made changes in `plots.mf_timeseries.py` if it helps reviewing :)

alexdanjou:
If it's to painful to review, I'll create a new branch, copy paste this `plot/mf_timeseries.py` in it (it will also help me resove the conflict issue) and make a new PR.

melodb:
> If it's to painful to review, I'll create a new branch, copy paste this `plot/mf_timeseries.py` in it (it will also help me resove the conflict issue) and make a new PR.

As you prefer. I think the conflict is only in mf_timeseries anyway

fluxy/plots/mf_timeseries.py
```diff
@@ -54,6 +55,426 @@ def plot_mf_timeseries(*args, **kwargs) -> plt.Figure:
     return plot_timeseries(*args, **kwargs)
 
 
+def _prepare_aggreg_month_var(da_var: xr.DataArray | xr.Dataset) -> xr.Dataset:
+    """
+    Aggregate by month the dataset/array variable(s). The outputed dataset as two dimensions: "time" (array from 1 to 12)
```
melodb: Typo: "has" two dimensions

fluxy/plots/mf_timeseries.py
```diff
@@ -54,6 +55,426 @@ def plot_mf_timeseries(*args, **kwargs) -> plt.Figure:
...
+        ds: dataset to infer the new variable from
+        var: new variable to infer. Options are '<var1>_above_BC' and '<var1>_<var2>_diff' with <var1> and <var2> being one of
+            'prior' (for 'mf_prior'), 'posterior' (for 'mf_posterior'), 'observed' (for 'mf_observed').
+        unc_var: uncertainty to associated with the new variable. Currently not implemented. Will raise an error if not None.
+    Returns:
+        ds: dataset with new variable inside.
+    """
+    if unc_var:
+        raise NotImplementedError(
+            "No uncertainty can be plotted when the variable is inferred from others."
```
melodb: I would link the error message to the input and give instructions on how to fix it. For example: "No uncertainty can be plotted on {var} because this variable will be inferred from others. Please set include to '{var}': None"

fluxy/plots/mf_timeseries.py
```diff
@@ -54,6 +55,426 @@ def plot_mf_timeseries(*args, **kwargs) -> plt.Figure:
...
+    time_freq_min: FrequencyType,
+    plot_type: Literal["separate", "together", "diff"],
+) -> dict[str, xr.Dataset]:
+    """
+    Create dictionnary of datasets containing all the data that will be plotted.
+    Args:
+        ds_all: dictionnary of dataset from which the variable are taken
+        include: variables to plot in the main panel. If is a dictionnary : the keys are the variables to plot and the
+            values the uncertainty that will be shaded around them.
+        diff_include: additionnal (or not) variables to the one in include.
```
melodb: Rephrase: "additional variables to those in include."
In my view "(or not)" does not add clarity.

fluxy/plots/mf_timeseries.py
```diff
@@ -54,6 +55,426 @@ def plot_mf_timeseries(*args, **kwargs) -> plt.Figure:
...
+            "The include dictionary is empty. Please provide variables to include in the plot."
+        )
+    if isinstance(include, str):
+        include = {include: None}
+    elif isinstance(include, (list, tuple)):
+        include = {var: None for var in include}
+
+    data_to_plot = {m: xr.Dataset() for m in ds_all.keys()}
+    if isinstance(diff_include, list):
+        include.update({var: None for var in diff_include if var not in include.keys()})
```
melodb: This is not how `diff_include` is supposed to work. `diff_include` is only relevant for the histogram, not for the timeseries. If `diff_include` is an empty list, the histogram of all variables in `include` should be plotted. If `diff_include` is a non empty list, for instance equal to "mf_posterior", the histogram of "mf_posterior" - "mf_observed" will be plotted. This is currently done inside `plot_histogram`.

fluxy/plots/mf_timeseries.py
```diff
@@ -54,6 +55,426 @@ def plot_mf_timeseries(*args, **kwargs) -> plt.Figure:
...
+                    and not var.endswith("_diff")
+                ):
+                    raise NotImplementedError(
+                        "`aggreg_month` disabled this for variable that are not mole fractions (i.e. names not starting with `mf`)."
+                    )
+                ds_var = _prepare_aggreg_month_var(ds_clean[var])
+                if include[var] is not None:
+                    logger.warning(
+                        f"`{include[var]}` present as value of include dict for {var} is overwritten as you put `aggreg_month=True`."
+                        + " The uncertainty plotted is the 0.159 and 0.851 percentile of the variable for the corresponfing month."
```
melodb: Typo: corresponding

fluxy/plots/mf_timeseries.py
```diff
@@ -54,6 +55,426 @@ def plot_mf_timeseries(*args, **kwargs) -> plt.Figure:
...
+                    )
+            else:
+                ds_var = _prepare_var(ds_clean, var, unc_var)
+
+            if unc_var:
+                if plot_type == "diff":
+                    raise ValueError(
+                        f"Option plot_type='diff' does not accept uncertainties. Replace '{unc_var}' by None."
+                    )
```
melodb, line 292: Maybe here I would set `unc_var = None` and print a warning instead of forcing the user to manually set it to None.

fluxy/plots/mf_timeseries.py
```diff
@@ -54,6 +55,426 @@ def plot_mf_timeseries(*args, **kwargs) -> plt.Figure:
...
+    ds_dict: dict[str, xr.Dataset],
+    model_labels: dict[str, str],
+    model_colors: dict[str, list],
+    plot_type: Literal["separate", "together", "diff"],
+) -> dict[str, xr.Dataset]:
+    """
+    Set labels and colors that will be used by plot_timeseries and plot_histogram as attributes of the variables dataset.
+    For variables "mf_observed" and "observed_above_BC", the color will be black (and not one of model_colors) if more than one variable is plotted.
+    Args:
+        ds_dict: dictionnary containing the dataset with the variable to be plotted (and only them).
```
melodb: Typo: "variables" (plural) instead of "variable"

fluxy/plots/mf_timeseries.py
```diff
@@ -54,6 +55,426 @@ def plot_mf_timeseries(*args, **kwargs) -> plt.Figure:
...
+            if var in ["mf_observed", "observed_above_BC"] and len(vars_to_plot) > 1:
+                plot_color = "black"
+
+            plot_label = (f"{model_label} {config.mf_labels.get(var, var)}",)
+
+            ds_dict[m][var].attrs.update(
+                {"plot_label": plot_label, "plot_color": plot_color}
+            )
+
+        ds_dict[m].attrs["color"] = model_color
```
melodb: Why do you need this if you already set the color per variable?

fluxy/plots/mf_timeseries.py
```diff
@@ -54,6 +55,426 @@ def plot_mf_timeseries(*args, **kwargs) -> plt.Figure:
...
+    plot_units = list()
+
+    for m in ds_dict:
+        for var in ds_dict[m].data_vars:
+            plot_units.append(ds_dict[m][var].attrs["units"])
+
+    plot_units = list(set(plot_units))
+    if len(plot_units) != 1:
+        raise ValueError(
+            f"{ds_dict[m].data_vars.keys()} in {ds_dict.keys()} do not have the same units. So far, the following were found: {plot_units}."
```
melodb, line 411: I would add the following to the error message: "Select only one model to plot, or run 'slice_mf' with 'mf_units_print' equal to a valid mole fraction unit before running 'plot_timeseries'."

fluxy/plots/mf_timeseries.py
```diff
@@ -54,6 +55,426 @@ def plot_mf_timeseries(*args, **kwargs) -> plt.Figure:
...
+def add_xlims_and_ticks(
+    ax: Axes, yearly_freq: bool, res_dict: dict[str, dict], aggreg_month: bool
+):
+    """
+    Add x limits, ticks and ticks labels to matplotlib axes. Optimize them by looking at if they are monthly, yearly, or monthly aggregated, and covered time range.
+    Args:
+        ax: axis to add xlim and xticks to.
+        yearly_freq: set to True if the data plotted have a yearly frequency.
+        res_dict: dictionnary containing the data plotted. Should have one key per regions plotted, the values being dictionnaries with 3 keys: "inventory", "posterior" and "prior";
+            whose values are the output of add_inventory_barplot, add_posterior_plot, add_prior_plot). The time data stored in them is used to infer the xlims.
```
melodb: I guess these should be deleted?

fluxy/plots/mf_timeseries.py
```diff
@@ -54,6 +55,426 @@ def plot_mf_timeseries(*args, **kwargs) -> plt.Figure:
...
+            time = res_dict[key_1].time.values.astype("datetime64[D]").tolist()
+            min_x = np.nanmin([*time, min_x])
+            max_x = np.nanmax([*time, max_x])
+
+    # set xticks
+    year_range = date(max_x.year, 1, 1) - date(min_x.year, 1, 1)
+    if yearly_freq:
+        min_x = date(min_x.year, 1, 1)
+        max_x = date(max_x.year + 1, 1, 1)
+    xlim = [min_x - (max_x - min_x) / 50, max_x + (max_x - min_x) / 50]
```
melodb, line 457: I have the impression that none of this applies to the mf plots. Am I right? The frequency of the data is only a relevant parameter for the country fluxes.

fluxy/plots/mf_timeseries.py
```diff
@@ -344,47 +658,42 @@ def plot_timeseries(
...
-            int(end_date.astype("datetime64[M]") - start_date.astype("datetime64[M]"))
-            > 12
-        ):
-            ax[iax, 0].xaxis.set_minor_locator(MonthLocator())
-            ax[iax, 0].xaxis.set_minor_formatter(NullFormatter())
-            ax[iax, 0].xaxis.set_major_locator(YearLocator())
-        else:
-            ax[iax, 0].xaxis.set_major_locator(MonthLocator())
-            if presentation_mode:
-                ax[iax, 0].tick_params(axis="x", rotation=70)
```
melodb, line 374: The rotation of the ticks might be needed when using presentation_mode (large fonts).

fluxy/plots/mf_timeseries.py
```diff
@@ -54,6 +55,426 @@ def plot_mf_timeseries(*args, **kwargs) -> plt.Figure:
...
+        logger.warning(
+            f"Variable {unc_var_in} not found {ds.attrs.get('model','')} so reading uncert from {unc_var}."
+        )
+
+    # Creating variable
+    if unc_var.split("_")[0] == "percentile":
+        unc = ds[unc_var].to_dataset()
+        unc["percentile"] = ["lower", "upper"]
+    elif unc_var.split("_")[-1] in ["prior", "posterior"]:
+        unc_lower = (ds[var] - ds[unc_var]).expan_dims(
```
melodb: When I set `unc_var != None`, the code breaks here with error message: `AttributeError: 'DataArray' object has no attribute 'expan_dims'`. I get this problem with both percentile_* and stdev_* variables

melodb (commented):
Thanks Alex for restructuring plot_timeseries! It was indeed getting quite messy. I left small comments in the code itself - there are some small things to be fixed. Moreover, I was wondering if maybe we should create a new file (e.g. named "utils_mf.py") and move all the new auxiliary functions there. This would ruin your internal function strategy but I think it will make mf_timeseries more manageable.

melodb:
As mf_timeseries was considerably restructured in this PR, I have requested a review from @lionel42

fluxy/plots/mf_timeseries.py
alexdanjou: Yes it is a function that will be in common with flux timeseries (it's in the PR, I'm waiting ALice to review, I'll write to her about it to sort that quickly).

fluxy/plots/mf_timeseries.py
alexdanjou: I changed it to "variables that will be plot in the secondary (histogram) panel. In this function, they are treated as the ones passed with `include` parameter."

fluxy/plots/mf_timeseries.py
alexdanjou: This is does not change how the data are prepared :)
Unless I missed something, `include` is not updated in the main body of the function. So it won't add the variables for the plotting. We just prepare all the data the same way.

fluxy/plots/mf_timeseries.py
alexdanjou: I supressed the 2 last sentences of the description of `res_dict`. Is that what you meant?

melodb: Honestly, I don't understand my comment either. Just ignore it. Sorry for that.

fluxy/plots/mf_timeseries.py
alexdanjou: My mistake, I did missed something... Good catch! It'll be corrected in the next push!

fluxy/plots/mf_timeseries.py
alexdanjou: typo, it should be indeed `expand_dims` ! Good catch, thanks :)

alexdanjou: (but there seems to be other errors on that, will correct them...)

alexdanjou:
Hey @melodb and @hdelongueville! The PR is ready again if you want to review it :)

fluxy/plots/mf_timeseries.py
alexdanjou: Ahaha no no it was a fair point! I understood it afterwards (I think). The 2 last sentences had nothing to do there for sure. But I wasn't understanding why the line before was included as well. I think it was related to your next comment (as this variable is not used in mf timeseries but just in flux timeseries).

alexdanjou:
oups no the tests have failed so not ready I guess..

alexdanjou:
and now it's ready

fluxy/plots/mf_timeseries.py
alexdanjou: I disagree with that. If I make `plot_timeseries` modify under the hood a value of the include dictionnary it will prpagate into the dictionnary and it might be unseen. The solution would be to copy the parameters at the start of the function and work on a copy of it, but it will become quite heavy. to me it's better if it stays that way.

However, as you are the main user, if you confirm that you prefer the way you said for user-friendliness, I'll do it :)

melodb: Ok, that's fine for me ;)

alexdanjou:
> Thanks Alex for restructuring plot_timeseries! It was indeed getting quite messy. I left small comments in the code itself - there are some small things to be fixed. Moreover, I was wondering if maybe we should create a new file (e.g. named "utils_mf.py") and move all the new auxiliary functions there. This would ruin your internal function strategy but I think it will make mf_timeseries more manageable.

I agree with putting in a separate file. But I would prefer to merge and open a new issue. That way I'll also put the xticks and lims function in common with flux timeseries

fluxy/plots/mf_timeseries.py
```diff
@@ -54,6 +55,441 @@ def plot_mf_timeseries(*args, **kwargs) -> plt.Figure:
...
+                ]
+            }
+        )
+        unc = xr.merge([unc_lower, unc_upper], compat="no_conflicts", join="outer")
+    else:
+        unc = ds[unc_var].expand_dims(
+            {
+                "percentile": [
+                    "std",
+                ]
```
lionel42: remove the comma in these lists, to have a better formatting on fewer lines

fluxy/plots/mf_timeseries.py
```diff
@@ -54,6 +55,441 @@ def plot_mf_timeseries(*args, **kwargs) -> plt.Figure:
...
+            bc = ds["mf_bc_prior"]
+
+        var0 = var.split("_")[0]
+        if var0 not in acceptable_var:
+            raise NotImplementedError(
+                "The variable names currently available when plotting a variable value above BC are of the form '<var>_above_BC', "
+                "with var being one of 'prior' (for 'mf_prior'), 'posterior' (for 'mf_posterior'), 'observed' (for 'mf_observed')."
+            )
+        ds[var] = ds["mf_" + var0] - bc
+        ds[var].attrs = ds["mf_" + var0].attrs
```
lionel42: use fstrings for formatting names

fluxy/plots/mf_timeseries.py
```diff
@@ -54,6 +55,441 @@ def plot_mf_timeseries(*args, **kwargs) -> plt.Figure:
...
+
+    elif var.endswith("_diff"):
+        var1, var2 = var.split("_")[:2]
+        if var1 not in acceptable_var or var2 not in acceptable_var:
+            raise NotImplementedError(
+                "The variable names currently available when plotting a difference of two variables are of the form '<var1>_<var2>_diff', "
+                "with var1 and var2 being one of 'prior' (for 'mf_prior'), 'posterior' (for 'mf_posterior'), 'observed' (for 'mf_observed')."
+            )
+        ds[var] = ds["mf_" + var1] - ds["mf_" + var2]
+        ds[var].attrs = ds["mf_" + var1].attrs
```
lionel42: fstrings here as well

fluxy/plots/mf_timeseries.py
```diff
@@ -54,6 +55,441 @@ def plot_mf_timeseries(*args, **kwargs) -> plt.Figure:
...
+        step = (max_x.year - min_x.year) // 8 + 1
+        xticks = [date(year, 1, 1) for year in range(min_x.year, max_x.year, step)]
+        if (max_x.year - min_x.year) % step == 0:
+            xticks = np.append(xticks, max_x)
+        ax.set_xticks(xticks)
+        ax.set_xticklabels(xticks.astype("datetime64[Y]"))
+        ax.xaxis.set_major_locator(YearLocator())
+    else:
+        ax.xaxis.set_minor_locator(MonthLocator())
+        ax.xaxis.set_major_locator(YearLocator())
```
lionel42: move outsied of if else, if you apply it in both cases

lionel42 (changes requested):
Thanks this is a very good refactoring. 

Please fix the minor comments I left.

I also noticed that we should discuss about enforcing black formatting in the whole code at some point, maybe for a separate issue.

fluxy/plots/mf_timeseries.py
melodb: expan_dims (instead of `expand_dims`) can still be found in the code. Here and in the calculation of unc_upper.

fluxy/plots/mf_timeseries.py
```diff
@@ -54,6 +55,434 @@ def plot_mf_timeseries(*args, **kwargs) -> plt.Figure:
...
+    rotate_xticks: bool = False,
+):
+    """
+    Add x limits, ticks and ticks labels to matplotlib axes. Optimize them by looking at if they are monthly, yearly, or monthly aggregated, and covered time range.
+    Args:
+        ax: axis to add xlim and xticks to.
+        yearly_freq: set to True if the data plotted have a yearly frequency.
+        res_dict: dictionnary containing the data plotted.
+        aggreg_month: if True, the data plotted are supposed to be a monthly aggregated so 12 ticks are created, whose labels are the 3 first letters of each month.
+    """
```
melodb, line 433: Add `rotate_xticks` to list of arguments.

melodb (changes requested):
I have tested the code and even after fixing expan_dims to expand_dims, I get the following error:
`ValueError: unable to convert unnamed DataArray to a Dataset without providing an explicit name
`
In _prepare_data_to_plot, L.295:
`unc = xr.merge([unc_lower, unc_upper], compat="no_conflicts", join="outer")`

Once this is fixed, I can try it again.

lionel42 (approved):

alexdanjou:
> I have tested the code and even after fixing expan_dims to expand_dims, I get the following error: `ValueError: unable to convert unnamed DataArray to a Dataset without providing an explicit name ` In _prepare_data_to_plot, L.295: `unc = xr.merge([unc_lower, unc_upper], compat="no_conflicts", join="outer")`
> 
> Once this is fixed, I can try it again.

Sorry about this `expan_dims` still present. I was sure I had fixed it, along with the dataarray/dataset error. MAybe I got mix with between my branches :/
Can you test it again, and if you have still this error, tell me with what data you're using, and what parameters you passed to the function.

Thanks!

melodb (approved):
All working fine now. Thanks a lot!

PR #220, opened 2025-10-15
Update annex

melodb:
**Config:**
- add limits to spatial maps as input
- update start dates according to PARIS meeting
- remove RHIME runs from SF6

**Annex code:**
- print data to table according to table_start_date
- Replace NIR by NID
- Fix species names in main table
- Use inventory year for table caption

**fluxy:**
- Fix number of columns in label if `annex_mode=True`

### DIFF
--- a/fluxy/cli/annex_plot_generator.py
+++ b/fluxy/cli/annex_plot_generator.py
@@ -186,7 +186,12 @@ def produce_plots(
             plt.close()
 
             # Store results for .csv and table
-            annual_res = dict_to_str_dataframe(res_dict, inventory_years, species)
+            annual_res = dict_to_str_dataframe(
+                res_dict,
+                inventory_years,
+                species,
+                table_start_date=annex_config_data.start_date_table,
+            )
             annual_res_list.append(annual_res)
 
             # 1.2) Plot monthly country fluxes over PARIS time window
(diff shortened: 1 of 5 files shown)


fluxy/cli/annex_plot_generator.py
```diff
@@ -211,7 +220,10 @@ def produce_plots(
         else:
             # Store results for .csv and table
             annual_res = dict_to_str_dataframe(
-                res_dict[region], inventory_years, species
+                res_dict,
```
alexdanjou, line 223: res_dict[region] to res_dict seems an important change. Just want to make sure that you're sure of it ^^ But I guess your correction is right as it make it the same as the other place :)

Sorry again but I still cannot properly test as my jupyterlab is still broken

alexdanjou (approved):

fluxy/cli/annex_plot_generator.py
melodb: That's related to a change you did in the code back then. As far as I understood, now the first key of res_dict is prior/posterior/inventory and only then the region

melodb: L.802 in plot_country_flux:

    res_dict: dict[str, dict] = {
        "inventory": inventory_data,
        "posterior": posterior_data,
        "prior": prior_data,
    }

alexdanjou: I really don't like this dictionnary of dictionnary (I know I made it). Once this annex round is over if you agree I'll make it directly a dataframe.

melodb: I have completely lost the data structure of fluxy at this moment :p I trust your judgement

PR #221, opened 2025-10-17
Update annex color scale

melodb:
- update percentiles and country map limits
- fix in plot_country_flux to plot only rolling mean
- Apply rolling mean to InTEM longrun


fluxy/plots/flux_timeseries.py
```diff
@@ -215,9 +215,8 @@ def prepare_data_to_plot(
         ds_all_region.keys(), rolling_mean, plot_separate, resample
     ):
         if rm & ps:  # if rolling_mean and plot_separate
-            ds_to_plot[m] = calc_rolling_mean(
-                ds_to_plot[m + "_resample"] if rs else ds_to_plot[m]
-            )
+            model = m + "_resample" if rs else m
+            ds_to_plot[model] = calc_rolling_mean(ds_to_plot[model])
```
alexdanjou, line 219: Just to check I understood : you have a case when you want to resample the data and then apply a smoothing (/rolling mean) on them?

alexdanjou (approved):

fluxy/plots/flux_timeseries.py
melodb: Exactly. I want to resample InTEM monthly CH4 to yearly and then apply a rolling mean. Before, both lines were being plotted (with and without the rolling mean).

PR #222, opened 2025-10-17
Change format of results outputed by plot_country_flux and table number of digits inn the annex

alexdanjou:


fluxy/io.py
```diff
@@ -342,11 +342,8 @@ def read_model_output(
         if add_sites_to_flux and file_type == DataTypes.FLUX:
             ds_all[m] = add_sites_var(ds_all[m], filepath, m, period[i], config_data)
 
-        # Add species in attributes if not present
-        if "species" not in ds_all[m].attrs:
-            ds_all[m].attrs["species"] = species
-        elif "RHIME" in m and ds_all[m].attrs["species"]=="inert":
-            ds_all[m].attrs["species"] = species
+        # Overwrite species attributes
+        ds_all[m].attrs["species"] = species
```
melodb: I'm not sure I agree with this change. Why do you need to overwrite the species attribute in all files?

alexdanjou: It's fixing the bug you had with HFC23 when you tried the PR on Friday. For InTEM there is a typo and the species in the attributes is "hf23" (and not "hfc23") so the species attribute is ditched when the datasets are combined to make the PARIS mean (which causes a bug further down). Also, there is a bug in RHIME postprocessing and all our species attribute are set to "inert"...

I wanted to change this first thing today to rather check if the attribute was the same and overwrite and print a warning if it wasn't, but you were too quick for me XD

fluxy/cli/utils_annex_plot.py
```diff
@@ -33,99 +31,123 @@ def get_species_specific_settings(
 
     return settings_species
 
-
-def dict_to_str_dataframe(
+def create_str_dataframe(
```
melodb, line 34: Write function description

fluxy/cli/utils_annex_plot.py
```diff
@@ -33,99 +31,123 @@ def get_species_specific_settings(
...
-        paris_val = f"{comb['mean'][it]:.{n_digits}f} \\pm {(comb['max'][it]-comb['min'][it])/2:.{n_digits}f}"
-        inv_val = inv["value"][inv["time"].astype("datetime64[Y]") == time]
-        if len(inv_val) == 1:
-            output[str(time)] = [f"{inv_val[0]:.{n_digits}f}", paris_val]
+    if not sector:
+        if res.sector.unique().size!=1:
+            raise ValueError(
+                f"`sector` parameter should be provided when there is more than one region in `res` (currently present: {res.sector.unique()})."
+            )
+        sector = res.sector.unique()[0]
```
melodb, line 77: "sector" is never used. Should we delete this or restrict `res` to `res.sector == sector`?

fluxy/cli/utils_annex_plot.py
```diff
@@ -33,99 +31,123 @@ def get_species_specific_settings(
...
+        if max_exp<-1 and max_exp>=-4:
+            for var in ["mean_val","min_unc","max_unc"]:
+                data_per_species[var] *= 1e3
+            data_per_species["units"] = "$\\rm{GgCO}_{2}\\rm{-eq} \\cdot \\rm{yr}^{-1}$"
+            max_exp += 3
+        elif max_exp<-4:
+            for var in ["mean_val","min_unc","max_unc"]:
+                data_per_species[var] *= 1e6
+            data_per_species["units"] = "$\\rm{MgCO}_{2}\\rm{-eq} \\cdot \\rm{yr}^{-1}$"
+            max_exp += 6
```
melodb, line 110: This part of the code assumes that `country_flux_units_print = "Tg CO2-eq yr-1"` in config_annex_plot.py which might not always be the case. I don't think you need to loose the time to make it generic but maybe you can raise an error if different units are used for `country_flux_units_print`.

fluxy/cli/utils_annex_plot.py
```diff
@@ -33,99 +31,123 @@ def get_species_specific_settings(
...
+    data["val"] = data.apply(lambda x : x.mean_val if x.type=="inventory"
+                             else f"{x.mean_val} \\pm {x.unc}",
+                             axis=1)
+
+    output = data.pivot(index=["model","species","units"],columns="year",values = "val").reset_index()
+    output.columns.name = None
+    output.fillna(" ",inplace=True)
+
+    output.rename(columns={"model":"source"},inplace=True)
+    output["source"] = output["source"].apply(lambda x: x.replace("inventory_","NIR "))
```
melodb: It should be NID

fluxy/plots/flux_timeseries.py
```diff
@@ -332,11 +335,13 @@ def add_prior_plot(
...
-        "mean": ds_region.prior.values,
-    }
-
+    res = pd.DataFrame({"type":["prior",]*ds_region.time.size,
+                        "model":[ds_region.attrs["model_label"],]*ds_region.time.size,
+                        "sector":[ds_region.attrs["sector"],]*ds_region.time.size,
+                        "country":[ds_region.attrs["country"],]*ds_region.time.size,
+                        "species":[ds_region.attrs["species"],]*ds_region.time.size,
+                        "time": ds_region.time.values.astype("datetime64[ns]"),
+                        "mean_val": ds_region.posterior.values,})
```
melodb: I guess the following should be fixed: "mean_val": ds_region.**prior**.values

fluxy/plots/flux_timeseries.py
```diff
@@ -467,19 +477,10 @@ def add_ylim(
...
-            np.nanmax(post["max"]) for post in res_dict["posterior"][country].values()
-        ]
-        maxs_prior = [
-            np.nanmax(prior.get("max", prior.get("mean", np.nan)))
-            for prior in res_dict["prior"][country].values()
-        ]
-
-        max_cf.append(np.nanmax([*maxs_inventory, *maxs_posterior, *maxs_prior]))
+        df_country = plotted_data_df[plotted_data_df.country==country]
+        max_country = np.nanmax(df_country[["mean_val","min_unc","max_unc"]])
```
melodb: I guess `min_unc` does not need to be on the list

fluxy/io.py
```diff
@@ -342,6 +342,12 @@ def read_model_output(
         if add_sites_to_flux and file_type == DataTypes.FLUX:
             ds_all[m] = add_sites_var(ds_all[m], filepath, m, period[i], config_data)
 
+        # Overwrite species attributes
+        current_species = ds_all[m].attrs.get("species","not set")
+        if current_species!=species:
+            logger.warning(f"'species' attribute in dataset ({current_species}) differs from species {species}. It is overwritten.")
```
melodb: I think it would be good to output the model `m` leading to the warning.
Moreover, could you downgrade to info? Otherwise we will have this message whenever we plot RHIME. species is not a mandatory attribute so I guess it's fine downgrading to info.

fluxy/cli/utils_annex_plot.py
```diff
@@ -33,99 +31,123 @@ def get_species_specific_settings(
...
+
+    output["sort_col1"] = output.species.apply(lambda x : species_order.get_loc(x))
+    output["sort_col2"] = output.source.apply(lambda x : 0 if x==model else 1)
+    output.sort_values(by=["sort_col1","sort_col2"], inplace=True)
+    del output["sort_col1"], output["sort_col2"]
+
+    species_name = {"ch4":"CH_4", "n2o":"N_2O", "sf6": "SF_6", "nf3": "NF_3", "cf4": "PFC-14", "all_pfc": "Total PFC", "all_hfc": "Total HFC"}
```
melodb: The underscore won't work properly in latex. I guess it should be "CH$_4$", etc.

melodb (commented):
Thanks a lot for working on this! I left some small comments. Apart from that everything is working fine.

fluxy/cli/utils_annex_plot.py
alexdanjou: Indeed, I've had `res.sector == sector`

fluxy/cli/utils_annex_plot.py
alexdanjou: Good point.
I put a note in the description. But a correction will need to wait. The unit attribute does not mke it into the combined dataset apparently. So it will need a bit of digging to fix.
I suggest going on with this and opening an issue about it.

melodb (approved):
Working fine!

PR #229, opened 2025-10-29
Add new function plot_flux_map_combined_models_comparison

### LINKED ISSUE #226: Add a flux map function to plot the difference between average models
Add a function similar to `plot_flux_map_model_comparison`, but enabling to compare two groups of averaged models instead of two individual models. 

I already have a function called 'plot_flux_map_network_effect2' in a personal branch which does that (see example below). I am using it for the HFC paper and as we also used it for Italian and Norwegian NID Annexes, I thought I'll clean it and add it to `devel`.

Image

hdelongueville:
This PR adds a new function `plot_flux_map_combined_models_comparison`, which plots the spatial distribution of a selected flux variable for two groups of models and their difference. The function combines model outputs within each group, aggregates them over time (optionally seasonally), and produces a three-panel map.

- [x] Close #226 
- [x] Add new test
- [x] Update Notebook


melodb (approved):
Thanks for adding this. All is working fine. Should a new cell with this function be added to the notebook?

hdelongueville:
Yes that is a good idea! I'll add a new cell this week.

PR #232, opened 2025-11-06
Alternative flux filename

ayanu:
I suggest to change our default filenames for the flux files to "_flux.nc" 

In order to be compatible, I added this as an alternative ending in file_pattern, which is then called in read_model_output

Could this be merged with priority? I would like to submit this year's PARIS results with the new file endings.

### DIFF
--- a/fluxy/io.py
+++ b/fluxy/io.py
@@ -320,14 +320,14 @@ def read_model_output(
 
         # Check if file exists
         if not filepath.is_file():
-            logger.warning(f"Cannot find {file_type.value} file: {filepath}. Will try altnative name.")
             #  alternative filename with _flux ending
             if file_type==DataTypes.FLUX:
+                logger.warning(f"Cannot find {file_type.value} file: {filepath}. Will try alternative name.")
                 filepath = get_filename(
                 m,
                 species,
                 period_str,
-                file_pattern(file_type, alternativ=True),
+                file_pattern(file_type, alternative=True),
                 config_data,
                 data_dir,
                 read_standard_run,
--- a/fluxy/io.py
+++ b/fluxy/io.py
@@ -320,6 +320,7 @@ def read_model_output(
 
         # Check if file exists
         if not filepath.is_file():
+            logger.warning(f"Cannot find {file_type.value} file: {filepath}. Will try altnative name.")
             #  alternative filename with _flux ending
             if file_type==DataTypes.FLUX:
                 filepath = get_filename(
--- a/fluxy/io.py
+++ b/fluxy/io.py
@@ -320,8 +320,20 @@ def read_model_output(
 
         # Check if file exists
         if not filepath.is_file():
-            logger.warning(f"Cannot find {file_type.value} file: {filepath}.")
-            continue
+            #  alternative filename with _flux ending
+            if file_type==DataTypes.FLUX:
+                filepath = get_filename(
+                m,
+                species,
+                period_str,
+                file_pattern(file_type, alternativ=True),
+                config_data,
+                data_dir,
+                read_standard_run,
+            )
+            if not filepath.is_file():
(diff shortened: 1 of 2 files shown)


fluxy/io.py
```diff
@@ -320,8 +320,21 @@ def read_model_output(
 
         # Check if file exists
         if not filepath.is_file():
-            logger.warning(f"Cannot find {file_type.value} file: {filepath}.")
-            continue
+            logger.warning(f"Cannot find {file_type.value} file: {filepath}. Will try altnative name.")
```
melodb: Typo: "alternative"
Moreover, I would not add the message "Will try alternative name" here. Otherwise, it will be printed for the concentration files as well. I would add it inside the next if-statement.

fluxy/io.py
```diff
@@ -320,8 +320,21 @@ def read_model_output(
...
-            logger.warning(f"Cannot find {file_type.value} file: {filepath}.")
-            continue
+            logger.warning(f"Cannot find {file_type.value} file: {filepath}. Will try altnative name.")
+            #  alternative filename with _flux ending
+            if file_type==DataTypes.FLUX:
+                filepath = get_filename(
+                m,
+                species,
+                period_str,
+                file_pattern(file_type, alternativ=True),
```
melodb: Typo: "alternative"

fluxy/io.py
```diff
@@ -320,8 +320,21 @@ def read_model_output(
...
+                config_data,
+                data_dir,
+                read_standard_run,
+            )
+            if not filepath.is_file():
+                logger.warning(f"Cannot find {file_type.value} file: {filepath}.")
+                continue
```
melodb, line 337: I would move this if-statement inside the previous one (if ... FLUX:). Otherwise, message "Cannot find..." will be printed twice for the concentration files.

fluxy/types.py
```diff
@@ -14,11 +14,28 @@ class DataTypes(Enum):
...
-    """Returns the ending pattern for the given file type."""
+def file_pattern(
+    file_type: DataTypes, 
+    alternative: bool = False
+) -> str:
+    """
+    Returns the ending pattern for the given file type.
+    Args:
+        file_type (DataTypes):
+            Type of file (flux or conentration as defined in class DataTypes)
```
melodb: Typo: "concentration"

melodb (commented):
I left some small comments. I think you also need to fix the tests. If you need my help for that just let me know.

fluxy/io.py
ayanu: moved logging message

fluxy/io.py
ayanu: fixed

fluxy/io.py
ayanu: I will move first if statement into if FLUX. Then it should behave as I had intended. 
One warning message for concentration file and potentially two for flux file.

fluxy/types.py
ayanu: fixed

ayanu:
Tests did not pass because of a syntax error. I should not assume that I can write a single line of python code without making a mistake.

fluxy/types.py
```diff
@@ -34,7 +34,7 @@ def file_pattern(
 
         # Default file type
         return ".nc"
-        if alternative
+        if alternative:
             return "_flux.nc"
```
melodb, line 36: The code will never get here. The if-statement needs to be moved above the first return.

ayanu: How did I do this. That was not intended.

ayanu:
I hope it is all correct now. I promise not to touch the code anymore.

melodb (approved):
All working fine for me. Ready to be merged.

PR #233, opened 2025-11-06
Add options to plot separate and/or combined uncertainties in flux_timeseries

### LINKED ISSUE #150: Add an option to show the separate and/or combined uncertainties in `plot_country_flux`
Currently, when setting `plot_separate = True` and `plot_combine = True` in `plot_country_flux`, both the combined uncertainties and the separated uncertainties are being plotted. I find the figure a bit hard to read (see example below). 

I think it would be nice to have the options add_posterior_unc_separated and add_posterior_unc_combined to be able to chose which uncertainty to plot. 

Open to any other suggestions :)

Image

hdelongueville:
- [x] Close #150 
- [x] Add new parameters plot_separate_unc and plot_combined_unc to flux_timeseries
- [x] Update notebook
- [x] Add test

### DIFF
--- a/fluxy/plots/flux_timeseries.py
+++ b/fluxy/plots/flux_timeseries.py
@@ -127,9 +127,7 @@ def prepare_data_to_plot(
     model_labels: dict[str, str],
     model_colors: dict[str, list],
     plot_separate: bool | list[bool] = True,
-    plot_separate_unc: bool | None = None,
     plot_combined: bool | list[bool] = False,
-    plot_combined_unc: bool | None = None,
     resample: str | list[str] | None = None,
     rolling_mean: bool | list[bool] = False,
     resample_uncert_correlation: bool = False,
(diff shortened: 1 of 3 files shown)


fluxy/plots/flux_timeseries.py
```diff
@@ -281,20 +298,24 @@ def add_posterior_plot(
...
     )
-    ax.fill_between(
-        ds_region.time,
-        ds_region.posterior_lower,
-        ds_region.posterior_upper,
-        alpha=0.2,
-        color=ds_region.attrs["model_color"],
-    )
+
+    if plot_unc:
```
melodb: Function `add_prior_plot` has variable `add_prior_unc` as input. To keep the code consistent, I think the same should be done here. That is `plot_unc` should be renamed `add_post_unc` and be an input to function `add_posterior_plot`.

fluxy/plots/flux_timeseries.py
```diff
@@ -255,12 +270,14 @@ def prepare_data_to_plot(
...
             model_color = map_model_colors[key_mc][nb % len(map_model_colors[key_mc])]
             color_usage[key_mc] = color_usage[key_mc] + 1
+            plot_unc = plot_separate_unc
 
         if ("_resample" in m) and plot_resample_and_original:
             include_label += " (resampled)"
 
         ds_to_plot[m].attrs["model_label"] = include_label
         ds_to_plot[m].attrs["model_color"] = model_color
+        ds_to_plot[m].attrs["model_unc_to_plot"] = plot_unc
```
melodb: I would not add to the attributes because plotting options (plot_separate/combined/resample/etc are not added to the attributes). So I would not make an exception here.

fluxy/plots/flux_timeseries.py
melodb, line 775: To avoid adding the flag in the attributes, you can define `add_post_unc` in a similar way as `highlighted_post`. For example:

```
add_post_unc = ((m=="combined") & plot_combined_unc) |  ((m!="combined") & plot_separate_unc)
posterior_df = add_posterior_plot(ax, ds_region, highlighted_post, add_post_unc)
```

melodb (commented):
Thanks for implementing this! I did a suggestion of an alternative integration into the code. It does not really have an advantage in comparison to what you have implemented, but I think it's more consistent with the current structure of the code.

fluxy/plots/flux_timeseries.py
hdelongueville: Thanks for your review! I also thought about this at one point but in the end decided to go with the logic "adding _unc" to the main variables (i.e. `add_prior` --> `add_prior_unc`, `plot_combined` --> `plot_combined_unc`, `plot_separate` --> `plot_separate_unc`) and keep those parameters "as far as possible". 
However, as you suggested, it kinda makes sense to use `add_post_unc` for the function `add_posterior_plot`. I'll make the changes.

melodb (approved):
It looks good to me! I did not test it because I'm working on other branch but I guess you did it. Ready to merge from my side.

PR #234, opened 2025-11-06
Bugfix percentile dimension order

melodb:
While the cdl template specifies (index, percentile) as the dimensions of variables `percentile_*`, fluxy was assuming the opposite order. Moreover, when converting from the old to the new format, the dimensions order was wrongly set (in agreement with fluxy plotting, but not with the template).

The dimensions order is now exchanged in `edit_vars_and_attributes` and the plotting function `plot_timeseries` was updated accordingly.

The only problem is that I'm having a thread.lock error when slicing the data. Therefore, I now load the data immediately after reading, which might not be ideal memory-wise. @lionel42 is there a better way?

### DIFF
--- a/fluxy/io.py
+++ b/fluxy/io.py
@@ -807,6 +807,7 @@ def edit_vars_and_attributes(
                 attrs=ds[var_to_change].attrs,
             )
 
+    # Fix concentration/eddy flux datasets
     elif file_type in (DataTypes.CONCENTRATION, DataTypes.EDDY_FLUX):
         # Ensure integer dtype
         ds["number_of_identifier"] = ds["number_of_identifier"].astype(int)
@@ -825,6 +826,9 @@ def edit_vars_and_attributes(
                 .stack({"index": ["number_of_identifier", "time"]})
                 .reset_index("index")
             )
+
+        # Reorder dimensions
+        ds = ds.transpose('index', 'percentile', ...)
 
         if "assimilation_flag" not in ds:
             # Add assimilation_flag if not present
@@ -901,6 +905,7 @@ def edit_vars_and_attributes(
                 ds["assimilation_flag"][mask] = 0
                 logger.info(f"Masking out nan values in {model}, as a quick fix for a bug in InTEM concentration files.")
 
+    # Fix eddy flux dataset
     if file_type == DataTypes.EDDY_FLUX:
         # check some eddy flux variables
         if "ecflux_observed" not in ds:
--- a/fluxy/io.py
+++ b/fluxy/io.py
@@ -361,6 +361,9 @@ def read_model_output(
             logger.info(f"'species' attribute in dataset {m} ({current_species}) differs from species {species}. It is overwritten.")
             ds_all[m].attrs["species"] = species
 
+        # Load data to avoid thread.lock error
+        ds_all[m] = ds_all[m].load()
+
     return ds_all
 
 
@@ -828,7 +831,9 @@ def edit_vars_and_attributes(
             )
 
         # Reorder dimensions
-        ds = ds.transpose('index', 'percentile', ...)
+        for var in ds.data_vars:
+            if {'index', 'percentile'}.issubset(ds[var].dims):
+                ds[var] = ds[var].transpose('index', 'percentile', ...)
 
         if "assimilation_flag" not in ds:
             # Add assimilation_flag if not present
--- a/fluxy/io.py
+++ b/fluxy/io.py
@@ -361,9 +361,6 @@ def read_model_output(
             logger.info(f"'species' attribute in dataset {m} ({current_species}) differs from species {species}. It is overwritten.")
             ds_all[m].attrs["species"] = species
 
-        # Load data to avoid thread.lock error
-        ds_all[m] = ds_all[m].load()
-
     return ds_all
 
 
@@ -829,11 +826,6 @@ def edit_vars_and_attributes(
                 .stack({"index": ["number_of_identifier", "time"]})
                 .reset_index("index")
             )
-
-        # Reorder dimensions
-        for var in ds.data_vars:
-            if {'index', 'percentile'}.issubset(ds[var].dims):
-                ds[var] = ds[var].transpose('index', 'percentile', ...)
 
         if "assimilation_flag" not in ds:
             # Add assimilation_flag if not present
--- a/fluxy/plots/mf_timeseries.py
+++ b/fluxy/plots/mf_timeseries.py
@@ -266,8 +266,8 @@ def plot_timeseries(
                 # Define uncertainty band
                 flag_fill_between = False
                 if unc_var.split("_")[0] == "percentile":
-                    y1 = ds_plot[unc_var][0, :].values
-                    y2 = ds_plot[unc_var][1, :].values
+                    y1 = ds_plot[unc_var][:,0].values
+                    y2 = ds_plot[unc_var][:,1].values
                     flag_fill_between = True
                 elif unc_var.split("_")[-1] in ["prior", "posterior"]:
                     y1 = ds_plot[var].values - ds_plot[unc_var].values
--- a/fluxy/plots/mf_timeseries.py
+++ b/fluxy/plots/mf_timeseries.py
@@ -266,8 +266,8 @@ def plot_timeseries(
                 # Define uncertainty band
                 flag_fill_between = False
                 if unc_var.split("_")[0] == "percentile":
-                    y1 = ds_plot[unc_var][:,0].values
-                    y2 = ds_plot[unc_var][:,1].values
+                    y1 = ds_plot[unc_var].isel(percentile=0).values
+                    y2 = ds_plot[unc_var].isel(percentile=0).values
                     flag_fill_between = True
                 elif unc_var.split("_")[-1] in ["prior", "posterior"]:
                     y1 = ds_plot[var].values - ds_plot[unc_var].values
--- a/fluxy/plots/mf_timeseries.py
+++ b/fluxy/plots/mf_timeseries.py
@@ -267,7 +267,7 @@ def plot_timeseries(
                 flag_fill_between = False
                 if unc_var.split("_")[0] == "percentile":
                     y1 = ds_plot[unc_var].isel(percentile=0).values
-                    y2 = ds_plot[unc_var].isel(percentile=0).values
+                    y2 = ds_plot[unc_var].isel(percentile=1).values
                     flag_fill_between = True
                 elif unc_var.split("_")[-1] in ["prior", "posterior"]:
                     y1 = ds_plot[var].values - ds_plot[unc_var].values
--- a/fluxy/operators/select.py
+++ b/fluxy/operators/select.py
@@ -180,6 +180,7 @@ def slice_mf(
 
         # Slice data according to site
         if site is not None:
+            ds_all[m] = ds_all[m].load() # Needed to avoid thread.lock error
             try:
                 ds_all[m] = slice_site(ds_all[m], site)
             except ValueError as e:
--- a/fluxy/operators/select.py
+++ b/fluxy/operators/select.py
@@ -180,7 +180,6 @@ def slice_mf(
 
         # Slice data according to site
         if site is not None:
-            ds_all[m] = ds_all[m].load() # Needed to avoid thread.lock error
             try:
                 ds_all[m] = slice_site(ds_all[m], site)
             except ValueError as e:


fluxy/io.py
```diff
@@ -826,6 +830,11 @@ def edit_vars_and_attributes(
                 .reset_index("index")
             )
 
+        # Reorder dimensions
+        for var in ds.data_vars:
+            if {'index', 'percentile'}.issubset(ds[var].dims):
```
ayanu: As I am not familiar with the syntax here, I just want to make sure that this is not called every time 'index' and 'percentile' are in the dimvars, but only when they are in the given order.

ayanu (commented):
Looking good to me. 
Lionel left for vacation yesterday. Maybe ask someone else about the threat.lock error. Brendan?

fluxy/io.py
melodb: It is called every time index and percentile are in dimvars. But it does not swap the order, it simply ensures that the order is index followed by percentile. So if the order is already correct, it does not do anything.

melodb:
@brendan-m-murphy, I had to exchange the order of dimensions "index" and "percentile" when reading wrongly formatted input files. This was leading to a threat.lock error. I managed to fix it by loading the data after reading (`ds_all[m] = ds_all[m].load()` in io.py). Is there a better way to avoid this error?

fluxy/plots/mf_timeseries.py
```diff
@@ -266,8 +266,8 @@ def plot_timeseries(
                 # Define uncertainty band
                 flag_fill_between = False
                 if unc_var.split("_")[0] == "percentile":
-                    y1 = ds_plot[unc_var][0, :].values
-                    y2 = ds_plot[unc_var][1, :].values
+                    y1 = ds_plot[unc_var][:,0].values
```
brendan-m-murphy: Why not use .isel here? Then the order of the dimensions doesn't matter.

For instance,
```
y1 = ds_plot[unc_var].isel(percentile=0).values
```

I would avoid using numpy style indexing with xarray as much as possible, since it leads to these kind of problems and also means you need to remember what dimension corresponds to what axis. (Of course, fixing our outputs would also avoid this problem...)

brendan-m-murphy:
> @brendan-m-murphy, I had to exchange the order of dimensions "index" and "percentile" when reading wrongly formatted input files. This was leading to a threat.lock error. I managed to fix it by loading the data after reading (`ds_all[m] = ds_all[m].load()` in io.py). Is there a better way to avoid this error?

There's probably no harm in loading the data at this point, unless there is a lot more data than is needed.

We've been having trouble with h5netcdf/h5py on OpenGHG and OpenGHG Inversions. If you've updated your environment recently, that might cause I/O issues. (We don't really know what the problem is, we've just pinned these packages to older versions for now. The problem only seems to be with pip environments.)

brendan-m-murphy:
Also, perhaps I can put a PR in to convert the index dimension into a form that has coordinates for platform and time. That should make it much easier to use xarray to select e.g. sites. (Xarray supports stacked dimensions, so index can be a stack of time and platform.)

melodb:
> > @brendan-m-murphy, I had to exchange the order of dimensions "index" and "percentile" when reading wrongly formatted input files. This was leading to a threat.lock error. I managed to fix it by loading the data after reading (`ds_all[m] = ds_all[m].load()` in io.py). Is there a better way to avoid this error?
> 
> There's probably no harm in loading the data at this point, unless there is a lot more data than is needed.
> 
> We've been having trouble with h5netcdf/h5py on OpenGHG and OpenGHG Inversions. If you've updated your environment recently, that might cause I/O issues. (We don't really know what the problem is, we've just pinned these packages to older versions for now. The problem only seems to be with pip environments.)

Thanks, Brendan! Then I think we can keep it like this for now.

melodb:
@ayanu and @brendan-m-murphy, can any of you accept the PR so that it can be merged?

fluxy/io.py
```diff
@@ -361,6 +361,9 @@ def read_model_output(
             logger.info(f"'species' attribute in dataset {m} ({current_species}) differs from species {species}. It is overwritten.")
             ds_all[m].attrs["species"] = species
 
+        # Load data to avoid thread.lock error
+        ds_all[m] = ds_all[m].load()
```
brendan-m-murphy: If you have more details from this error (e.g. traceback or error message, or test/situation that caused it) that might be useful for debugging.

brendan-m-murphy (commented):
I think using xarray's selection feature instead of direct numpy-style indexing would be a good idea, since this automatically selects the correct axis based on the dimension name.

fluxy/plots/mf_timeseries.py
melodb: With your suggestion, the problem is fixed and there is no need to reorder the dimensions or load the data. So I will simply apply your suggestion and revert the remaining changes. @alexdanjou has restructured mf plotting quite a bit in his PR (https://github.com/openghg/fluxy/pull/216, under review). I think he removed the use of indexes as well.

brendan-m-murphy: Okay sounds good.

fluxy/io.py
melodb: If I exchange the dimensions order and do not load the data, I get this error when reading ELRIS or RHIME concentration files (PARISNID2026 outputs). ELRIS concentration files are in the new format, however, dimensions percentile and index are in the wrong order (fixed with transpose); RHIME is in the old format, it gets converted to the new format in fluxy but the current fluxy implementation also yields percentile and index in the wrong order (fixed with transpose).

The error traces back to:
Call to `slice_mf` in notebook > Call to `slice_site` in `slice_mf` > Site masking `ds = ds.where(mask, drop=True)` in `slice_site`

This is the error I get: `TypeError: cannot pickle '_thread.lock' object`

But this won't happen anymore with your suggestions because transpose won't be needed.

-------------------------------------
This is the full error message just in case you are curious:

TypeError                                 Traceback (most recent call last)
Cell In[8], [line 24](vscode-notebook-cell:?execution_count=8&line=24)
     20 ###################################
     22 ds_all_mf = read_model_output(data_dir,'concentration',species,models,config_data,period=period,read_standard_run=read_standard_run)
---> [24](vscode-notebook-cell:?execution_count=8&line=24) ds_all_mf_sliced = slice_mf(ds_all_mf.copy(),start_date,end_date,site,baseline_site=baseline_site,
     25                               baseline_filename=baseline_filename,data_dir=data_dir,
     26                               mf_units_print=mf_units_print,intake_height=intake_height)
     28 model_colors = set_model_colors(models)
     29 model_labels = set_model_labels(models,config_data,get_labels_from_file)

File /project/bmda/PARIS_jupyter/fluxy/fluxy/operators/select.py:184, in slice_mf(ds_all, start_date, end_date, site, baseline_site, baseline_filename, data_dir, mf_units_print, keep_unassimilated, intake_height)
    182 if site is not None:
    183     try:
--> [184](https://vscode-remote+ssh-002dremote-002bddm06676.vscode-resource.vscode-cdn.net/project/bmda/PARIS_jupyter/fluxy/fluxy/operators/select.py:184)         ds_all[m] = slice_site(ds_all[m], site)
    185     except ValueError as e:
    186         logger.warning(f"Error slicing site {site} from {m}: {e}")

File /project/bmda/PARIS_jupyter/fluxy/fluxy/operators/select.py:250, in slice_site(ds, site)
    247     raise ValueError(f"Site {site} not found in dataset.")
    249 mask = ds["number_of_identifier"] == site_index
--> [250](https://vscode-remote+ssh-002dremote-002bddm06676.vscode-resource.vscode-cdn.net/project/bmda/PARIS_jupyter/fluxy/fluxy/operators/select.py:250) ds = ds.where(mask, drop=True)
    252 return ds

File /bigdat/bmda/py311/lib64/python3.11/site-packages/xarray/core/common.py:1158, in DataWithCoords.where(self, cond, other, drop)
   1153 if not isinstance(cond, (Dataset, DataArray)):
   1154     raise TypeError(
   1155         f"cond argument is {cond!r} but must be a {Dataset!r} or {DataArray!r} (or a callable than returns one)."
   1156     )
-> [1158](https://vscode-remote+ssh-002dremote-002bddm06676.vscode-resource.vscode-cdn.net/bigdat/bmda/py311/lib64/python3.11/site-packages/xarray/core/common.py:1158) self, cond = align(self, cond)
   1160 def _dataarray_indexer(dim: Hashable) -> DataArray:
   1161     return cond.any(dim=(d for d in cond.dims if d != dim))

File /bigdat/bmda/py311/lib64/python3.11/site-packages/xarray/core/alignment.py:888, in align(join, copy, indexes, exclude, fill_value, *objects)
    692 """
    693 Given any number of Dataset and/or DataArray objects, returns new
    694 objects with aligned indexes and dimension sizes.
   (...)
    878 
    879 """
    880 aligner = Aligner(
    881     objects,
    882     join=join,
   (...)
    886     fill_value=fill_value,
    887 )
--> [888](https://vscode-remote+ssh-002dremote-002bddm06676.vscode-resource.vscode-cdn.net/bigdat/bmda/py311/lib64/python3.11/site-packages/xarray/core/alignment.py:888) aligner.align()
    889 return aligner.results

File /bigdat/bmda/py311/lib64/python3.11/site-packages/xarray/core/alignment.py:582, in Aligner.align(self)
    580     self.results = self.objects
    581 else:
--> [582](https://vscode-remote+ssh-002dremote-002bddm06676.vscode-resource.vscode-cdn.net/bigdat/bmda/py311/lib64/python3.11/site-packages/xarray/core/alignment.py:582)     self.reindex_all()

File /bigdat/bmda/py311/lib64/python3.11/site-packages/xarray/core/alignment.py:557, in Aligner.reindex_all(self)
    556 def reindex_all(self) -> None:
--> [557](https://vscode-remote+ssh-002dremote-002bddm06676.vscode-resource.vscode-cdn.net/bigdat/bmda/py311/lib64/python3.11/site-packages/xarray/core/alignment.py:557)     self.results = tuple(
    558         self._reindex_one(obj, matching_indexes)
    559         for obj, matching_indexes in zip(
    560             self.objects, self.objects_matching_indexes
    561         )
    562     )

File /bigdat/bmda/py311/lib64/python3.11/site-packages/xarray/core/alignment.py:[558](https://vscode-remote+ssh-002dremote-002bddm06676.vscode-resource.vscode-cdn.net/bigdat/bmda/py311/lib64/python3.11/site-packages/xarray/core/alignment.py:558), in <genexpr>(.0)
    556 def reindex_all(self) -> None:
    557     self.results = tuple(
--> 558         self._reindex_one(obj, matching_indexes)
    559         for obj, matching_indexes in zip(
    560             self.objects, self.objects_matching_indexes
    561         )
    562     )

File /bigdat/bmda/py311/lib64/python3.11/site-packages/xarray/core/alignment.py:546, in Aligner._reindex_one(self, obj, matching_indexes)
    543 new_indexes, new_variables = self._get_indexes_and_vars(obj, matching_indexes)
    544 dim_pos_indexers = self._get_dim_pos_indexers(matching_indexes)
--> [546](https://vscode-remote+ssh-002dremote-002bddm06676.vscode-resource.vscode-cdn.net/bigdat/bmda/py311/lib64/python3.11/site-packages/xarray/core/alignment.py:546) return obj._reindex_callback(
    547     self,
    548     dim_pos_indexers,
    549     new_variables,
    550     new_indexes,
    551     self.fill_value,
    552     self.exclude_dims,
    553     self.exclude_vars,
    554 )

File /bigdat/bmda/py311/lib64/python3.11/site-packages/xarray/core/dataset.py:3420, in Dataset._reindex_callback(self, aligner, dim_pos_indexers, variables, indexes, fill_value, exclude_dims, exclude_vars)
   3418         reindexed = self._overwrite_indexes(new_indexes, new_variables)
   3419     else:
-> [3420](https://vscode-remote+ssh-002dremote-002bddm06676.vscode-resource.vscode-cdn.net/bigdat/bmda/py311/lib64/python3.11/site-packages/xarray/core/dataset.py:3420)         reindexed = self.copy(deep=aligner.copy)
   3421 else:
   3422     to_reindex = {
   3423         k: v
   3424         for k, v in self.variables.items()
   3425         if k not in variables and k not in exclude_vars
   3426     }

File /bigdat/bmda/py311/lib64/python3.11/site-packages/xarray/core/dataset.py:1336, in Dataset.copy(self, deep, data)
   1239 def copy(self, deep: bool = False, data: DataVars | None = None) -> Self:
   1240     """Returns a copy of this dataset.
   1241 
   1242     If `deep=True`, a deep copy is made of each of the component variables.
   (...)
   1334     pandas.DataFrame.copy
   1335     """
-> [1336](https://vscode-remote+ssh-002dremote-002bddm06676.vscode-resource.vscode-cdn.net/bigdat/bmda/py311/lib64/python3.11/site-packages/xarray/core/dataset.py:1336)     return self._copy(deep=deep, data=data)

File /bigdat/bmda/py311/lib64/python3.11/site-packages/xarray/core/dataset.py:1372, in Dataset._copy(self, deep, data, memo)
   1370         variables[k] = index_vars[k]
   1371     else:
-> [1372](https://vscode-remote+ssh-002dremote-002bddm06676.vscode-resource.vscode-cdn.net/bigdat/bmda/py311/lib64/python3.11/site-packages/xarray/core/dataset.py:1372)         variables[k] = v._copy(deep=deep, data=data.get(k), memo=memo)
   1374 attrs = copy.deepcopy(self._attrs, memo) if deep else copy.copy(self._attrs)
   1375 encoding = (
   1376     copy.deepcopy(self._encoding, memo) if deep else copy.copy(self._encoding)
   1377 )

File /bigdat/bmda/py311/lib64/python3.11/site-packages/xarray/core/variable.py:932, in Variable._copy(self, deep, data, memo)
    929         ndata = indexing.MemoryCachedArray(data_old.array)  # type: ignore[assignment]
    931     if deep:
--> [932](https://vscode-remote+ssh-002dremote-002bddm06676.vscode-resource.vscode-cdn.net/bigdat/bmda/py311/lib64/python3.11/site-packages/xarray/core/variable.py:932)         ndata = copy.deepcopy(ndata, memo)
    934 else:
    935     ndata = as_compatible_data(data)

File /usr/lib64/python3.11/copy.py:172, in deepcopy(x, memo, _nil)
    170                 y = x
    171             else:
--> [172](https://vscode-remote+ssh-002dremote-002bddm06676.vscode-resource.vscode-cdn.net/usr/lib64/python3.11/copy.py:172)                 y = _reconstruct(x, memo, *rv)
    174 # If is its own copy, don't memoize.
    175 if y is not x:

File /usr/lib64/python3.11/copy.py:271, in _reconstruct(x, memo, func, args, state, listiter, dictiter, deepcopy)
    269 if state is not None:
    270     if deep:
--> [271](https://vscode-remote+ssh-002dremote-002bddm06676.vscode-resource.vscode-cdn.net/usr/lib64/python3.11/copy.py:271)         state = deepcopy(state, memo)
    272     if hasattr(y, '__setstate__'):
    273         y.__setstate__(state)

File /usr/lib64/python3.11/copy.py:146, in deepcopy(x, memo, _nil)
    144 copier = _deepcopy_dispatch.get(cls)
    145 if copier is not None:
--> [146](https://vscode-remote+ssh-002dremote-002bddm06676.vscode-resource.vscode-cdn.net/usr/lib64/python3.11/copy.py:146)     y = copier(x, memo)
    147 else:
    148     if issubclass(cls, type):

File /usr/lib64/python3.11/copy.py:211, in _deepcopy_tuple(x, memo, deepcopy)
    210 def _deepcopy_tuple(x, memo, deepcopy=deepcopy):
--> [211](https://vscode-remote+ssh-002dremote-002bddm06676.vscode-resource.vscode-cdn.net/usr/lib64/python3.11/copy.py:211)     y = [deepcopy(a, memo) for a in x]
    212     # We're not going to put the tuple in the memo, but it's still important we
    213     # check for it, in case the tuple contains recursive mutable structures.
    214     try:

File /usr/lib64/python3.11/copy.py:211, in <listcomp>(.0)
    210 def _deepcopy_tuple(x, memo, deepcopy=deepcopy):
--> [211](https://vscode-remote+ssh-002dremote-002bddm06676.vscode-resource.vscode-cdn.net/usr/lib64/python3.11/copy.py:211)     y = [deepcopy(a, memo) for a in x]
    212     # We're not going to put the tuple in the memo, but it's still important we
    213     # check for it, in case the tuple contains recursive mutable structures.
    214     try:

File /usr/lib64/python3.11/copy.py:146, in deepcopy(x, memo, _nil)
    144 copier = _deepcopy_dispatch.get(cls)
    145 if copier is not None:
--> [146](https://vscode-remote+ssh-002dremote-002bddm06676.vscode-resource.vscode-cdn.net/usr/lib64/python3.11/copy.py:146)     y = copier(x, memo)
    147 else:
    148     if issubclass(cls, type):

File /usr/lib64/python3.11/copy.py:231, in _deepcopy_dict(x, memo, deepcopy)
    229 memo[id(x)] = y
    230 for key, value in x.items():
--> [231](https://vscode-remote+ssh-002dremote-002bddm06676.vscode-resource.vscode-cdn.net/usr/lib64/python3.11/copy.py:231)     y[deepcopy(key, memo)] = deepcopy(value, memo)
    232 return y

File /usr/lib64/python3.11/copy.py:172, in deepcopy(x, memo, _nil)
    170                 y = x
    171             else:
--> [172](https://vscode-remote+ssh-002dremote-002bddm06676.vscode-resource.vscode-cdn.net/usr/lib64/python3.11/copy.py:172)                 y = _reconstruct(x, memo, *rv)
    174 # If is its own copy, don't memoize.
    175 if y is not x:

File /usr/lib64/python3.11/copy.py:271, in _reconstruct(x, memo, func, args, state, listiter, dictiter, deepcopy)
    269 if state is not None:
    270     if deep:
--> [271](https://vscode-remote+ssh-002dremote-002bddm06676.vscode-resource.vscode-cdn.net/usr/lib64/python3.11/copy.py:271)         state = deepcopy(state, memo)
    272     if hasattr(y, '__setstate__'):
    273         y.__setstate__(state)

File /usr/lib64/python3.11/copy.py:146, in deepcopy(x, memo, _nil)
    144 copier = _deepcopy_dispatch.get(cls)
    145 if copier is not None:
--> [146](https://vscode-remote+ssh-002dremote-002bddm06676.vscode-resource.vscode-cdn.net/usr/lib64/python3.11/copy.py:146)     y = copier(x, memo)
    147 else:
    148     if issubclass(cls, type):

File /usr/lib64/python3.11/copy.py:211, in _deepcopy_tuple(x, memo, deepcopy)
    210 def _deepcopy_tuple(x, memo, deepcopy=deepcopy):
--> [211](https://vscode-remote+ssh-002dremote-002bddm06676.vscode-resource.vscode-cdn.net/usr/lib64/python3.11/copy.py:211)     y = [deepcopy(a, memo) for a in x]
    212     # We're not going to put the tuple in the memo, but it's still important we
    213     # check for it, in case the tuple contains recursive mutable structures.
    214     try:

File /usr/lib64/python3.11/copy.py:211, in <listcomp>(.0)
    210 def _deepcopy_tuple(x, memo, deepcopy=deepcopy):
--> [211](https://vscode-remote+ssh-002dremote-002bddm06676.vscode-resource.vscode-cdn.net/usr/lib64/python3.11/copy.py:211)     y = [deepcopy(a, memo) for a in x]
    212     # We're not going to put the tuple in the memo, but it's still important we
    213     # check for it, in case the tuple contains recursive mutable structures.
    214     try:

File /usr/lib64/python3.11/copy.py:146, in deepcopy(x, memo, _nil)
    144 copier = _deepcopy_dispatch.get(cls)
    145 if copier is not None:
--> [146](https://vscode-remote+ssh-002dremote-002bddm06676.vscode-resource.vscode-cdn.net/usr/lib64/python3.11/copy.py:146)     y = copier(x, memo)
    147 else:
    148     if issubclass(cls, type):

File /usr/lib64/python3.11/copy.py:231, in _deepcopy_dict(x, memo, deepcopy)
    229 memo[id(x)] = y
    230 for key, value in x.items():
--> [231](https://vscode-remote+ssh-002dremote-002bddm06676.vscode-resource.vscode-cdn.net/usr/lib64/python3.11/copy.py:231)     y[deepcopy(key, memo)] = deepcopy(value, memo)
    232 return y

File /usr/lib64/python3.11/copy.py:172, in deepcopy(x, memo, _nil)
    170                 y = x
    171             else:
--> [172](https://vscode-remote+ssh-002dremote-002bddm06676.vscode-resource.vscode-cdn.net/usr/lib64/python3.11/copy.py:172)                 y = _reconstruct(x, memo, *rv)
    174 # If is its own copy, don't memoize.
    175 if y is not x:

File /usr/lib64/python3.11/copy.py:271, in _reconstruct(x, memo, func, args, state, listiter, dictiter, deepcopy)
    269 if state is not None:
    270     if deep:
--> [271](https://vscode-remote+ssh-002dremote-002bddm06676.vscode-resource.vscode-cdn.net/usr/lib64/python3.11/copy.py:271)         state = deepcopy(state, memo)
    272     if hasattr(y, '__setstate__'):
    273         y.__setstate__(state)

File /usr/lib64/python3.11/copy.py:146, in deepcopy(x, memo, _nil)
    144 copier = _deepcopy_dispatch.get(cls)
    145 if copier is not None:
--> [146](https://vscode-remote+ssh-002dremote-002bddm06676.vscode-resource.vscode-cdn.net/usr/lib64/python3.11/copy.py:146)     y = copier(x, memo)
    147 else:
    148     if issubclass(cls, type):

File /usr/lib64/python3.11/copy.py:211, in _deepcopy_tuple(x, memo, deepcopy)
    210 def _deepcopy_tuple(x, memo, deepcopy=deepcopy):
--> [211](https://vscode-remote+ssh-002dremote-002bddm06676.vscode-resource.vscode-cdn.net/usr/lib64/python3.11/copy.py:211)     y = [deepcopy(a, memo) for a in x]
    212     # We're not going to put the tuple in the memo, but it's still important we
    213     # check for it, in case the tuple contains recursive mutable structures.
    214     try:

File /usr/lib64/python3.11/copy.py:211, in <listcomp>(.0)
    210 def _deepcopy_tuple(x, memo, deepcopy=deepcopy):
--> [211](https://vscode-remote+ssh-002dremote-002bddm06676.vscode-resource.vscode-cdn.net/usr/lib64/python3.11/copy.py:211)     y = [deepcopy(a, memo) for a in x]
    212     # We're not going to put the tuple in the memo, but it's still important we
    213     # check for it, in case the tuple contains recursive mutable structures.
    214     try:

File /usr/lib64/python3.11/copy.py:146, in deepcopy(x, memo, _nil)
    144 copier = _deepcopy_dispatch.get(cls)
    145 if copier is not None:
--> [146](https://vscode-remote+ssh-002dremote-002bddm06676.vscode-resource.vscode-cdn.net/usr/lib64/python3.11/copy.py:146)     y = copier(x, memo)
    147 else:
    148     if issubclass(cls, type):

File /usr/lib64/python3.11/copy.py:231, in _deepcopy_dict(x, memo, deepcopy)
    229 memo[id(x)] = y
    230 for key, value in x.items():
--> [231](https://vscode-remote+ssh-002dremote-002bddm06676.vscode-resource.vscode-cdn.net/usr/lib64/python3.11/copy.py:231)     y[deepcopy(key, memo)] = deepcopy(value, memo)
    232 return y

File /usr/lib64/python3.11/copy.py:161, in deepcopy(x, memo, _nil)
    159 reductor = getattr(x, "__reduce_ex__", None)
    160 if reductor is not None:
--> [161](https://vscode-remote+ssh-002dremote-002bddm06676.vscode-resource.vscode-cdn.net/usr/lib64/python3.11/copy.py:161)     rv = reductor(4)
    162 else:
    163     reductor = getattr(x, "__reduce__", None)

TypeError: cannot pickle '_thread.lock' object

brendan-m-murphy: Hmm that's a pretty weird error. 

I think the `where` could be avoided if you convert `time` to a coordinate. I have some code to do this (the part about making `index` a stacked coordinate that I mentioned above). I guess `.where` causes problems with data opened with dask that isn't "computed" (similar to loaded); I'm not sure if it is the same problem.

If `time` was a coordinate, you could just do
```
ds_all[m] = ds_all[m].sel(time=slice(start_date, end_date))
```
to slice the times, although you might still need `.where` to select the assimilation flag.

fluxy/plots/mf_timeseries.py
```diff
@@ -266,8 +266,8 @@ def plot_timeseries(
                 # Define uncertainty band
                 flag_fill_between = False
                 if unc_var.split("_")[0] == "percentile":
-                    y1 = ds_plot[unc_var][0, :].values
-                    y2 = ds_plot[unc_var][1, :].values
+                    y1 = ds_plot[unc_var].isel(percentile=0).values
+                    y2 = ds_plot[unc_var].isel(percentile=0).values
```
brendan-m-murphy:
```suggestion
                    y2 = ds_plot[unc_var].isel(percentile=1).values
```
This should select the second value I think.

brendan-m-murphy (changes requested):
Looks like a typo? Looks good besides that.

fluxy/plots/mf_timeseries.py
melodb: Yes, sorry. Fixed now.

brendan-m-murphy (approved):
Looks good!

fluxy/io.py
melodb: After reading the data, we set `time` and `number_of_identifier` to coordinates in `edit_vars_and_attributes`:
`ds = ds.assign_coords({var: ds[var] for var in ["number_of_identifier", "time", "platform"]})`

But `time` is not set to a dimension. It is indexed by `index`. So we currently use the following for time selection:
```
mask = (ds_all[m]["time"] >= start_date) & (ds_all[m]["time"] <= end_date)
mask &= ds_all[m]["assimilation_flag"] == 1
ds_all[m] = ds_all[m].where(mask, drop=True)
```

And the following for site selection:
```
mask = ds["number_of_identifier"] == site_index
ds = ds.where(mask, drop=True)
```

When the old format was used, `time` and` platform` were dimensions. I don't remember anymore why we did not restructured from the new format (index-based) to `time` and `platform` in fluxy. I have the impression there was an argument against. Currently, we also select the data to plot based on the inlet height (for example, to distinguish TAC-100 from TAC-185):
```
mask = ds["intake_height"] == float(intake_height)
ds = ds.where(mask, drop=True)
```

Maybe this was easier to handle with the index-based data?

brendan-m-murphy: I used a function like this:
```
def restore_mindex(
    data: xr.Dataset,
    index_var: str = "index",
    platform_var: str = "platform",
    indicator_var: str = "number_of_identifier",
) -> xr.Dataset:
    platform_arr = data[platform_var].values[data[indicator_var].values.astype("int")]
    mindex = pd.MultiIndex.from_arrays([platform_arr, data.time.values])
    return data.assign_coords(xr.Coordinates.from_pandas_multiindex(mindex, index_var))
```

to make `index` a coordinate that is a stack of `time` and `platform` coordinates (and also creates dimensions for `time` and `platform`). I'm not 100% sure that this will let you select by `time` and `platform`... you might need to do `.unstack("index")` first (which might introduce NaNs if different platforms have different time values). We use something similar in OpenGHG Inversions to work with our old output format, and I found have the time and site (or platform) coordinates useful.

Also the `platform_arr` replaces the indices with the labels, but you could just use `data[indicator_var].values` instead if you want the indices.

(If you already have `time` and `platform`, then `ds.stack(index=("time", "platform")).dropna("index")` will give you an index coordinate like fluxy uses now. You can't save stacked dimensions/coordinates like this to netCDF, so either you need to unstack (and add NaNs), or convert to something like the format in the CDL templates.)

 Since `assimilation_flag` isn't a dimension you'd need to use `.where` in that case anyway. But potentially you could do `ds_all[m].sel(time=slice(start_date, end_date).sel(platform="TAC")`. (If say, `"TAC" == ds_all[m]["platform"][site_index]`)

PR #236, opened 2025-11-10
Add option to plot multiple "combined models" in flux_timeseries

### LINKED ISSUE #231: Add the option to compute multiple combined flux by model tag
Currently, the `plot_combined` option in `plot_country_flux` only creates one combined "Model Mean" across models for which `plot_combined = True`. I'd like to extend this so that we can create _multiple_ combined means based on shared tages (e.g. NAME vs FLEXPART).

I propose adding a new option (yes, another one..), for example:
`plot_combined_per_tag = True`

When enabled, models would be grouped by the tag after the model name, and a combined mean would be computed for each group.

Example
`models = ['InTEM_NAME', 'InTEM_FLEXPART', 'ELRIS_NAME', 'ELRIS_FLEXPART', 'RHIME_NAME', 'RHIME_FLEXPART']`
This would produce two combined means:
`Mean NAME`: `['InTEM_NAME', 'ELRIS_NAME', 'RHIME_NAME']`
`Mean FLEXPART`: `['InTEM_FLEXPART', 'ELRIS_FLEXPART', 'RHIME_FLEXPART']`

This would make it easier to compare aggregated NAME vs FLEXPART results directly on the same plot.

hdelongueville:
This PR extends the flux_timeseries plotting workflow to handle multiple combined model entries instead of assuming only one "combined" dataset.

- [x] close #231 
- [x] add test
- [x] update notebook


hdelongueville:
It currently doesn't work with `resample`.
In the case where, for example:
```
models = ['model1_a', 'model2_a', 'model3_b']
resample = ['year', 'year', None]
plot_combined = True
```
Do we want to combine only the resampled models or all three models?
`combined_models_dict = {'Mean (from resampled data)': ['model1_a_resampled', 'model2_a_resampled']}`
or
`combined_models_dict = {'Mean': ['model1_a_resampled', 'model2_a_resampled', 'model3_b']}`

Moreover, in the case where `plot_resample_and_original = True`, do we want to combine only resampled data or plot 2 distinct means?

I have to say, I am a bit lost with those options as I don't really use them..

melodb:
I have tried the code and it breaks if you try to combine yearly-resampled with seasonal-resampled data:
```
plot_combined = True
resample = ["year","year","season"]
--> ValueError: operands could not be broadcast together with shapes (6,) (24,)
```

If one of the resample options is set to None:
```
plot_combined = True
resample = ["year","year",None]
```
the code automatically combines the original data instead of the resampled data.

If all resample options of the datasets to combine are equal and different from None:
```
plot_combined = True
resample = "year"
```
the code successfully combines the resampled data.

That's basically the role of this if-statement:
https://github.com/openghg/fluxy/blob/13337729b6f6fec77ca5df1c282999beb853f692/fluxy/plots/flux_timeseries.py#L223

So I think we can follow a similar strategy:
- if all resample options of the plots to combine are the same and different from None (="year" or "season"), we combine the resampled data.
- if all resample options of the plots to combine are equal to None, we combine the original data.
- if all resample options are not equal, we combine the original data and print a warning.

Regarding option `plot_resample_and_original`, currently in the code it only concerns the separate plots. So I would say you ignore it when plotting the combined line.

What do you think?

hdelongueville:
That sounds like a good plan, thanks!

hdelongueville:
Normally, I've fixed the resampling while using `plot_combined`. 
The last remaining "incoherence" is that when 
```
plot_combined = True
resample = 'year'
plot_resample_and_original = True
```
only the Mean resampled is plotted. I suggest to open an issue for that and include both the original mean and the resampled mean in another PR.

fluxy/plots/flux_timeseries.py
```diff
@@ -147,6 +148,7 @@ def prepare_data_to_plot(
         plot_combined: If True, the model is included in combined average result to be plotted. List must be of same size as models,
             e.g. [False, True, True].
             If a single boolean is provided, the same flag is assumed for all models.
+        combined_models_dict: dictionnary defining the different combined models to plot. Keys are the name of the combined model.
```
melodb: Maybe replace last sentence by "Keys are the labels of the combined results."
And add "Only used if plot_combined is set to True."

fluxy/plots/flux_timeseries.py
```diff
@@ -162,6 +164,11 @@ def prepare_data_to_plot(
         ds_to_plot : dictionnary of datasets to plot
     """
 
+    if isinstance(plot_combined, bool) and plot_combined:
```
melodb: Can this if be replaced by `if plot_combined == True`?

fluxy/plots/flux_timeseries.py
```diff
@@ -219,32 +226,63 @@ def prepare_data_to_plot(
...
     if any(plot_combined):
-        if all([resamp for comb, resamp in zip(plot_combined, resample) if comb]):
-            ds_to_combine = {
-                m: calc_rolling_mean(ds) if rm else ds
-                for rm, (m, ds) in zip(rolling_mean, ds_resampled.items())
+        if is_plot_combined_single_true:
+            if combined_models_dict is None:
+                combined_models_dict = {"Mean": list(ds_all_region.keys())}
+        else:
+            combined_models_dict = {
```
melodb, line 264: Maybe here you can also check if `combined_models_dict` is None. If it is not None, maybe the following warning can be printed "combined_models_dict will be re-written according to the elements of plot_combined. Label 'Mean' will be used in the plots. To combine the models listed in combined_models_dict, set plot_combined = True."

fluxy/plots/flux_timeseries.py
```diff
@@ -219,32 +226,63 @@ def prepare_data_to_plot(
...
-            ds_combined["combined"].attrs[
-                "model_label"
-            ] = "PARIS mean (from resampled data)"
+
+        combined_resample = [resamp for comb, resamp in zip(plot_combined, resample) if comb]
+        use_resampled = (
+            any(resample) 
+            and all(r is not None for r in combined_resample)
+            and len(set(combined_resample)) == 1
+        )
```
melodb: You can replace these 3 conditions by 2 with the following:
```
use_resampled = (
    len(unique_resample := set(combined_resample)) == 1
    and unique_resample != {None}
)
```

fluxy/plots/flux_timeseries.py
```diff
@@ -219,32 +226,63 @@ def prepare_data_to_plot(
...
-            ds_combined["combined"].attrs["model_label"] = "PARIS mean"
-        ds_to_plot.update(ds_combined)
+
+        for group_label, model_list in combined_models_dict.items():
+                check_missing_models = set(model_list) - set(ds_to_combine.keys())
+                if check_missing_models:
+                    raise ValueError(
+                        f"Models in group '{group_label}' are not available: {check_missing_models}. "
+                        f"Available models: {list(ds_to_combine.keys())}"
+                    )
```
melodb: In my view, this check should be performed at the beginning of the script. For example, you can add an else in the if-statement of L.232-233 and check if all dictionary values exist in `models`.

fluxy/plots/flux_timeseries.py
```diff
@@ -219,32 +226,63 @@ def prepare_data_to_plot(
...
+                combine_mask = [model in model_list for model in ds_to_combine.keys()]
+                ds_combined = combine_dataset(ds_to_combine, combine_mask)
+                ds_combined["combined"].attrs["model_label"] = group_label
+                if any('_resample' in s for s in model_list) and plot_resample_and_original:
+                    ds_combined["combined"].attrs["model_label"] += " (resampled)"
+                new_key = group_label.replace(" ", "_")
+                ds_combined = {f"combined_{new_key}": ds_combined["combined"]} # rename key to include group label
```
melodb: I guess that at this stage `ds_combined["combined"]` still exists. Should it be deleted?

scripts/PARIS_inversion_results.ipynb
```diff
@@ -175,6 +175,7 @@
     "plot_separate = True #If True, includes all model results as separate lines (or insert a list of boolean of the same length as models to specify which models to plot)\n",
     "plot_separate_unc = True #If True, includes all model uncertainty as shaded areas\n",
     "plot_combined = False #If True, combined results, averaged from all models (or insert a list of boolean of the same length as models to specify which models to combine)\n",
+    "                      #To plot multiple combined results, use (e.g.) combined_models_dict = {'NAME Mean': ['InTEM_NAME', 'ELRIS_NAME'], 'FLEXPART Mean': ['InTEM_FLEXPART', 'ELRIS_FLEXPART']} and set plot_combined = True.\n",
```
melodb: To facilitate usage, I would add `combined_models_dict = None` to the list of variables to edit and `combined_models_dict = combined_models_dict` to the function call.

fluxy/plots/flux_timeseries.py
```diff
@@ -219,32 +226,63 @@ def prepare_data_to_plot(
...
+        )
+        if use_resampled:
+                combined_models_dict = {
+                    group_label: [f"{model}_resample" for model in model_list]
+                    for group_label, model_list in combined_models_dict.items()
+                }
+                ds_to_combine = {
+                    m: calc_rolling_mean(ds) if rm else ds
+                    for rm, (m, ds) in zip(rolling_mean, ds_resampled.items())
+                }
```
melodb, line 283: The current implementation does not allow resampling to be evaluated for each combined model group. For example, with the following input parameters:
```
plot_combined = True
combined_models_dict = {'NAME Mean': ['InTEM_NAME', 'ELRIS_NAME','RHIME_NAME'],
                        'FLEXPART Mean': ['ELRIS_FLEXPART','InTEM_FLEXPART']}
resample = ["year","year","year",None,None]
```
we are not able to plot "NAME Mean" as a yearly average and "FLEXPART Mean" as monthly. I guess this is not super important so I'm ok with not fixing this at the moment.

melodb (commented):
Thanks a lot for working on this! I left some small comments. I have tried the notebook and everything is working very well! Maybe we can open an issue to expand the `plot_combined` options in function `plot_flux_map_over_time` as well.

alexdanjou:
Do we need both `plot_combined` and `combined_models_dict`? Maybe we can just add the option to pass a dict to `plot_combined`?

melodb:
> Do we need both `plot_combined` and `combined_models_dict`? Maybe we can just add the option to pass a dict to `plot_combined`?

There was a long discussion on this xD see issue https://github.com/openghg/fluxy/issues/231

alexdanjou:
> > Do we need both `plot_combined` and `combined_models_dict`? Maybe we can just add the option to pass a dict to `plot_combined`?
> 
> There was a long discussion on this xD see issue #231

Whoopsie daisy.. I didn't look at the linked issue, sorry about that. Indeed, it is answering my question quite thoroughly

fluxy/plots/flux_timeseries.py
hdelongueville: Unfortunately, replacing with your suggestion fails when resample is None.

melodb: Sorry, I might have overlooked something.

hdelongueville: Checking if any(resample) is still needed but I find your code cleaner. 
```
use_resampled = (
            any(resample) 
            and len(unique_resample := set(combined_resample)) == 1
            and None not in unique_resample
        )
```
a mix of both ;)

fluxy/plots/flux_timeseries.py
hdelongueville: Good idea!

fluxy/plots/flux_timeseries.py
hdelongueville: Nice suggestion! As it is not urgent, as suggest to fix this later. I opened the related issue #244

fluxy/plots/flux_timeseries.py
melodb: Taking into account that you have added this check in L.236, can it be deleted from here?

fluxy/plots/flux_timeseries.py
melodb: I now understand the problem. When `resample=None`, it is rewritten as a list of `False` in `update_list_params`. While if we set `resample=[None]*len(models)`, it stays as a list of `None`. That's why `unique_resample` can either be equal to `{None}` or to `{False}`. In case you prefer, you can replace the conditions `any(resample) and None not in unique_resample` by `unique_resample not in ({None}, {False})`, which is basically the same, just slightly shorter.

melodb (approved):
All working fine on my end! I reopened two issues and left a reply in case you want to consider them.

fluxy/plots/flux_timeseries.py
hdelongueville: Very good point, thanks! I've changed to follow your suggestion.

fluxy/plots/flux_timeseries.py
hdelongueville: Ups you're right!

hdelongueville:
> All working fine on my end! I reopened two issues and left a reply in case you want to consider them.

Thanks for reviewing it again! I've followed your suggestion and removed the redundant bits. :)

PR #238, opened 2025-11-13
Additional fix for _flux.nc filenames

ayanu:
When adding site information during reading of flux files, concentration file name is accessed based on flux filename. 
Need to treat alternative filename. 

Current solution is a bit hard-coded and not making use of defined DataTypes. 

Was tested with example '_flux.nc' files.


melodb (approved):

melodb:
In the defined DataTypes, we don't distinguish the two type of flux files, so I think the current implementation is ok.

ayanu:
No, we don't but we wouldn't need to hard code the 'concentration' extension either. But let's not dwell on it for now.  Who is going to press the merge button?

PR #240, opened 2025-11-14
Add option to input filepath parameters

melodb:
- It's now possible to input a dictionary with filepath parameters. Expected values are "data_dir", "model_dir", "species_dir", "sub_dir" and "model_name". The keys of the dictionary are the elements of `models`.
- period and suffix=file_pattern are direct inputs to `read_model_output`
- species_print is given by species (input to read_model_output) or from species dictionary in models_info (base_model_name is used as key, we still assume base_model_name to be `split("_")[0]` of the elements of `models`).

**NOTES:**
- functions `make_template` and `fill_template` from Brendan are used.
- application of black formatter: that from L.272-316 no changes were made to the code, it was simply placed inside the if-statement in L.271.

### DIFF
--- a/fluxy/io.py
+++ b/fluxy/io.py
@@ -339,17 +339,28 @@ def get_filename(
         suffix="suffix",
     )
 
+    # Define dictionary with template arguments
+    fill_kwargs = {
+        "data_dir": data_dir,
+        "model_dir": base_model_name,
+        "species_dir": species,
+        "sub_dir": sub_dir,
+        "model_name": model_name,
+    }
+
+    fill_kwargs.update(
+        {
+            **filepath_kwargs,
+            "species_print": species_print,
+            "period": period,
+            "suffix": file_pattern,
+        },
+    )
+
     # Get full path to file
     filepath = fill_template(
         filepath_temp,
-        data_dir=filepath_kwargs.get("data_dir", data_dir),
-        model_dir=filepath_kwargs.get("model_dir", base_model_name),
-        species_dir=filepath_kwargs.get("species_dir", species),
-        sub_dir=filepath_kwargs.get("sub_dir", sub_dir),
-        model_name=filepath_kwargs.get("model_name", model_name),
-        species_print=species_print,
-        period=period,
-        suffix=file_pattern,
+        **fill_kwargs,
     )
 
     return Path(filepath)
--- a/fluxy/io.py
+++ b/fluxy/io.py
@@ -3,6 +3,7 @@ import json
 import logging
 import os
 import re
+import string
 from io import BytesIO
 from pathlib import Path
 from urllib.request import urlopen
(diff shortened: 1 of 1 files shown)


melodb:
Could you check if the current implementation is in agreement with our discussion? It's working fine for me.

Regarding the use of functions `make_template` and `fill_tempate`, honestly I find them a bit of an overkill. We have one single `read_model_output` function and one single file template. So basically those function definitions and the call to those functions can be simply replaced by:

```
    # Define filepath
    data_dir = Path(filepath_kwargs.get("data_dir", data_dir))
    sub_dir = filepath_kwargs.get("sub_dir", sub_dir)
    model_name = filepath_kwargs.get("model_name", model_name)

    filepath = (
        data_dir
        / base_model_name
        / species
        / sub_dir
        / f"{model_name}_{species_print}_{period}{file_pattern}"
    )
```
which I find more straightforward to understand. Let me know your preference.

brendan-m-murphy:
I'll have a closer look, but I guess the advantage of using a template would come from separating the code that gets the arguments for the current setup from the code that makes the file paths and then loads the data. This opens the option to using data stored in a different way. (For instance, I couldn't use the I/O functions for the runs on our server, so I had to do a lot of work to get the data in the format fluxy wants.)

To add this particular feature, you probably don't need to allow custom file path templates. I could make a separate PR that makes the custom file structure possible. Up to you. (Either way, I think passing the arguments by a dictionary is a good idea.)

fluxy/io.py
```diff
@@ -239,20 +325,35 @@ def get_filename(
     ):
         species_print = species_tag
 
-    # Define filepath
-    data_dir = Path(data_dir)
+    # Get file pattern
```
brendan-m-murphy, line 327: To allow custom file structures, the processing above this point could be a separate function. Then the part below could be carried out (possibly with a custom template). (Maybe to make this easier, `fill_template` could ignore kwargs that don't appear in the template.)

melodb:
If the template is useful, I think we can keep it. If something can be improved in the code or if I'm not using it properly, please feel free to make suggestions.

Currently, `species`, `period` and `file_pattern` are input variables to `read_model_output`. Therefore, they are not taken from `filepath_kwargs`. This implies that the filename structure remains mandatory (`{model_name}_{species}_{period}{file_pattern}`). Even though there are some checks in the code, I think this is important to make sure that we are not plotting HFC-134a from one model vs HFC-32 from another.

fluxy/io.py
```diff
@@ -239,20 +324,35 @@ def get_filename(
...
-        data_dir
-        / base_model_name
-        / species
-        / sub_dir
-        / f"{model_name}_{species_print}_{period}{file_pattern}"
+
+    # Hard-coded filename template
+    filepath_temp = make_template(
+        "data_dir",
+        "model_dir",
```
brendan-m-murphy, line 335: Do you want an optional `"sub_data_dir?"` here?

fluxy/io.py
```diff
@@ -239,20 +324,35 @@ def get_filename(
...
+        "species_dir",
+        "sub_dir?",
+        ("model_name", "species_print", "period"),
+        suffix="suffix",
     )
 
-    return filepath
+    # Get full path to file
+    filepath = fill_template(
+        filepath_temp,
```
brendan-m-murphy, line 362: There is a pattern you can use to simplify this slightly: collect the alternative arguments from the `.get` statements in a dictionary, like `inferred_kwargs = {"data_dir": data_dir}` etc. then doing something like `fill_kwargs = inferred_kwargs | filepath_kwargs` and then doing `fill_template(filepath_temp, **fill_kwargs)`.

brendan-m-murphy (approved):
It looks good to me if it adds everything you need. In the issue you mentioned another optional part of the path, but maybe that is beyond the scope of this PR?

fluxy/io.py
brendan-m-murphy: Or I guess just make a `fill_kwargs` dict and then update it with `filepath_kwargs` before calling `fill_template`.

fluxy/io.py
brendan-m-murphy: Anyway, this isn't really important at this stage.

melodb:
> It looks good to me if it adds everything you need. In the issue you mentioned another optional part of the path, but maybe that is beyond the scope of this PR?

I realized that the user can specify the second optional part of the path in data_dir directly. Taking into account that in this case data_dir is defined model by model, there is no need to specify a sub-folder in a different parameter.

fluxy/io.py
melodb: As data_dir is defined model by model. The user can define it directly in data_dir

fluxy/io.py
melodb: Is this what you had in mind?
I add species_print, period and suffix afterwards because I don't want them to be overwritten by filepath_kwargs.

```
    # Define dictionary with template arguments
    fill_kwargs = {
        "data_dir": data_dir,
        "model_dir": base_model_name,
        "species_dir": species,
        "sub_dir": sub_dir,
        "model_name": model_name,
    }

    fill_kwargs.update(
        {
            **filepath_kwargs,
            "species_print": species_print,
            "period": period,
            "suffix": file_pattern,
        },
    )

    # Get full path to file
    filepath = fill_template(
        filepath_temp,
        **fill_kwargs,
    )
```

fluxy/io.py
melodb: I'm not sure I follow. Would this separate function be called all the time or only in specific cases? It's not clear to me why this change would allow for custom file structures. Isn't that hard-coded in the call to make_template? If you have something in mind, feel free to commit directly. In this way you become also an author of this PR which is fair.

fluxy/io.py
brendan-m-murphy: Yea that looks good (maybe not actually shorter like I thought).

fluxy/io.py
brendan-m-murphy: I think a separate PR would be better since this PR already adds the feature you need.

I'm not 100% what the change would look like, but basically it would separate the current logic into parts that 1) get arguments from the model names like you do now, 2) find files given a template and dictionary with all the info to fill the template. So 1) would do some work to fill the dictionary needed for part 2), but you could use a different file structure if you provide a dictionary will all the info needed to file your template.

brendan-m-murphy (approved):
Okay looks good to me! I'll make a new issue to add the "custom file structure" feature.

fluxy/io.py
melodb: But I think it looks cleaner :)

PR #245, opened 2025-12-02
Remove mandatory vars

melodb:
I have changed the order of input variables in `read_model_output` because I wanted to set a default for `data_dir`. I made all necessary changes to functions, notebooks and tests. Tests are passing nicely.

@lionel42, is this ok? I would prefer not to rename the function and create a new one just to support the old arguments order.


lionel42 (changes requested):
@melodb  This is a huge breaking change, since you change the order of the arguments in one of the most important function of fluxy. 

Basically everyone will need to update their scripts, so it is a bad idea. 

If you want to add a default to data_dir, since it is the first argument , you will have to add a default to all the other arguments and keep the order. 

Other wise you could add another function that reads the arguments and ensure compabitibility with the previous argument order.

I prefer the first solution, since it is easier and cleaner to implement.

melodb:
I also prefer the first option. It won't require changes in the tests/notebooks so it might be easier to create a new PR.

PR #246, opened 2025-12-02
Add defaults

melodb:
Add defaults to all read_model_output variables.
Make species_info not mandatory in select.py

The default values are useful when data_dir is defined inside model_filepath_dict.

### DIFF
--- a/fluxy/operators/select.py
+++ b/fluxy/operators/select.py
@@ -55,9 +55,10 @@ def slice_flux(
 
     """
     ds_all_sliced = dict()
+    species_info = config_data.get("species_info",{})
 
     if species is not None:
-        species_info = config_data["species_info"][species]
+        species_info = species_info.get(species, None)
 
     if type(start_date) is str:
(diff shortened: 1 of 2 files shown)


fluxy/operators/select.py
```diff
@@ -55,9 +55,10 @@ def slice_flux(
 
     """
     ds_all_sliced = dict()
+    species_info = config_data.get("species_info",{})
 
     if species is not None:
-        species_info = config_data["species_info"][species]
+        species_info = species_info.get(species, None)
```
lionel42: Don't name `species_info` 2 different objects

lionel42 (approved):

fluxy/operators/select.py
melodb: species_info is only used as an input to function scale_variables if species is not None. So I think I can simply write the following:
`species_info = config_data.get("species_info",{}).get(species, None)`

PR #252, opened 2025-12-09
small enhancements

alexdanjou:
- Add option to select sites while doing the stats
- change default for `resample_uncert_correlation` in `plot_country_flux` so that it is assuming no correlation by defaults when calculating the uncertainties of resampled data (as it should).


alexdanjou:
Whaouh forgot that wrong branch here

PR #253, opened 2025-12-09
Small enhancements

alexdanjou:
- Add option to select sites while doing the stats
- Change default for `resample_uncert_correlation` in `plot_country_flux` so that it is assuming no correlation by defaults when calculating the uncertainties of resampled data (as it should).

### DIFF
--- a/fluxy/operators/mf.py
+++ b/fluxy/operators/mf.py
@@ -114,6 +114,7 @@ def stats_mf(
     stats_type: Literal[
         "prior", "posterior", "prior_above_BC", "posterior_above_BC"
     ] = "prior",
+    sites: list = None,
 ) -> pd.DataFrame:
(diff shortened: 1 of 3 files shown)


mo-aliceramsden (approved):
Looks good to me. Good idea to add in an option to select sites when calculating the stats, that should speed up analysis, when we don't need to look at all the sites together

melodb (approved):

melodb:
@alexdanjou, before merging, would you mind checking if this works for the Taylor plots as well?

alexdanjou:
> @alexdanjou, before merging, would you mind checking if this works for the Taylor plots as well?

It is :) (I use it mainly for them, because otherwise it's impossible to see anything with methane 40 sites...)

PR #256, opened 2025-12-09
Minor plotting adjustment options for UK NID-format plots and new all-species stacked bar chart plot

mo-aliceramsden:
Various minor plotting options, so the met office group can use fluxy to create plots for the UK NID, in the established NID-annex format.
A new plotting function for a stacked bar chart with all species. Includes both the inventory and one inverse model.

New plotting options include:
- Removing titles and labels from spatial plots
- Moving x axis labels to the centre of the period in flux timeseries plots
- Plotting all data or only data with overlapping time periods (e.g. if option is selected, it will plot all_hfc data for 2012-2024 even if all species don't have data for 2012. This is useful for some of our NID plots, if we want to plot all_hfc even if some of the minor HFCs are missing)
- Plotting inventory uncertainty as error bars (think this is info is currently only available for the UK in the inventory netcdfs, but could be added to more countries, if this info is available in their NIDs)

Everyone's very busy with paper writing at the moment, so definitely no rush to get this merged in to devel! I just thought I'd get this PR ready and aligned with the latest devel now, as the met office group will be using this branch in January to create NID annex plots.


fluxy/operators/flux_combine.py
```diff
@@ -18,7 +19,7 @@ def combine_dataset(
         A dictionnary with 'combined' as key and the combined dataset as value.
     """
     ds_to_combined = [ds for i, ds in enumerate(ds_all.values()) if plot_combined[i]]
-    ds_to_combined_aligned = align_time(ds_to_combined)
+    ds_to_combined_aligned = align_time(ds_to_combined,only_overlapping=only_overlapping)
```
alexdanjou: missing space after ,

fluxy/operators/flux_prepare_inventory.py
```diff
@@ -60,8 +60,10 @@ def retrieve_inventories(
         )
 
     ds_sectors = {y: list() for y in inventory_years}
+    ds_uncert_sectors = {y: list() for y in inventory_years}
+    
     for sector in sectors:
-        tmp = _retrieve_inventories_sector(
+        tmp,tmp_stdev = _retrieve_inventories_sector(
```
alexdanjou: missing space after ,

fluxy/operators/flux_prepare_inventory.py
```diff
@@ -118,16 +116,18 @@ def _retrieve_inventories_sector(
...
 
     inventories_list = list()
+    inventories_uncert_list = list()
 
     inv_cmap = get_cmap("Greys")
     inv_colors = [inv_cmap(i) for i in np.linspace(0.5, 0.9, len(inventory_years))]
 
     for year, inv_color in zip(inventory_years, inv_colors):
-        ds_inv = extract_region_inventory_flux(
+        ds_inv,ds_inv_stdev = extract_region_inventory_flux(
```
alexdanjou: missing space

fluxy/operators/flux_prepare_inventory.py
```diff
@@ -75,17 +77,13 @@ def retrieve_inventories(
...
-                tmp[i].expand_dims(
-                    dim={
-                        "sector": [
-                            sector,
-                        ]
-                    }
-                )
-            )
+            ds_sectors[y].append(tmp[i].expand_dims(dim={"sector": [sector]}))
+            if tmp_stdev is not None:
```
alexdanjou: I'm not sure that this if statement is necessary. My impression is that it will be never `None` but will be instead a list of `None`.

fluxy/operators/flux_prepare_inventory.py
```diff
@@ -140,5 +140,9 @@ def _retrieve_inventories_sector(
         )
         ds_inv.attrs["plot_color"] = inv_color
         inventories_list.append(ds_inv.sel(time=slice(start_date, end_date)))
+        if ds_inv_stdev is not None:
```
alexdanjou, line 147: maybe put the `return` in this `if` statement? First case, `return inventories_list, inventories_uncert_list`, else `return inventories_list, None`. This would make my previous comment obsolete. It's the say number of lines, but you don't output a list of `None`, which seems cleaner to me. (Especially if you want to test that it exists afterwards, as you do)

fluxy/operators/regions.py
```diff
@@ -329,18 +336,26 @@ def extract_region_inventory_flux(
...
+        if country_search in inv_ds["country"]:
+            inv_stdev_ds = inv_stdev_ds.sel(country=country_search)
+        elif country in inv_ds["country"]:
+            inv_stdev_ds = inv_stdev_ds.sel(country=country)
+    else:
+        inv_stdev_ds = None
 
     if country_search in inv_ds["country"]:
-        return inv_ds.sel(country=country_search)  # new format
+        return inv_ds.sel(country=country_search),inv_stdev_ds  # new format
```
alexdanjou: missing space

fluxy/operators/regions.py
```diff
@@ -329,18 +336,26 @@ def extract_region_inventory_flux(
...
     # Look for the code if country_codes is defined, otherwise assume the code was given as input
     country_search = country_codes.get(country, country)
+    
+    if inv_stdev_ds is not None:
+        inv_stdev_ds = inv_stdev_ds * scaling_factor * gwp
+        if country_search in inv_ds["country"]:
+            inv_stdev_ds = inv_stdev_ds.sel(country=country_search)
+        elif country in inv_ds["country"]:
+            inv_stdev_ds = inv_stdev_ds.sel(country=country)
+    else:
```
alexdanjou: unecessary

fluxy/operators/regions.py
```diff
@@ -329,18 +336,26 @@ def extract_region_inventory_flux(
...
-        return inv_ds.sel(
-            country=country
-        )  # old format (would only work if the user specifies the country name)
+        return inv_ds.sel(country=country),inv_stdev_ds  # old format (would only work if the user specifies the country name)
 
     # if grouped countries:
     available_countries = inv_ds["country"].values.astype(str)
     dict_regions: dict[str, str] = r_data.get("regions", {})
-
+    inv_stdev_ds = None
```
alexdanjou: i would remove that and put directly `return inv_ds.sum(dim="country", keep_attrs=True), None` instead of `return inv_ds.sum(dim="country", keep_attrs=True),inv_stdev_ds` further down

fluxy/plots/flux_map.py
```diff
@@ -676,21 +688,21 @@ def plot_flux_map_over_time(
 
             # Adjust ticks layout
             if is_single_season:
-                if col in [0, 1]:
+                if col in [0, 1] or include_title_and_labels == False:
```
alexdanjou: `include_title_and_labels == False` replace by `not include_title_and_labels`? same further down

fluxy/plots/flux_map.py
```diff
@@ -221,13 +223,17 @@ def plot_flux_map(
                 ax_i.set_xticklabels([])
             if col > 0:
                 ax_i.set_yticklabels([])
+                
+            if include_title_and_labels == False:
```
alexdanjou: `not include_title_and_labels`?

fluxy/plots/flux_timeseries.py
```diff
@@ -465,10 +467,16 @@ def add_inventory_barplot(
...
         annex_mode: If True, replace Inventory label with a more concise version for National Inventory Report Annexes.
+        plot_inventory_uncertainty: If True and uncertainty available, plots inventory error bars. If a list is provided, should be of same size as inventory_years.
     Returns:
         res: dataframe with one line per timestamp and 7 columns ("type", "model", "sector", "country", "species", 
             "time", "mean_val")
     """
+    
+    if type(plot_inventory_uncertainty) == bool:
+        plot_inventory_uncertainty = [plot_inventory_uncertainty] * len(inventory_years)
+        logger.warning("Plotting inventory uncertainty for all inventory_years. "+
```
alexdanjou: I would downgrad this as an info

fluxy/plots/flux_timeseries.py
```diff
@@ -478,7 +486,7 @@ def add_inventory_barplot(
         end_date_inv = str(max([np.datetime64(date) for date in end_date]))
     else:
         end_date_inv = end_date
-    inventories_to_plot = retrieve_inventories(
+    inventories_to_plot,inventories_uncert_to_plot = retrieve_inventories(
```
alexdanjou: space :D

fluxy/plots/flux_timeseries.py
```diff
@@ -495,6 +503,11 @@ def add_inventory_barplot(
     res = pd.DataFrame()
     for i_inv, inventory in enumerate(inventories_to_plot):
         time_as_datetime = inventory.time.values.astype("datetime64[D]").tolist()
+        
+        if plot_inventory_uncertainty[i_inv] and inventories_uncert_to_plot[i_inv] is not None:
```
alexdanjou: `is not None` is unecessary

fluxy/plots/flux_timeseries.py
```diff
@@ -628,7 +641,7 @@ def prepare_inventory_sector_barplot(
 
     """
 
-    inventories = retrieve_inventories(
+    inventories,inventories_stdev = retrieve_inventories(
```
alexdanjou: space :)

fluxy/plots/flux_timeseries.py
```diff
@@ -643,8 +656,9 @@ def prepare_inventory_sector_barplot(
     )
 
     inventories = [inv.to_dataset(name="inv_data") for inv in inventories]
+    inventories_stdev = [inv_std.to_dataset(name="inv_data_stdev") if inv_std is not None else None for inv_std in inventories_stdev]
 
-    return inventories
+    return inventories,inventories_stdev
```
alexdanjou: space

fluxy/plots/flux_timeseries.py
```diff
@@ -643,8 +656,9 @@ def prepare_inventory_sector_barplot(
     )
 
     inventories = [inv.to_dataset(name="inv_data") for inv in inventories]
+    inventories_stdev = [inv_std.to_dataset(name="inv_data_stdev") if inv_std is not None else None for inv_std in inventories_stdev]
```
alexdanjou: `is not None` unecessary

fluxy/plots/flux_timeseries.py
```diff
@@ -760,6 +774,7 @@ def add_xlims_and_ticks(
         res_dict: dictionnary containing the data plotted. Should have one key per regions plotted, the values being dictionnaries with 3 keys: "inventory", "posterior" and "prior";
             whose values are the output of add_inventory_barplot, add_posterior_plot, add_prior_plot). The time data stored in them is used to infer the xlims.
         aggreg_month: if True, the data plotted are supposed to be a monthly aggregated so 12 stciks are created, whose labels are the 3 first letters of each month.
+        xticks_at_centre: if True, set the x ticks at the centre of each time period (year or month) rather than at the beginning.
```
alexdanjou: is it really doing it for year or month or just year? I don't see any changes when it's monthly data

fluxy/plots/flux_timeseries.py
```diff
@@ -1004,10 +1042,13 @@ def plot_country_flux(
             )
 
             if add_prior:
-                prior_df = add_prior_plot(ax, ds_region, annex_mode, add_prior_unc)
-                plotted_data_df = pd.concat(
-                    [plotted_data_df, prior_df], ignore_index=True
-                )
+                try:
```
alexdanjou: I'm not found of `try` and `except` in the code. If you have a different error that arise inside it will be catch in it and it's then a pain to debug.
Can you amend maybe `add_prior_plot`, or make a test before, to see if there is prior data and skip this if not.

fluxy/plots/flux_timeseries.py
```diff
@@ -1265,3 +1312,160 @@ def plot_country_sector_flux_bar(
...
+        inventory_filename: Name of inventory file: {inventory_filename}_{species}_{inventory_year}
+        data_dir: Path to top data directory, used to read inventory data files.
+        sector: Sector to plot.
+        country_flux_units_print: Units for fluxes to be printed on inventory bars.
+        y_lim: Y-axis limits for the plot.
+    Returns:
+        fig: A stacked bar plot of all species for the region.
+    """
+
+    species_colors = {'ch4':'dodgerblue',
```
alexdanjou: move to config. species data maybe?

fluxy/plots/flux_timeseries.py
```diff
@@ -1265,3 +1312,160 @@ def plot_country_sector_flux_bar(
...
+            end_date,
+            country_flux_units_print,
+            s_data,
+            r_data,
+            inventory_years,
+            inventory_filename,
+            sectors=sector,
+        )
+
+        if inventories_uncert_to_plot[species][0] is None:
```
alexdanjou, line 1486: `if not inventories_uncert_to_plot[species][0]`

fluxy/plots/flux_timeseries.py
```diff
@@ -1265,3 +1312,160 @@ def plot_country_sector_flux_bar(
...
+
+        if s == 0:
+            plot_times = ds_to_plot[species][models[s]].time.values
+            uncert_combined = ds_to_plot[species][models[s]]['posterior_upper'].values-ds_to_plot[species][models[s]]['posterior_lower'].values
+            inventories_uncert_combined = inventories_uncert_to_plot[species][0]
+        else:
+            uncert_combined = np.sqrt(uncert_combined**2 + (ds_to_plot[species][models[s]]['posterior_upper'].values - ds_to_plot[species][models[s]]['posterior_lower'].values)**2)
+            inventories_uncert_combined = np.sqrt(inventories_uncert_combined**2 + inventories_uncert_to_plot[species][0]**2)
+
+    for s,species in enumerate(all_species):
```
alexdanjou: space :D

fluxy/plots/flux_timeseries.py
```diff
@@ -1265,3 +1312,160 @@ def plot_country_sector_flux_bar(
...
+        ax.bar(plot_times,
+            ds_to_plot[species][models[s]]['posterior'].values,
+            width=width,
+            bottom=bottom,
+            color=species_colors[species],
+            label=s_data.get(species, {}).get('species_print', species),
+            yerr=uncert,
+            error_kw={'capsize':2})
+
+        if s == 0:
```
alexdanjou, line 1542: is it possible to regroup the `s == xxx` statements?

fluxy/operators/regions.py
```diff
@@ -129,9 +129,10 @@ def _extract_region_flux_sector(
                 ds_region = ds_region.sel({"country_2": country_list})
 
             for v in ["posterior", "prior"]:
-                ds_region[v] = ds_region[f"flux_{sector}_{v}_country"].sum(
-                    dim="country", keep_attrs=True
-                )
+                if f"flux_{sector}_{v}_country" in ds_region.keys():
```
alexdanjou: The `if` testing the variables presence, is this because sometimes you don't have a prior? Maybe you can tests that further up and then correct `target_vars` and the list in which we're iterating for v. Also you will need to change the returns description as all the variables won't be present anymore

alexdanjou: It may not be shorter, but I think it will improve readability (I had to reach `try:.. except: ...` in the other file to understand that sometimes you didn't have prior and thus understand the necessity of this ^^).

alexdanjou (commented):
Great! I made a few comments, mostly about readability :)

fluxy/operators/flux_prepare_inventory.py
mo-aliceramsden: That if statement is contained in a loop of the inventory_years, so I'd have to add another if statement later, for that to work. I think I kept inventories_uncert_list as a list because it could be that one (but not not all) of inventory_years has uncertainty info, which would be lost if we just returned None if any years were missing the info

fluxy/operators/regions.py
mo-aliceramsden: I think I kept this in because sometimes inv_stdev is None and the .sel line fails then, so the .sel can't be included in the return line

fluxy/plots/flux_timeseries.py
mo-aliceramsden: I think I kept the if s==0/else statements separate as I needed one before the data was plotted, to set the lower limit of the stacked bars, and one after this, to adjust the new lower limit of the bars. But I'll have a think to see if there's a better way of writing that

fluxy/operators/regions.py
mo-aliceramsden: yes, that's it, I added that line so we could plot datasets without priors. I've added some other datasets (e.g. the CEH monthly methane prior) but created versions of these files in inversion-output format, so we can do all the detailed comparison with these that we do with the inversions (seasonal cycles, compare to inventories etc.) but the files don't contain 'prior' variables. Your suggestion is good, I've put an adjustment in to fix that

fluxy/operators/flux_prepare_inventory.py
alexdanjou: ah ok!

fluxy/plots/flux_timeseries.py
alexdanjou: No no worries, I didn't realized that, sorry.

fluxy/operators/regions.py
alexdanjou: Maybe I'm missing something, but the if statement is testing if `inv_stdev_ds` is None or not. It is thus entering the `else` condition only if `inv_stdev_ds==None`. So I don't understand the need to do again `inv_stdev_ds=None`.
I realized that my comment was one line too high (I was talking about this line and the one after), maybe that was misleading. My brain is not at it's peak lately so maybe I'm saying nonsense

mo-aliceramsden: Yes I understand now, thanks for checking this again. I'll put a fix in for that. And I completely understand about having a tired brain, I'm the same, definitely reached the time of the year where we all need a break!

configs/models_info.json
```diff
@@ -1,6 +1,6 @@
 {
     "filename_tags":{
-        "std": "<model>_obs_<model>_baseline_optimized",
+        "std": "PARISNID2026",
```
melodb: In general, I would avoid commiting edits to this file (unless they correspond to new settings from new options). But I understand that the update is quite usefull for the PARIS users. Nevertheless, I would not rename "std" because it is usefull to reproduce the NID2025 plots. I would rather create a new tag. For example "paris" or "annex"

configs/regions_info.json
```diff
@@ -22,7 +22,8 @@
         "GREECE": "GRC",
         "SPAIN": "ESP",
         "PORTUGAL": "PRT",
-        "NORWAY": "NOR"
+        "NORWAY": "NOR",
+        "GermanFactory":"GermanFactory"
```
melodb: This is just my opinion, but I would not commit these type of changes which are specific to a given project. I think we should keep the json files general, but of course also practical for the PARIS team which are the main users. When it comes to non-PARIS projects, in my view it's not practical that everyone starts adding their specific needs. I think that it would be better that people save their own settings files in a specific branch or local folder.

configs/species_info.json
```diff
@@ -2,7 +2,8 @@
     "ch4": {
         "species_print": "CH$_4$",
         "gwp": 28,
-        "molar_mass": 16.04
+        "molar_mass": 16.04,
+        "color": "dodgerblue"
```
melodb: I don't love the idea of mixing species information with plotting settings. I understand that you need this for the stacked bar plot, but I'm afraid it would lead to confusion because in the mf timeseries plots we use the colors per model defined in `config.py`. `config.py` is for sure not the best place to set up colors, but I would give priority to keeping the code consistent and maybe one day move the colors all together to a different place. What do you think in adding variable `species_color_palette` in `config.py`? You can either specify a dictionary or just a list from where colors are taken in order.

fluxy/operators/flux_align_dataset.py
```diff
@@ -6,12 +6,14 @@
 logger = logging.getLogger(__name__)
 
 
-def align_time(ds_list: list[xr.Dataset]) -> list[xr.Dataset]:
+def align_time(ds_list: list[xr.Dataset],
+               only_overlapping: bool = True) -> list[xr.Dataset]:
     """
     Check the time coordinates of a list of xarray datasets and, if they differ, align them with the time coordinate of the first dataset in the list.
```
melodb: Maybe once this function is done, the description can be updated.

fluxy/operators/flux_align_dataset.py
```diff
@@ -34,10 +36,15 @@ def align_time(ds_list: list[xr.Dataset]) -> list[xr.Dataset]:
...
-    if period is not None:
+    if period is not None and only_overlapping:
         min_date = max([x.time.min() for x in ds_list]) - period / 2
         max_date = min([x.time.max() for x in ds_list]) + period / 2
         ds_list = [ds.sel(time=slice(min_date, max_date)) for ds in ds_list]
+        
+    else:
+        all_min = [x.time.min() for x in ds_list]
+        min_date_loc = [i for i,date in enumerate(all_min) if date == min(all_min)][0]
+        ds_list = [ds.reindex(indexers={'time':ds_list[min_date_loc]['time']},fill_value=np.nan,method=None) for ds in ds_list]
```
melodb: Isn't `reindex` too restrictive? In addition to function `create_flux_total_fgases`, `align_time` is also used to prepare data for the spatial maps and to combine datasets. Therefore, it is used with datasets outputted by different inversion systems, which might have small shifts in the time stamps that we want to neglect (for example, one dataset with timestamp "15.07.2018 00:00:00" and other with "15.07.2018 12:00:00"). Currently, we re-write those timestamps according to the first dataset but we don't delete them. As far as I understand, `reindex` would delete them.

Moreover, the dataset starting earlier might not be the more comprehensive dataset. For example, its end date can be sooner than the end date of the other datasets. You might want to check `max(all_max)` as well.

fluxy/operators/regions.py
```diff
@@ -106,6 +106,8 @@ def _extract_region_flux_sector(
         "prior_lower",
         "prior_upper",
     ]
+        
+    v_present = dict(zip(ds_all.keys(),[['posterior','prior'] if f"flux_{sector}_prior_country" in ds else ['posterior'] for m,ds in ds_all.items()]))
```
melodb: Taking into account that "flux_total_prior_country" is a mandatory variable, I think this variable is assumed to exist thoughtout the code regardless of the sector. If it exists for some datasets and not for others, did you check if `add_prior = True` works fine, both with `plot_combined = True/False`?

fluxy/operators/regions.py
```diff
@@ -329,18 +337,23 @@ def extract_region_inventory_flux(
     country_codes = r_data.get("country_codes", {})
     # Look for the code if country_codes is defined, otherwise assume the code was given as input
     country_search = country_codes.get(country, country)
+    
+    if inv_stdev_ds is not None:
+        inv_stdev_ds = inv_stdev_ds * scaling_factor * gwp
+        if country_search in inv_ds["country"]:
+            inv_stdev_ds = inv_stdev_ds.sel(country=country_search)
+        elif country in inv_ds["country"]:
+            inv_stdev_ds = inv_stdev_ds.sel(country=country)
```
melodb, line 361: These if-statements already exists below (L.348, L.350) so maybe you can follow the same approach used for `inv_ds` and return `inv_stdev_ds.sel(country=country_search)` or `inv_stdev_ds.sel(country=country)`. Moreover, you might want to add "units" and "year" as attributes to `inv_stdev_ds` in analogy with `inv_ds` (L.333-334).

fluxy/operators/regions.py
melodb, line 275: Maybe update function description with inventory stdev.

fluxy/plots/utils.py
melodb, line 243: Update function description

fluxy/operators/regions.py
melodb, line 58: Warning: Lionel did some changes to this function in his PR which might lead to conflits (see PR #248)

fluxy/plots/flux_timeseries.py
```diff
@@ -465,10 +467,16 @@ def add_inventory_barplot(
...
+        plot_inventory_uncertainty: If True and uncertainty available, plots inventory error bars. If a list is provided, should be of same size as inventory_years.
     Returns:
         res: dataframe with one line per timestamp and 7 columns ("type", "model", "sector", "country", "species", 
             "time", "mean_val")
     """
+    
+    if type(plot_inventory_uncertainty) == bool:
+        plot_inventory_uncertainty = [plot_inventory_uncertainty] * len(inventory_years)
+        logger.info("Plotting inventory uncertainty for all inventory_years. "+
+                       "Turn this off setting `plot_inventory_uncertainty` as a list of True/False")
```
melodb: Can you use function `update_list_params` and call it inside `prepare_data_to_plot`?

fluxy/plots/flux_timeseries.py
```diff
@@ -776,27 +791,33 @@ def add_xlims_and_ticks(
...
         xticks = np.array(
             [date(year, 1, 1) for year in range(min_x.year, max_x.year, step)]
         )
-        if (max_x.year - min_x.year) % step == 0:
-            xticks = np.append(xticks, max_x)
-        ax.set_xticks(xticks)
-        ax.set_xticklabels(xticks.astype("datetime64[Y]"))
-        ax.xaxis.set_major_locator(YearLocator())
+        if xticks_at_centre:
+            xticks = xticks.astype('datetime64[D]') + np.timedelta64(182, 'D')
```
melodb, line 788: If the frequency is monthly and `xticks_at_centre = True`, there will be 182 days added to xticks, which is not what we want. Maybe "or xticks_at_centre" should be deleted from the if-statement in L.794.

fluxy/plots/flux_timeseries.py
```diff
@@ -840,7 +868,7 @@ def add_legend(
                 else "legendHandles"
             )
             for l in leg.__getattribute__(handle_name)[
-                : (-1 if plot_inventory else None)
+                : (-n_inv if plot_inventory else None)
```
melodb, line 856: Will this work if we set `plot_inventory = False` and `inventory_years = ["2025","2024"]`?

fluxy/plots/flux_timeseries.py
```diff
@@ -939,6 +972,10 @@ def plot_country_flux(
         return_res: Wheter or not including a dictionnary with the results as output
         rolling_mean : If True, calculates a rolling mean (xx years) for each of the data to plot.
         aggreg_month: if True, plot the data aggregated by month. Used to study seasonnal cycle.
+        sector: Sector to plot. Can be 'total' or any of the sectors defined in the datasets.
+        only_overlapping: If True, only plot the time range where all models have data.
```
melodb: As far as I understand, in this function `only_overlapping` is only used when combining the datasets. So maybe the variable description can be updated to "If True, plots combined datasets only for the time range where all models have data."

melodb (commented):
Thanks for these updates! I did not have the chance to try the code nor the new function, but I leave you already with my comments on the code. I will try to test it soon.

configs/models_info.json
```diff
@@ -1,11 +1,13 @@
...
         "ibaseline": "<model>_obs_intem_baseline_fixed",
         "iall": "intem_obs_intem_baseline_fixed_intem_unc"
     },
 
     "model_labels":{
+        "InTEM_NAME_EUROPE_NID2026" : "InTEM NID2026",
+        "InTEM_NAME_EUROPE_NID2025" : "InTEM NID2025",
```
melodb: I would remove these edits.

configs/species_info.json
melodb: I now see that this was your initial approach and that @alexdanjou suggested moving the colors to species_info. I still think it would be better not to mix plotting settings with species intrinsic info. Maybe we can think of creating a "plot_settings.json" file in the future.

scripts/PARIS_inversion_results_NID2026.ipynb
melodb, line 1: Alternatively, you can also create a new branch originating from this one and add the notebook with your settings there. In this way you separate what will be merged from your specific edits to generate the NID. Whenever you make a change here you can merge this branch with your NID-notebook branch to keep it updated. Just a suggestion.

fluxy/operators/flux_align_dataset.py
melodb: **Some updated notes after testing the code:**
- In the curent implementation, you only check for the minimum date and take the first dataset with the minimum date. I plotted all_hfc for the InTEM PARISNID2026 dataset and it works fine with the current species_info file because "hfc125" is the first element. If I change the order and place "hfc4310mee" first, option `only_overlapping = False` will not work nicely: the date is aligned based on "hfc4310mee" time values and 2023 is not plotted. As I have previously suggested, I think the max time can be checked to make it more robust.
- Then, regarding my concern on the use of `reindex`, I now see that `only_overlapping` is not an input to the spatial map function so option `only_overlapping = False` cannot be specified. However, it is an input to functions "plot_country_flux" and "plot_country_sector_flux_bar". Even though we don't use it in our notebook, nothing prevents a user for specifying it as an input to combine datasets. I have added it as an input to plot_country_flux and tested option `only_overlapping = False` when plotting the combined HFC-32 emissions of InTEM, ELRIS and RHIME (PARISNID2026). Due to `reindex`, the combined dataset takes into account only RHIME data. I would then suggest that you remove variable `only_overlapping` from the argument list of "plot_country_flux" and "plot_country_sector_flux_bar" to make sure that they are only used with `only_overlapping = True`.
- The current implementation of `reindex` works fine to combine all_hfc/pfc from the same inversion system because we always output exactly the same timestamp for all species. I think it would be more robust if we allow some deviations like we currently do to combine different inversion systems. I don't think this needs to be taken care now, but maybe create an issue on this limitation. We can take care of it in a future PR.
- Finally, would it be possible to print a warning for each pair (year, species) which is set to NaN when `only_overlapping = False`? I'm a bit uncomfortable with a "Total HFC" plot which is made based on a list of HFCs provided in species_info and to have absolutely no information about the species being ignored.

**Summary of actions:**
- check max time in addition to min time
- delete `only_overlapping` as input from "plot_country_flux" and "plot_country_sector_flux_bar"
- create an issue about the limitation of `reindex`
- print a warning for each year/species being omited from Total HFC/PFC plot

(sorry for the long post)

fluxy/operators/flux_align_dataset.py
melodb: The limitations of `only_overlapping = False` can be highlighted in the function description.

fluxy/plots/flux_timeseries.py
melodb: **Updated comment after testing:**
When I set:
```
plot_inventory = True
plot_inventory_uncertainty = True
```

The code fails with error "TypeError: object of type 'NoneType' has no len()" when trying to get `len(inventory_years)`. It seems that `inventory_years` is never converted to a list.
Differently from my first suggestion (to pass the inventory options to "prepare_data_to_plot"), I would rather call "update_list_params" directly in "plot_country_flux" before the call to "add_inventory_barplot" inside `if plot_inventory`. This would make sure that `plot_inventory` and `plot_inventory_uncertainty` are always lists of the correct size. Check what you think.

fluxy/operators/regions.py
melodb: As I don't have a dataset without prior, I cannot test this. I would be happy if you could test this case. If the code breaks, you don't need to fix it is this PR. Just create an issue.

mo-aliceramsden:
I've made some temporary edits to some files (e.g. PARIS_inversion_results.ipynb) so others at the met office can use these edits whilst we're creating the NID. But I'll revert these changes before merging to devel.

fluxy/plots/flux_map.py
```diff
@@ -46,8 +46,10 @@ def plot_flux_map(
     zoom_degree: float = 1,
     only: Literal["posterior", "prior", "diff"] | None = None,
     fallback_sites: list[str] | None = None,
-    resample_uncert_correlation=False,
+    resample_uncert_correlation: bool = False,
     sector: str = "total",
+    include_title_and_labels: bool = True,
+    swap_symbol: bool = False
```
melodb: I think it would be more flexible to simply allow the user to specify the symbols and set 'o' and '^' as defualts. For example:
```
site_marker: str = 'o',
city_marker: str = '^'
```
In a future PR, we can reorganize all these plotting settings into a dictionary to reduce the number of input variables to the function.

fluxy/plots/flux_map.py
```diff
@@ -615,7 +639,7 @@ def plot_flux_map_over_time(
     is_diff = "diff" in var
     cmap = cmap_diff if is_diff else cmap
     border_color = c_border_diff if is_diff else c_border
-    marker_color = "black" if is_diff else "red"
+    marker_color = "black" if is_diff else "darkblue"
```
melodb: Maybe set marker color as input parameter too. For example:
```
marker_color: str = "black"
marker_color_diff: str = "red"
```

fluxy/plots/flux_map.py
```diff
@@ -501,6 +520,9 @@ def plot_flux_map_over_time(
     fallback_sites: list[str] | None = None,
     resample_uncert_correlation: bool = False,
     sector: str = "total",
+    include_title_and_labels: bool = True,
+    add_gridlines: bool = False,
+    swap_symbol: bool = False
```
melodb: For consistency, I would add these new input variables to all flux_map functions: "plot_flux_map", "plot_flux_map_model_comparison", "plot_flux_map_over_time", "plot_flux_map_combined_models_comparison"

fluxy/plots/flux_timeseries.py
```diff
@@ -491,10 +499,15 @@ def add_inventory_barplot(
...
-
+    
     res = pd.DataFrame()
     for i_inv, inventory in enumerate(inventories_to_plot):
         time_as_datetime = inventory.time.values.astype("datetime64[D]").tolist()
+        
+        if plot_inventory_uncertainty[i_inv] is True and inventories_uncert_to_plot[i_inv] is not None:
+            yerr = inventories_uncert_to_plot[i_inv].values
+        else:
+            yerr = None
```
melodb: When the uncertainty value exists but it's equal to zero, the top/bottom of the error bar are plotted which makes the plot a bit weird. Maybe you can also set `yerr = None` if all uncertainty values are zero.

fluxy/operators/regions.py
```diff
@@ -329,18 +337,23 @@ def extract_region_inventory_flux(
...
     if country_search in inv_ds["country"]:
-        return inv_ds.sel(country=country_search)  # new format
+        return inv_ds.sel(country=country_search), inv_stdev_ds  # new format
     elif country in inv_ds["country"]:
-        return inv_ds.sel(
-            country=country
-        )  # old format (would only work if the user specifies the country name)
+        return inv_ds.sel(country=country), inv_stdev_ds  # old format (would only work if the user specifies the country name)
 
     # if grouped countries:
```
melodb, line 371: The plotting of inventory uncertainty is not implemented for grouped regions (e.g. BENELUX). If there is an easy smart way to do it, I would suggest implementing it. Otherwise, print a warning stating that the uncertainty error bar will not be plotted, and create an issue so that we don't forget about this.

fluxy/plots/flux_timeseries.py
melodb: **Updated note after testing:**
I now see that the x-axis does not look good when plotting only a couple of months: major ticks are still set to years. For example:
```
species = 'ch4'
start_date = '2018-01-01'
end_date = '2018-06-01'
```
In this case, the only tick label is "2018". If `xticks_at_centre = True`, the label goes beyond the extent of the plot. Therefore, I would keep my suggestion that "or xticks_at_centre" should be deleted from the if-statement in L.794.

I have worked a bit on the x-axis labels, mainly for the mf plot, but I can also improve it here. I will submit a PR after all pending PRs and fluxy renaming.

fluxy/plots/flux_timeseries.py
melodb: I have tested it ans it works fine.

fluxy/plots/flux_timeseries.py
melodb: As mentioned in other comment. I think this variable should be removed from the argument list of this function.

scripts/PARIS_inversion_results.ipynb
```diff
@@ -915,6 +923,87 @@
...
+    "        ds_all_flux_scaled[species] = read_flux_total_fgases(data_dir,species,[models[s]],config_data,\n",
+    "                                                         regions,start_date,end_date,\n",
+    "                                                         period=period[s],only_overlapping=only_overlapping)\n",
+    "    else:\n",
+    "        ds_all_flux[species] = read_model_output(data_dir,\"flux\",species,[models[s]],config_data,period=period[s],read_standard_run=read_standard_run)\n",
+    "        ds_all_flux_scaled[species] = slice_flux(ds_all_flux[species],config_data,start_date,end_date,species=species,country_flux_units_print=country_flux_units_print)\n",
+    "        \n",
+    "#not used in this plot, but neede to process data without errors\n",
+    "model_colors = set_model_colors(models)\n",
+    "model_labels = set_model_labels(models,config_data,get_labels_from_file)"
```
melodb: The code should work fine without these variables. From what I have tested, if you remove these lines and remove `model_colors` and `model_label` in the call for "plot_all_species_stacked_bar" in the next cell, the plot works fine.

fluxy/plots/flux_timeseries.py
```diff
@@ -1265,3 +1309,154 @@ def plot_country_sector_flux_bar(
...
+                                 model_colors: dict[str, str] = {},
+                                 model_labels: dict[str, str] = {},
+                                 start_date: str | None = None,
+                                 end_date: str | None = None,
+                                 inventory_years: list[str] | None = None,
+                                 inventory_filename: str = "UNFCCC_inventory",
+                                 plot_inventory_uncertainty: bool = True,
+                                 data_dir: str | None = None,
+                                 sector: str = "total",
+                                 country_flux_units_print: str = "Tg CO2-eq yr-1",
```
melodb: The units should not be an input to the function and should not have a default value. This leads to inconsistent plots. Maybe you can follow the approach used in "plot_country_flux", where function "get_posterior_unit" is used to get the units directly from the dataset attributes.

Moreover, I did not find any check that all units are equal and in CO2-eq. I have tested setting the units to "Tg yr-1" and I got a plot anyway, even though the values didn't make sense. I think an error should be raised in this case.

fluxy/plots/flux_timeseries.py
```diff
@@ -1265,3 +1309,154 @@ def plot_country_sector_flux_bar(
...
+            flux_sum = ds_to_plot[species][models[s]]['posterior'].values
+            inventory_sum = inventories_to_plot[species][0].values
+        else:
+            flux_sum += ds_to_plot[species][models[s]]['posterior'].values
+            inventory_sum += inventories_to_plot[species][0].values
+
+        ax.set_xticks(plot_times+(width/2))
+        ax.set_xticklabels((plot_times+(width/2)).astype('datetime64[Y]'))
+
+    ax.set_ylabel('Tg y$^{-1}$ CO$_2$-eq')
```
melodb: The units in the y-label should not be hard coded. See comment above.

scripts/PARIS_inversion_results.ipynb
```diff
@@ -915,6 +923,87 @@
...
+    "###################################\n",
+    "### edit variables in this block\n",
+    "all_species = ['nf3','sf6','all_pfc','all_hfc','n2o','ch4']\n",
+    "period = ['yearly','yearly','yearly','yearly','monthly','monthly'] #inversion period, must be a string or a list of the same length as models, e.g. ['monthly','yearly']\n",
+    "models = ['InTEM_NAME_EUROPE_FLAT_PARISNID2026',\n",
+    "          'InTEM_NAME_EUROPE_FLAT_PARISNID2026',\n",
+    "          'InTEM_NAME_EUROPE_std',\n",
+    "          'InTEM_NAME_EUROPE_std',\n",
+    "          'InTEM_NAME_EUROPE_EDGAR_PARISNID2026',\n",
+    "            'InTEM_NAME_EUROPE_EDGAR_PARISNID2026']\n",
```
melodb: In this function, the user is able to specify one model for each species. We could for instance read CH4 from InTEM, N2O from ELRIS and all_hfc from RHIME and plot everything stacked. Differently, when computing all_hfc/pfc, we force the user to define the `standard_run` in "models_info".  Inside "read_flux_total_fgases", `read_standard_run=True` is hard-coded in the call for "read_model_output". This approach guarantees that all flux data come from the same inversion model. Should we follow the same approach here? If yes, in your example, you would simply set `models = ['InTEM']`.

Moreover, you could allow the possibility of plotting one bar per model. For example, `models = ['InTEM', 'ELRIS', 'RHIME']` would plot 3 stacked bars, one after the other. Just a suggestion in case that's easy.

fluxy/plots/flux_timeseries.py
```diff
@@ -1265,3 +1309,154 @@ def plot_country_sector_flux_bar(
...
+            inventory_filename,
+            sectors=sector,
+        )
+
+        if inventories_uncert_to_plot[species][0] is None:
+            inventories_uncert_to_plot[species][0] = np.zeros_like(inventories_to_plot[species][0].values)
+
+        if s == 0:
+            plot_times = ds_to_plot[species][models[s]].time.values
+            uncert_combined = ds_to_plot[species][models[s]]['posterior_upper'].values-ds_to_plot[species][models[s]]['posterior_lower'].values
```
melodb, line 1492: `posterior_upper - posterior_lower` corresponds to 2 x stdev. This is normally what we plot with `fill_between`. I think `yerr = uncert` would plot an error bar equal to 2 x yerr, therefore, 4 x stdev. Is this what you want? In my view, it would be better to keep it consistent with "plot_country_flux". Therefore:
`uncert_combined = ds_to_plot[species][models[s]]['posterior_upper'].values-ds_to_plot[species][models[s]]['posterior'].values`

fluxy/plots/flux_timeseries.py
```diff
@@ -1265,3 +1309,154 @@ def plot_country_sector_flux_bar(
...
+                                 models: list[str],
+                                 regions: list[str],
+                                 config_data: dict[str, dict] = {},
+                                 model_colors: dict[str, str] = {},
+                                 model_labels: dict[str, str] = {},
+                                 start_date: str | None = None,
+                                 end_date: str | None = None,
+                                 inventory_years: list[str] | None = None,
+                                 inventory_filename: str = "UNFCCC_inventory",
+                                 plot_inventory_uncertainty: bool = True,
```
melodb, line 1404: Variable plot_inventory_uncertainty is not used anywhere. Uncertainty bars are always plotted.

melodb (commented):
Thank you Alice once again for this PR. The stacked bar species plot looks really nice! I have tested the code and added some more comments. Some are replies to previous comments of mine, I think you can click on "previous comment" to see the context. Otherwise check them in "Files changed".

I hope this is not too much work. Please feel free to disagree with some of the comments.

configs/regions_info.json
mo-aliceramsden: Good point, I've removed the GermanFactory option, as that was only used for short-term testing on this branch. 

I've added codes for England, Scotland, Wales and Northern Ireland, as we'll want to be able to plot those areas long term. Those devolved nations don't have offical 3-letter country codes, so I've created some, checking that those codes don't already exist in 'International Organization for Standardization' database.

configs/species_info.json
mo-aliceramsden: Good idea, I've added a dictionary of 'species_color_palette' to config.py

fluxy/operators/flux_align_dataset.py
mo-aliceramsden: Thanks for considering this so carefully, I've hopefully taken all your points into consideration (but do let me know if I've missed something!):

- I've updated the code which reindexes the datasets by time, so it does this based on a newly-created array of times based on min/max and the time period of each ds (rather than just using the first model with the earliest start date). There might be a shorter way of doing this, but this way does seem to be more robust that the reindex option, and should take into account any slight differences in the timesteps between models (as you mention above).
- I've taken the `only_overlapping` option out of `plot_country_flux `and `plot_country_sector_flux_bar`, you're right that this option isn't needed in our current setup if we're already aligning datasets with the call to `read_flux_total_fgases`.
- I've added a warning when `only_overlapping `is set to False, that prints out the dates used for each species and model. This does add a lot of text printed, when using multiple models for multiple locations, but this should hopefully make it a lot more transparent about what's being included.

fluxy/operators/regions.py
mo-aliceramsden: I think I put this option in so we can plot some other datasets as if they were inversion model outputs (e.g. EDGAR or UK CEH priors) which only have 'posterior' variables in their datasets. I've tried this locally with some test files, and it works when `plot_combined` is True, but not when it is False. I'm not sure we'll need a situation when this does work with `plot_combined` is True, as we'll have very few files of this type (with posterior variables, but no prior variables). I can still add an issue if you can think of a scenario where this functionality might be useful?

fluxy/plots/flux_timeseries.py
mo-aliceramsden: Yes, you're right, this does look odd with short timeseries, I'll remove the xticks_at_centre from the overall if statement here

fluxy/operators/regions.py
mo-aliceramsden: I'll create an issue to sort this later, because at the moment the inventory netcdfs only have uncertainty information for the UK anyway. Hopefully this is something we can include for all countries in the future, but I think we'd have to look through each national inventory report to find this information for each country individually (and some don't include this info yet) so it might be a long term project...!

fluxy/plots/flux_timeseries.py
mo-aliceramsden: Yes, good spot! I've corrected this

scripts/PARIS_inversion_results.ipynb
mo-aliceramsden: Good idea on using the standard run for all species here, that makes sense. I've added that in, so the user just needs to specify one model and all standard runs are read in.
I tried to set this function up to plot all species for multiple models (as different bars) but this was making the plot very messy and hard to format. So I think it best if we keep this plot as a simple single model + inventory for now, and I can update this in the future if we decide we need this.

fluxy/operators/flux_map_diff.py
```diff
@@ -66,6 +73,11 @@ def define_var_plot(
         else:
             if var_p in ds:
                 var_p_bis = var_p
+            elif var_p.replace('_inversion_grid','_out') in ds:
+                var_p_bis = var_p.replace("_inversion_grid", "_out")
+                logger.warning(
+                    f"'{var_p}' not found in dataset(s) (inversion system = {ds.attrs['inversion_system']}), replaced by '{var_p_bis}' for this model."
+                )
```
melodb: I think it would be better to rename the variable upfront when reading the data. If you add it to dictionary `legacy_names` in io.py, I think this will be done automatically in function "edit_vars_and_attributes".

fluxy/config.py
```diff
@@ -264,3 +273,10 @@ def get_default_sector_colors() -> list[str]:
     """
 
     return sector_color_palette
+
+def get_default_species_colors() -> list[str]:
```
melodb: As it stands, I think this function is not needed. As long as the import statement `from fluxy import config` exists on the top of your file, you can access the dictionary with `config.species_color_palette`. That's what we do for the remaining color dictionaries.

fluxy/operators/flux_align_dataset.py
```diff
@@ -30,29 +33,56 @@ def align_time(ds_list: list[xr.Dataset],
...
+        logger.warning(
+            "Datasets have only one time value — aligning them with the time "
+            "coordinate of the first dataset in the list "
+            "without checking time difference."
+        )
+        
 
     # Reduce datasets to their overlapping time range (only if period can be inferred)
-    if period is not None and only_overlapping:
+    if period is not None and only_overlapping == True:
```
melodb: I think `== True` is not needed.

fluxy/operators/flux_align_dataset.py
```diff
@@ -30,29 +33,56 @@ def align_time(ds_list: list[xr.Dataset],
...
         ds_list = [ds.sel(time=slice(min_date, max_date)) for ds in ds_list]
-        
+
     else:
-        all_min = [x.time.min() for x in ds_list]
-        min_date_loc = [i for i,date in enumerate(all_min) if date == min(all_min)][0]
-        ds_list = [ds.reindex(indexers={'time':ds_list[min_date_loc]['time']},fill_value=np.nan,method=None) for ds in ds_list]
+        time_warning = ["only_overlapping is set to False so all following data is included:"]
+                
+        min_date = min([x.time.min() for x in ds_list]) - period / 2
```
melodb: As far as I understand, regardless of the values of `only_overlapping`, if `period == None`, we will enter this else statement. However, in this case, variable `period` will be equal to `None` and there will be an error when computing `min_date`.  Should there be two nested if-statements? For example:
```
if period is not None:    
     if only_overlapping:
          (...)
     else:
          (...)
```
I worked a bit on this part of the code and made a proposition. It also fixes other small problems. Check PR https://github.com/openghg/fluxy/pull/277

fluxy/operators/flux_align_dataset.py
```diff
@@ -120,7 +150,9 @@ def align_map_data(
     Args:
         ds_all (dict[xr.Dataset | xr.DataArray]):
             Dictionary of model names and corresponding xarray Datasets/DataArrays.
-
+        only_overlapping (bool):
+            If True, reduces datasets to their overlapping time range before aligning.
+            If False, includes all data and fills in missing time steps with NaNs.
```
melodb, line 189: `only_overlapping` is in the arguments list, but it's not an input to the function. Moreover, it is not used in the call to align_time. I have tested adding it and the code works fine with both `only_overlapping = True/False` with the changes I made in PR #277. If my PR gets merged in, I would then suggest adding `only_overlapping` as input to "align_map_data" with default equal to `True`, and as input paramenter to the call to "align_time".

fluxy/plots/flux_timeseries.py
```diff
@@ -271,7 +269,7 @@ def prepare_data_to_plot(
 
         for group_label, model_list in combined_models_dict.items():
                 combine_mask = [model in model_list for model in ds_to_combine.keys()]
-                ds_combined = combine_dataset(ds_to_combine, combine_mask, only_overlapping)
+                ds_combined = combine_dataset(ds_to_combine, combine_mask)
```
melodb: If we merge PR #277, you can actually use both options of `only_overlapping` when setting `plot_combined=True`, so we can actually keep it as input. The same is true for the spatial maps.

scripts/PARIS_inversion_results.ipynb
```diff
@@ -1022,9 +1022,7 @@
     "        ds_all_flux[species] = read_model_output(data_dir,\"flux\",species,[models[s]],config_data,period=period[s],read_standard_run=read_standard_run)\n",
     "        ds_all_flux_scaled[species] = slice_flux(ds_all_flux[species],config_data,start_date,end_date,species=species,country_flux_units_print=country_flux_units_print)\n",
     "        \n",
-    "#not used in this plot, but neede to process data without errors\n",
-    "model_colors = set_model_colors(models)\n",
-    "model_labels = set_model_labels(models,config_data,get_labels_from_file)"
+    "#not used in this plot, but neede to process data without errors"
```
melodb: Typo: Delete this comment

fluxy/plots/flux_timeseries.py
```diff
@@ -1435,6 +1434,14 @@ def plot_all_species_stacked_bar(all_species: list[str],
...
 
     width = np.timedelta64(150,'D')
+    
+    unit = []
+    for s in all_species:
+        unit.append(get_unit(ds_all_flux_scaled[s]))
+    if all(x == unit[0] for x in unit):
+        country_flux_units_print = unit[0]
+    else:
+        logger.error('Units for all species\' datasets are not equal.')
```
melodb: In this case, shouldn't we raise an error and stop the code instead of simply printing an error message?

scripts/PARIS_inversion_results.ipynb
melodb: Maybe set `read_standard_run = True` in the notebook.

scripts/PARIS_inversion_results.ipynb
```diff
@@ -992,12 +992,7 @@
     "### edit variables in this block\n",
     "all_species = ['nf3','sf6','all_pfc','all_hfc','n2o','ch4']\n",
     "period = ['yearly','yearly','yearly','yearly','monthly','monthly'] #inversion period, must be a string or a list of the same length as models, e.g. ['monthly','yearly']\n",
-    "models = ['InTEM_NAME_EUROPE_FLAT_PARISNID2026',\n",
-    "          'InTEM_NAME_EUROPE_FLAT_PARISNID2026',\n",
-    "          'InTEM_NAME_EUROPE_std',\n",
-    "          'InTEM_NAME_EUROPE_std',\n",
-    "          'InTEM_NAME_EUROPE_EDGAR_PARISNID2026',\n",
-    "            'InTEM_NAME_EUROPE_EDGAR_PARISNID2026']\n",
+    "model = ['InTEM']\n",
```
melodb: Maybe keep the variable name "models" for consistency with the remaining functions. I know that the code does not work with more than one for now, but we can try to extend it in the future. Inside the plotting function, you can check the length of `models` and if it has more than one you can print a warning saying that only the first model will be plotted.

scripts/PARIS_inversion_results.ipynb
```diff
@@ -1014,11 +1009,11 @@
     "\n",
     "for s,species in enumerate(all_species):\n",
     "    if 'all' in species:\n",
-    "        ds_all_flux_scaled[species] = read_flux_total_fgases(data_dir,species,[models[s]],config_data,\n",
+    "        ds_all_flux_scaled[species] = read_flux_total_fgases(data_dir,species,[model],config_data,\n",
```
melodb: Why does the variable `model` need to be inside a list? It is already a list. In the other usage of "read_flux_total_fgases" in the notebook, we don't do this.

configs/models_info.json
```diff
@@ -46,21 +47,21 @@
...
-                    "hfc125": "NAME_EUROPE_EDGAR",
-                    "hfc152a": "NAME_EUROPE_FLAT",
-                    "hfc32": "NAME_EUROPE_EDGAR",
-                    "hfc227ea": "NAME_EUROPE_FLAT",
-                    "hfc23": "NAME_EUROPE_FLAT",
-                    "hfc245fa": "NAME_EUROPE_FLAT",
-                    "hfc365mfc": "NAME_EUROPE_FLAT",
-                    "hfc4310mee": "NAME_EUROPE_FLAT",
-                    "hfc236fa": "NAME_EUROPE_FLAT",
+                    "hfc134a": "NAME_EUROPE_EDGAR_std",
```
melodb: Taking into account the new label "paris", you might want to change "_std" to "_paris" in the default file names.

fluxy/operators/regions.py
melodb: I think it's ok for now. If we run into problems we can later fix it.

fluxy/operators/regions.py
```diff
@@ -283,8 +285,7 @@ def extract_region_inventory_flux(
         inventory_year: year of inventory to get.
         inventory_filename: Name of inventory file: {inventory_filename}_{species}_{inventory_year}
     Returns:
-        dataset with country selected
-
+                dataset (and uncertainty dataset, if this is available) with country selected.
```
melodb: Strange indentation?

fluxy/plots/utils.py
```diff
@@ -31,15 +32,15 @@ def update_list_params(params_to_check: list, expected_size: int) -> list:
...
+    for p,param in enumerate(params_to_check):
         if param is None:
             updated_params.append([False] * expected_size)
         elif type(param) is list:
             if len(param) == expected_size:
                 updated_params.append(param)
             else:
                 raise ValueError(
-                    f"{param} must be a boolean or a list of booleans of the same length as models."
+                    f"{params_to_check_names[p]}:{param} must be a boolean or a list of booleans of the same length as models or inventory_years."
```
melodb: I think this error message might be misleading. Maybe make an if-statement to define `ref_list` and then update the error message to:
`f"{params_to_check_names[p]}:{param} must be a boolean or a list of booleans of the same length as {ref_list}."`

scripts/PARIS_inversion_results.ipynb
```diff
@@ -978,7 +1064,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.10.6"
+   "version": "3.1.undefined"
```
melodb: Revert version changes to avoid merge conflicts.

scripts/PARIS_inversion_results.ipynb
```diff
@@ -960,6 +969,83 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from fluxy.io import read_model_output, read_flux_total_fgases\n",
+    "from fluxy.operators.select import slice_flux\n",
+    "from fluxy.config import set_model_colors,set_model_labels\n",
```
melodb: remove import of unused functions:
`from fluxy.config import set_model_colors,set_model_labels`

scripts/PARIS_inversion_results.ipynb
```diff
@@ -960,6 +969,83 @@
...
+    "                                      model_colors=model_colors,\n",
+    "                                      model_labels=model_labels,\n",
+    "                                      start_date=start_date,\n",
+    "                                      end_date=end_date,\n",
+    "                                      plot_inventory_uncertainty=plot_inventory_uncertainty,\n",
+    "                                      inventory_years=inventory_years,\n",
+    "                                      inventory_filename=inventory_filename,\n",
+    "                                      data_dir=data_dir,\n",
+    "                                      y_lim=y_lim)"
+   ]
```
melodb, line 1055: I would add a Markdown before this cell with a subtitle just for consistentcy with the rest of the notebook. Moreover, I would define `y_lim=None` and add the comment `### edit variables in this block` as in the other notebook cells.

configs/models_info.json
```diff
@@ -46,21 +47,21 @@
...
+                    "hfc245fa": "NAME_EUROPE_FLAT_std",
+                    "hfc365mfc": "NAME_EUROPE_FLAT_std",
+                    "hfc4310mee": "NAME_EUROPE_FLAT_std",
+                    "hfc236fa": "NAME_EUROPE_FLAT_std",
                     "pfc218": "NAME_EUROPE_FLAT_std",
                     "pfc318": "NAME_EUROPE_FLAT_std",
-                    "pfc116": "NAME_EUROPE_EDGARMINVAL_std",
-                    "cf4": "NAME_EUROPE_EDGARMINVAL_std",
+                    "pfc116": "NAME_EUROPE_FLAT_std",
+                    "cf4": "NAME_EUROPE_FLAT_std",
```
melodb: Add "nf3" to the list

fluxy/plots/flux_timeseries.py
```diff
@@ -1071,14 +1104,21 @@ def plot_country_flux(
...
                     [plotted_data_df, prior_df], ignore_index=True
                 )
 
         # plot inventory
         if plot_inventory:
+            
+            plot_inventory_uncertainty, inventory_years, = update_list_params(
+                [plot_inventory_uncertainty, inventory_years],
+                ['plot_inventory_uncertainty','inventory_years'],
+                expected_size=len(inventory_years),
```
melodb: The code breaks here when I set:
```
plot_inventory = True
plot_inventory_uncertainty = [False]
inventory_years = None
```
Error message:
TypeError: object of type 'NoneType' has no len()

I think somewhere in the code the case `inventory_years = None` is translated to the most recent year, but maybe it happens after this line.

melodb (commented):
Thanks a lot for all the work! I did some tests and left some comments. They are in general minor. Regarding function "align_time", I have implemented a suggestion in PR #277. Check what you think.

fluxy/plots/flux_timeseries.py
mo-aliceramsden: Yes, I think at some point we lost the functionality for using the most recent inventory year. I've added this back in for now. There may be a cleverer way of doing this, but it should work for now!

fluxy/operators/flux_map_diff.py
mo-aliceramsden: I tried your suggestion out: removing this section of code and passing this variable renaming to `legacy_names` seemed to work for the plotting but made the tests that use the `old_format` files to fail. So I've added this back in for now. I'll make a note to check on this again in the future when I've hopefully got a bit more time!

melodb: I just checked the error message and it seems to be related with variable 'percentile_flux_total_prior_out'. The code removes 'percentile' and tries to look for variable 'flux_total_prior_out'. Does the tests pass if you also add 'percentile_flux_total_prior_out' and 'percentile_flux_total_posterior_out' to legacy_names?

fluxy/plots/flux_timeseries.py
melodb: That functionary still exists but it's only used inside function "extract_region_inventory_flux" in "regions.py" when actually reading the inventory files. If you test devel, you will see that it is working nicely.

For this PR, I think there is no need to implement it again in the middle of "plot_country_flux". As far as I can see, variable `plot_inventory_uncertainty` is only really needed in "add_inventory_barplot" after the call to "retrieve_inventories". `plot_inventory_uncertainty` and `inventory_years` are already input variables to "add_inventory_barplot" so maybe you can make this check only inside "add_inventory_barplot" after the call to "retrieve_inventories". Maybe you can use `len(inventories_uncert_to_plot)` as your expected length?

melodb:
I took the freedom to merged my PR just to speed things up. We are almost there! I think only 3 things are missing:
- Unresolved comment 1: check if adding 'percentile_flux_total_prior_out' and 'percentile_flux_total_posterior_out' to legacy_names solves the problem.
- Unresolved comment 2: check if moving the checker for plot_inventory_uncertainty after "retrieve_inventories" solves the problem.
- Could you check if your test case of all_pfc works by adding only period/2 to max date?

Thanks a lot for all the work!

fluxy/operators/flux_map_diff.py
mo-aliceramsden: Yes, you're right, the tests work with that fix too, thanks!

fluxy/plots/flux_timeseries.py
mo-aliceramsden: Good idea, I've moved the check for list length into `add_inventory_barplot`, and based this check off the length of `inventory_years`, to reduce confusion if the user doesn't select an inventory year and incorrectly sets `inventories_uncert_to_plot` (as with the option to include a list of True/False for `inventories_uncert_to_plot`, that's probably the easiest intput to mis-interpret)

mo-aliceramsden:
> I took the freedom to merged my PR just to speed things up. We are almost there! I think only 3 things are missing:
> 
>     * Unresolved comment 1: check if adding 'percentile_flux_total_prior_out' and 'percentile_flux_total_posterior_out' to legacy_names solves the problem.
> 
>     * Unresolved comment 2: check if moving the checker for plot_inventory_uncertainty after "retrieve_inventories" solves the problem.
> 
>     * Could you check if your test case of all_pfc works by adding only period/2 to max date?
> 
> 
> Thanks a lot for all the work!

Thanks for this summary, that's been very helpful! I think I've address all those points:

* Adding the _percentile variable renaming to legancy_names works well.
* Moving the error checking for len plot_inventory_uncertainty is working well too.
* Adding period/2 to the max date for align_time still allows for all_pfc to be combined correctly, including the final year available this time.

Thank you for those suggestions!!

melodb:
I took the freedom to add some final corrections. I felt it was easier this way. Everything worked fine for me with all combinations of plot_inventory = True/False, plot_inventory_uncertainty = None/True/False/list, inventory_years=None/list. However I did not update my local inventory folder and I cannot really see if the error bars would show up.

If you agree with the changes and all works fine for you, I would be happy for this to be merged. I would only suggest that you remove the NID2026 notebook if that's ok for you. Moreover, if you could apply black formatter to all files you changed that would be great.

melodb:
**Note:** I would suggest applying black formatter to all files you changed except "flux_timeseries.py". This will minizime conflicts with PR #257 afterwards.

mo-aliceramsden:
Thanks for those updates, I've applied the black formatter to all the changed files (apart from flux_timeseries.py as you recommend). I'm happy for this to be merged in now, the inventory uncertainty plotting is working as expect from my end. Thank you for your work on this!

melodb (approved):
Thanks a lot for all the work and patience! I think we can confidently merge this in devel. I will let you do the honors. I would go for "Squash and merge" to simplify devel's history. Moreover, after so many commits, you might want to summarize the merge commit message :p

fluxy/operators/regions.py
```diff
@@ -113,6 +113,20 @@ def _extract_region_flux_sector(
...
+                (
+                    ["posterior", "prior"]
+                    if f"flux_{sector}_prior_country" in ds
+                    else ["posterior"]
+                )
+                for m, ds in ds_all.items()
+            ],
+        )
+    )
+
```
melodb: I now see that this variable is not used anywhere. Just in case you want to delete it before merging.

PR #257, opened 2025-12-15
Add mf timeseries for multiple sites

alexdanjou:
I allows to plot multiple sites on the same graph. It is I think interesting to look at and compare the mf seasonnal cycle between some choosen sites.

See example (not choosen randomly 😅):
image


melodb:
Didn't @hdelongueville work on the same thing?

hdelongueville:
I think here it is on the same plot while "mine" is on several subplot.

hdelongueville:
I'll look at it when I'm creating a PR for mine :)

hdelongueville:
I had a quick look at this PR and it seems that it is not "just adding a new mf function" but is also changing some of flux_timeseries.. which suggests that it is not a straightforward PR. Because I don't have a lot of time, I'll let you decide what you want to do with it, but looks like a nice feature to have in Fluxy!

mo-aliceramsden:
It might be worth having a quick chat about this at the next PARIS meeting (when we have time) to decide what format we want these plots in, then we can decide which of the two PRs on this to merge in to develop?

melodb (commented):
I took a quick look at the code and it seems that there is a new function `add_line_plot` which is now used to plot both the mf timeseries and the flux timeseries. We might want to check the pros and cons of this option. If we keep it, we might want to move this function to other location (maybe mf_timeseries.py is not the best one if it is a generic function). The approach followed here seems a bit different than the approach followed in #266. We might want to uniformize the code.

I tested the notebook and all plots are working fine. I have added a cell with a call to the new function `plot_sites_list_mf` to facilitate testing. After my commit, the tests are failling. Honestly, I don't know why they were passing before and are failling now. I have simply added a cell in the notebook.

moPeterAndrews:
Once I updated all the scripts it worked without issue, so not sure what's happening with yours Daniela.

Comparing to #266, this is probably better for comparing sites, as sites are on the same graph, and each graph is for a model, whereas #266 is a different axis for each site. Not sure which approach is better, though this seems to be a bit faster, and the seasonal cycle plotting works for this version where it doesn't on #266

melodb:
> Once I updated all the scripts it worked without issue, so not sure what's happening with yours Daniela.
> 
> Comparing to #266, this is probably better for comparing sites, as sites are on the same graph, and each graph is for a model, whereas #266 is a different axis for each site. Not sure which approach is better, though this seems to be a bit faster, and the seasonal cycle plotting works for this version where it doesn't on #266

Thanks, Peter, for checking this! There is no problem with my version. It's just the GitHub tests which are failling (see message "All checks have failed" close to the red cross).

Regarding the comparison with #266, I think both plots are usefull: 1) one subplot per model with all sites vs 2) one subplot per site with all models. It depends on the analysis you are performing. We can try to move along with this PR, but add option 2 as well.

melodb:
@moPeterAndrews: I think it fails simply because devel is failling. The error message points to:

E     File "/home/runner/work/fluxy/fluxy/fluxy/operators/flux_map_resample.py", line 274
E       f"{first_time.strftime([[1:3], [7:9]])}"
E                             ^
E   SyntaxError: f-string: expecting '=', or '!', or ':', or '}'

which was introduced 3 merges ago. This will be fixed soon.

moPeterAndrews:
I just added the ability to switch between plotting sites and models on separate graphs - the only change is in plot_sites_list_mf, it just switches which order it loops over sites and models.

fluxy/plots/mf_timeseries.py
lionel42: Below  a new option for plot type is introduced. 

If we continue adding plot_types, it would be better to use an Enum instead of literal to be able to track correctly all those options.

fluxy/plots/mf_timeseries.py
```diff
@@ -475,6 +478,84 @@ def add_xlims_and_ticks(
     ax.set_xlim(xlim)
 
 
+def add_line_plot(
+    ax: Axes, ds: xr.Dataset, plot_type: str, add_unc: bool = True
```
lionel42: See comment for the enum replacement for plottype and apply here

fluxy/plots/mf_timeseries.py
```diff
@@ -475,6 +478,84 @@ def add_xlims_and_ticks(
...
+        ax.errorbar(
+            time_as_datetime,
+            y=ds.sel(percentile="mean"),
+            yerr=ds.sel(percentile="std"),
+            alpha=0.4,
+            fmt="none",
+            color=ds.attrs["plot_color"],
+        )
+        res["min_unc"] = ds.sel(percentile="mean").values - ds.sel(percentile="std")
+        res["max_unc"] = ds.sel(percentile="mean").values + ds.sel(percentile="std")
```
lionel42: remove duplicates by extracting mean and std as varialbes first

lionel42: We should externalize this functionality to another function. 

It would be nice to have something 

```
unc_min, unc_max = get_unc_minmax(ds)
```

fluxy/plots/mf_timeseries.py
```diff
@@ -836,7 +876,7 @@ def plot_sites_timeseries(
     return fig
 
 
-def plot_histogram(
+def add_histogram_plot(
```
lionel42: dangerous renaming, this function might be already used by others elsewhere

fluxy/plots/mf_timeseries.py
```diff
@@ -960,3 +1000,203 @@ def plot_histogram(
     ax.set_xlabel(legend_hist)
 
     return None
+
+def check_site_list(site_list: list | None, ds_all: dict[str, xr.Dataset])->list:
+    """
```
lionel42: Utitility function not specific to plots, should be moved in another place

fluxy/plots/mf_timeseries.py
```diff
@@ -960,3 +1000,203 @@ def plot_histogram(
...
+
+
+def plot_sites_list_mf(
+    ds_all,
+    sites,
+    species,
+    variable,
+    unc_variable,
+    model_labels: dict[str, dict],
+    aggreg_month = False,
```
lionel42: add type hints

fluxy/plots/mf_timeseries.py
```diff
@@ -960,3 +1000,203 @@ def plot_histogram(
...
+
+    plot_type="multiple_sites"
+    models = ds_all_p.keys()
+    species_info = config_data.get("species_info", {}).get(species, {})
+    plotted_data_df = pd.DataFrame()
+
+    # Look for sites
+    sites = check_site_list(sites, ds_all_p)
+    
+    if data_on_single_graph=="models":
```
lionel42: This if statement creates a lot of duplicated code. 

Basically in both cases we do the same (fig creation, loop over sites and models, add the line function)

A more elegant solution, where we have more if statements but localize them on each specific cases (fig creation, add_line_to_plot function)

fluxy/plots/mf_timeseries.py
```diff
@@ -960,3 +1000,203 @@ def plot_histogram(
...
+        ax[iaxes].set_axisbelow(True)
+    # min_mf = min(min_mf, ax[iax, 0].get_ylim()[0])
+    # max_mf = max(max_mf, ax[iax, 0].get_ylim()[1])
+
+    # if y_lim is None:
+    #     y_lim = [min_mf - 0.05 * (max_mf - min_mf), max_mf + 0.1 * (max_mf - min_mf)]
+
+    # # Set all the axes to the same y-axis limits
+    # for iax, ax0 in enumerate(ax[:, 0]):
+    #     ax0.set_ylim(y_lim)
```
lionel42: Delete old code instead of commenting

tests/test_operators.py
```diff
@@ -77,7 +77,7 @@ def test_stats_mf(stat):
 @pytest.mark.parametrize("model", test_models_with_inlet)
 def test_slice_height(model):
 
-    ds_all_mf__with_inlet_sliced = slice_site(ds_all_mf_with_inlet[model], site="TAC")
+    ds_all_mf__with_inlet_sliced = slice_site(ds_all_mf_with_inlet[model], sites="TAC")
```
lionel42: We should not change tests !!! This will break scripts that people already implemented

lionel42 (changes requested):
This is a very nice feature.
However the code still need some rework.

* format  new code with the black formatter
* adding tests for the new plot function (and if possible utility functions)

lionel42:
@moPeterAndrews are you interested in finishing this PR ? 

I have added a review and there is quite some work to do, but I am happy to help if needed

moPeterAndrews:
I'll work through these points, yes, no problem. Thanks Lionel

tests/test_operators.py
moPeterAndrews: Looks like Alex changed this because slice_site now has functionality for multiple sites, so continuing to use "site" as an argument doesn't make sense, and he changed it to "sites". So far as I can tell, this is the only place it's used as a keyword argument, and shouldn't break anything else.

lionel42: Okay I see, the real problem is about changing the `slice_site` function itself. 

Maybe you will have to go there and add the possibility to select just one site or many sites.

Something like this:

```
def slice_site(
    ds: xr.Dataset, site: str | None = None, sites: list[str] | None = None, raise_error: bool = True
) -> xr.Dataset | None:


    # Here logic to tranform site into sites based on what is given in the input 
``` 

Then you would add tests to check if the function behaves properly.


The principles behind this is that we should never break existing api without notice. It might not be used in the code, but some people might have used this function somewhere in their code.

fluxy/operators/select.py
```diff
@@ -225,29 +225,57 @@ def slice_mf(
...
-    mask = ds["number_of_identifier"] == site_index
-    ds = ds.where(mask, drop=True)
+        if site_index is not None:
+            mask += ds["number_of_identifier"] == site_index
+        else:
+            msg = f"Site {site} not found for model {ds.attrs['exp_name']}."
+            if raise_error:
+                raise ValueError(msg)
+            else:
+                logger.warning(msg)
```
lionel42: Implement a function warn or raise, since this is used in multiple places

fluxy/plots/mf_timeseries.py
```diff
@@ -19,13 +22,23 @@
...
     get_unique_site_height_pairs,
+    slice_site_dict_of_datasets,
+    check_site_list,
 )
 from fluxy.plots.utils import set_min_decimal_points
 
 logger = logging.getLogger(__name__)
 
 
+class plot_types(Enum):
```
lionel42: Use CamelCase for class names

fluxy/plots/mf_timeseries.py
```diff
@@ -960,3 +1039,149 @@ def plot_histogram(
...
+        axes_looper = models
+        # Create figure
+        ncols = int(np.sqrt(len(models)))
+        nrows = ceil(len(models) / ncols)
+
+    elif data_on_single_graph == "sites":
+        axes_looper = sites
+
+        ncols = int(np.sqrt(len(sites)))
+        nrows = ceil(len(sites) / ncols)
```
lionel42: We can easily extract the code to calculate ncols and nrows from the if elif statement

fluxy/plots/mf_timeseries.py
```diff
@@ -960,3 +1039,149 @@ def plot_histogram(
...
+def plot_sites_list_mf(
+    ds_all: dict[str, xr.Dataset],
+    sites: list[str],
+    species: str,
+    variable: str,
+    unc_variable: str | None,
+    model_labels: dict[str, dict],
+    aggreg_month: bool = False,
+    config_data: dict[str, dict] = {},
+    data_on_single_graph: str = "models",
```
lionel42: Use literal if you have only 2 possibilites

fluxy/plots/mf_timeseries.py
lionel42: Also have to handle the else case and raise error on invalid values

lionel42 (changes requested):
Looks good, we are almost done.

There are a lot of changes due to the black formatting included. I will open a separate issue on this for the next big release to ensure we only accept PRs that are formatted correctly and not have this issue in the future.

fluxy/operators/select.py
moPeterAndrews: As this is used only twice, I'm not sure that this is really necessary?

lionel42: Not 100% necessary, you can keep it like this

fluxy/operators/select.py
```diff
@@ -225,29 +225,57 @@ def slice_mf(
     return ds_all
 
 
-def slice_site(ds: xr.Dataset, site: str) -> xr.Dataset:
+def slice_site(
+    ds: xr.Dataset,
+    sites: str | list[str]=[None],
+    site: str | None = None,
```
melodb: In my view, we could keep only the input variable `site` and accept `site: str | list[str]`. This was the approach followed in function "slice_mf", which is still in place. Is there any advantage in adding `sites` besides being gramatically correct? If not, I would recommend simplifying here to `site` only. If you prefer to keep both, I would add `site` back to the function description (maybe with the note "kept for compatibility with previous versions"?).

In addition, I would remove the default values for `sites` and `site`. That is, remove `=[None]` and `=None`. The code does not seem ready to handle those situations.

fluxy/operators/select.py
```diff
@@ -225,29 +225,57 @@ def slice_mf(
...
         ds (xarray dataset):
-            Dataset with mf data of a given model, sliced to only include data for the given site.
+            Dataset with mf data of a given model, sliced to only include data for the given site(s).
     """
+    if isinstance(site,str):
+        sites = [site]
+    if not isinstance(sites, list):
+        sites = [
+            sites,
+        ]
```
melodb: Is there a reason to check for `not isinstance(..., list)` instead of `isinstance(..., str)`? As far as I understand, it will not work properly if sites is a set or a tuple. Moreover, as the comma is not needed, I think it would look cleaner without it. For example:
```
if isinstance(sites, str):
        sites = [sites]
```
Anyway, as I mentioned above, I think we can suppress `sites` and use only `site`.

fluxy/operators/select.py
```diff
@@ -225,29 +225,57 @@ def slice_mf(
     return ds_all
 
 
-def slice_site(ds: xr.Dataset, site: str) -> xr.Dataset:
+def slice_site(
+    ds: xr.Dataset,
+    sites: str | list[str]=[None],
+    site: str | None = None,
+    raise_error: bool = True,
```
melodb: Change default of `raise_error` to `False`. In my view, warnings are enough for a recurrent usage of the function. I don't see the need for the code to break if only one site is missing.

fluxy/operators/select.py
```diff
@@ -225,29 +225,57 @@ def slice_mf(
...
+        site_indices.append(site_index)
 
-    mask = ds["number_of_identifier"] == site_index
-    ds = ds.where(mask, drop=True)
+        if site_index is not None:
+            mask += ds["number_of_identifier"] == site_index
+        else:
+            msg = f"Site {site} not found for model {ds.attrs['exp_name']}."
+            if raise_error:
+                raise ValueError(msg)
```
melodb: Currently, if `raise_error == True`, an error is raised if one single site is missing. In my view, it is preferable not to raise an error here, but only in L.276 if none of the sites listed exist in the dataset. If you look at the call to "slice_site" in "slice_mf", you can see that it is inside a try/except block and that the whole dataset is removed if a `ValueError` is catched. I don't see an advantage in this behaviour as different datasets use different sites and we might want to plot all existing ones. With the current implementation, we are limited to plot only the common sites. Therefore, I would suggest replacing L.263-268 by:

```
else:
    logger.warning(f"Site {site} not found for model {ds.attrs['exp_name']}.")
```

fluxy/operators/select.py
```diff
@@ -225,29 +225,57 @@ def slice_mf(
...
-    return ds
+    if mask.any():
+        ds = ds.where(mask, drop=True)
+        return ds
+    else:
+        msg = f"No data for any sites {sites} with indices {site_indices} in model {ds.attrs['exp_name']}."
+    if raise_error:
+        raise ValueError(msg)
+    else:
+        logger.warning(msg)
```
melodb: The definition of `msg` can get out of the `else`. Moreover, if `raise_error == False`, should we `return None` after printing the warning? I guess that's what will happen, but maybe it's cleaner to actually write it.

fluxy/operators/select.py
```diff
@@ -94,7 +94,7 @@ def slice_mf(
     ds_all: dict[str, xr.Dataset],
     start_date: str = None,
     end_date: str = None,
-    site: str = None,
+    site: str | list[str] = None,
```
melodb: Add `| None` to variable type.

fluxy/operators/select.py
```diff
@@ -293,11 +321,13 @@ def get_site_index(ds: xr.Dataset, site: str) -> int | None:
...
     """
-
+    
     if site in ds["platform"]:
         index = np.where(ds["platform"] == site)[0][0]
         return index
-
+    elif "sitenames" in ds.keys() and site in ds["sitenames"]:
+        index = np.where(ds["sitenames"] == site)[0][0]
+        return index
```
melodb: I don't think this is needed. When reading the dataset `sitenames` should be renamed to `platform` (see `legacy_names` in "io.py").

fluxy/plots/flux_timeseries.py
```diff
@@ -305,110 +305,72 @@ def prepare_data_to_plot(
...
-def add_posterior_plot(
-    ax: Axes, ds_region: xr.Dataset, highlighted_line: bool, add_post_unc: bool
+def add_line_plot(
+    ax: Axes, ds: xr.Dataset, variable: Literal["posterior","prior"], highlighted_line: bool = False, add_unc: bool = False
 ) -> dict[str, dict]:
     """
-    Plot the posterior data on the axis. The variable posterior of the dataset ds_region is plotted as a line (color and label found in the dataset
-    attributes) and the uncertainty (variables lower_posterior, upper_posterior in the dataset) is plotted as a semi-transparent filled space.
-    Axes:
+    Plot the posterior/prior data on the axis. The variable posterior/prior of the dataset ds_region is plotted as a line (color and label found in the dataset
```
melodb: Remove "ds_region" from the function description.

fluxy/plots/mf_timeseries.py
```diff
@@ -475,6 +490,113 @@ def add_xlims_and_ticks(
     ax.set_xlim(xlim)
 
 
+def get_minmax_unc(
+    ds: xr.Dataset, time_as_datetime: list, ax
```
melodb: Add type to `ax` and to elements of `time_as_datetime`. Moreover, I find the order a bit strange. In all other plotting functions, I think `ax` is the first input followed by `ds`. Maybe also compute `time_as_datetime` inside this function instead of having this variable as input. This will make this function more independent from the calling function which I think is nice.

fluxy/plots/mf_timeseries.py
```diff
@@ -475,6 +490,113 @@ def add_xlims_and_ticks(
     ax.set_xlim(xlim)
 
 
+def get_minmax_unc(
```
melodb: I think the name of this function is a bit misleading. It does not only get the min/max unc but actually plots the uncertainty bands. What about "add_unc_plot"? It would be in line with function "add_line_plot", which also  returns the data actually plotted.

fluxy/plots/mf_timeseries.py
```diff
@@ -475,6 +490,113 @@ def add_xlims_and_ticks(
...
+        std_values = ds.sel(percentile="std").values
+        min_unc = mean_values - std_values
+        max_unc = mean_values + std_values
+    else:
+        min_unc = None
+        max_unc = None
+    return min_unc, max_unc
+
+
+def add_line_plot(
```
melodb: This function has the same name as other one implemented in "fluxy/plots/flux_timeseries.py" in this PR. I find that a bit confusing. Moreover, it does not only plot lines, but also scatter plots. What about "add_mf_plot" or "add_mf_line_scatter_plot"?

fluxy/plots/mf_timeseries.py
```diff
@@ -475,6 +490,113 @@ def add_xlims_and_ticks(
...
+    ax: Axes, ds: xr.Dataset, plot_type: str, add_unc: bool = True
+) -> pd.DataFrame:
+    """
+    Add line plot with uncertainty if requested.
```
melodb: I would rephrase to "Add line/scatter plot (...)"

fluxy/plots/mf_timeseries.py
```diff
@@ -475,6 +490,113 @@ def add_xlims_and_ticks(
...
+    Args:
+        ax: axes on which to plot
+        ds: dataset containing posterior/prior data
```
melodb: Rephrase to "dataset containing mole fraction data to plot"

fluxy/plots/mf_timeseries.py
```diff
@@ -475,6 +490,113 @@ def add_xlims_and_ticks(
...
+    }
+
+    if plot_type == "multiple_sites":
+        type_plot = ds.attrs["plot_label"]
+    else:
+        type_plot = ds.name
+
+    if ds.name in ["mf_observed", "observed_above_BC"] or plot_type == "diff":
+        plot_func = ax.scatter
+        add_kwargs = {"s": 8, "marker": "s"}
```
melodb: I would suggest updating the kwargs dictionary instead of creating a new one: `kwargs.update({"s": 8, "marker": "s"})`. In this way, `**add_kwargs` can be removed as input to `plot_func`.

fluxy/plots/mf_timeseries.py
```diff
@@ -569,59 +695,12 @@ def plot_timeseries(
...
-                    fmt="none",
-                    color=ds_plot.attrs["plot_color"],
-                )
+            res = add_line_plot(ax[iax, 0], data_to_plot[m][var], plot_type)
+            plotted_data_df = pd.concat([plotted_data_df, res], ignore_index=True)
 
         # Plot histogram
         if ax.shape[1] == 2:
-            plot_histogram(
+            multi_histogram_plot(
```
melodb: Is there any advantage in renaming this function?

fluxy/plots/mf_timeseries.py
```diff
@@ -960,3 +1039,149 @@ def plot_histogram(
...
+    species: str,
+    variable: str,
+    unc_variable: str | None,
+    model_labels: dict[str, dict],
+    aggreg_month: bool = False,
+    config_data: dict[str, dict] = {},
+    data_on_single_graph: str = "models",
+):
+    """
+    Plot timeseries of multiple sites on the same subplots. One subplot corresponds to one model.
```
melodb: Update function description so that it is consistent with both options of `data_on_single_graph`.

fluxy/plots/mf_timeseries.py
```diff
@@ -960,3 +1039,149 @@ def plot_histogram(
...
+    species_info = config_data.get("species_info", {}).get(species, {})
+    plotted_data_df = pd.DataFrame()
+
+    # Look for sites
+    sites = check_site_list(sites, ds_all_p)
+
+    if data_on_single_graph == "models":
+        axes_looper = models
+        # Create figure
+        ncols = int(np.sqrt(len(models)))
```
melodb: I can see that setting `ncols != 1` helps saving some space. However, I would prefer keeping the plot style equal to "plot_timeseries". Moreover, I think it would be nice to be able to reproduce Helene's plot when `data_on_single_graph == "sites"`. That would require the following changes:

**Implementation:**
- add possibility to add multiple variables (the current implementation only accepts one variable with the respective uncertainty; maybe the `include` dictionary can be used instead as for function "plot_timeseries")

**Plotting style:**
- `ncols = 1`
- shared y-label with only the species name
- site name on the upper left corner
- add legend with variable names only on top subplot.

The improvements in the style of the plot can be left for another PR if you prefer, but I think the possibility to plot multiple variables should be implemented already now.

fluxy/plots/mf_timeseries.py
```diff
@@ -960,3 +1039,149 @@ def plot_histogram(
...
+    for isite, site in enumerate(sites):
+        for im, m in enumerate(models):
+            # Select site
+            if data_on_single_graph == "models":
+                attrs_update = {"plot_label": site, "plot_color": f"C{isite:02d}"}
+                ax_index = im
+            elif data_on_single_graph == "sites":
+                attrs_update = {
+                    "plot_label": model_labels[m],
+                    "plot_color": f"C{im:02d}",
```
melodb: When `data_on_single_graph == "sites"`, I think we should use the colors specified in `model_colors`. This variable should be added as input to the function. Moreover, the same approach followed in "plot_timeseries" should be used here: use of dictionary `mf_color_index` in "config.py" and use of black color when plotting `mf_observed` together with other variables.

fluxy/plots/mf_timeseries.py
```diff
@@ -960,3 +1039,149 @@ def plot_histogram(
...
+            res = add_line_plot(
+                ax[ax_index],
+                data_to_plot[m][variable],
+                plot_type=plot_type,
+                add_unc=unc_variable,
+            )
+            plotted_data_df = pd.concat([plotted_data_df, res], ignore_index=True)
+
+    for iaxes, label in enumerate(axes_looper):
+        ax[iaxes].set_title(label)
```
melodb, line 1228: When `data_on_single_graph == "models"`, the title should be equal to `model_labels` and not to `models`.

fluxy/plots/mf_timeseries.py
```diff
@@ -960,3 +1039,149 @@ def plot_histogram(
...
+                    f"({unit})",
+                ]
+            )
+        )
+
+        handles, labels = fig.axes[-1].get_legend_handles_labels()
+        fig.legend(
+            handles,
+            labels,
+            loc="upper center",
```
melodb, line 1244: Improve legend location so that it's not on top of first subplot title.

melodb (commented):
Thanks a lot Peter for following up on this PR! I have tested it and left some comments. If you need any help from my side for the fixes just let me know.

fluxy/operators/select.py
lionel42: I agree with this, much clearer and simpler to have a single variable

fluxy/plots/mf_timeseries.py
lionel42: This is indeed something confusing.

It mixes 2 capabilities. 
* getting the uncertainty to use (minmax)
* plotting either errorbars or fillbetween 

It would be nice to separate these 2 functionalities. 

I would keep the name `get_minmax_unc`  to simply do the getting the uncertainty part. 
Choosing between errobars or fillbetween can be done in the main plotting function.

I am also not sure why having percentiles or mean/std should change the way the data is plot, both types should be possible.

This would solve the problem of having ax as an argument of the function as above.

fluxy/plots/mf_timeseries.py
moPeterAndrews: I think this is a mistake from a different branch I was working on, I've now removed it.

fluxy/operators/select.py
moPeterAndrews: I think I've resolved this reasonably well, but happy to go over it again if needs be!

melodb: What I meant here is that you can delete "else" and move the code below one identation level up. That is:

```
if mask.any():
     ds = ds.where(mask, drop=True)
     return ds

msg = f"No data for any sites {sites} with indices {site_indices} in model {ds.attrs['exp_name']}."
if raise_error:
     raise ValueError(msg)
else:
     logger.warning(msg)
     return None
```
Just a detail though.

fluxy/plots/mf_timeseries.py
```diff
@@ -1092,60 +1118,77 @@ def plot_sites_list_mf(
         raise ValueError("data_on_single_graph should be 'models' or 'sites'")
 
     ncols = int(np.sqrt(len(axes_looper)))
+    ncols=1
     nrows = ceil(len(axes_looper) / ncols)
```
melodb: If `ncols = 1`, I guess the calculation of `nrows` can be simplified to: `nrows = len(axes_looper)`

fluxy/plots/mf_timeseries.py
```diff
@@ -1045,15 +1069,16 @@ def plot_sites_list_mf(
     ds_all: dict[str, xr.Dataset],
     sites: list[str],
     species: str,
-    variable: str,
-    unc_variable: str | None,
+    include: VariableType,
     model_labels: dict[str, dict],
+    model_colors: dict[str, list],
```
melodb: Set default of `model_labels` and `model_colors` to empty dictionary `{}` so that the code works without them.

fluxy/plots/mf_timeseries.py
```diff
@@ -1092,60 +1118,77 @@ def plot_sites_list_mf(
...
         sharey="row",
         sharex="col",
         squeeze=False,
     )
-
+    
     ax = ax.flatten()
-
+    if model_colors is None:
+        model_colors = config.set_model_colors(ds_dict.keys())
```
melodb: Up to now, we have been calling "config.set_model_colors()" directly in the notebook after reading the data to make sure that colors are consistent across all plots regardless of the slicing. For consistency, I would suggest not calling "config.set_model_colors()" here. If you set the default of `model_colors` as an empty dictionary, you can check if `m` exists in `model_colors`. If it does not exist, you can set `model_colors[m] = config.get_default_colors()`. This is the approach followed in function "prepare_data_to_plot".

fluxy/plots/mf_timeseries.py
```diff
@@ -1092,60 +1118,77 @@ def plot_sites_list_mf(
...
-            plotted_data_df = pd.concat([plotted_data_df, res], ignore_index=True)
+            for variable in include.keys():
+                if variable in ["mf_observed", "observed_above_BC"] and len(include.keys()) > 1:
+                    attrs_update = attrs[data_on_single_graph].copy()
+                    attrs_update["plot_color"] = "black"
+                    marker=obs_markers[marker_index]
+                elif data_on_single_graph == "sites":
+                    attrs_update = attrs[data_on_single_graph].copy()
+                else:
+                    attrs_update = attrs[data_on_single_graph].copy()
```
melodb: The `elif` and the `else` are equal so I guess the `elif` is redundant?

fluxy/plots/mf_timeseries.py
```diff
@@ -1092,60 +1118,77 @@ def plot_sites_list_mf(
...
 
     for iaxes, label in enumerate(axes_looper):
+        if axes_looper == models:
+            label=model_labels[label]
         ax[iaxes].set_title(label)
         ax[iaxes].set_ylabel(
             " ".join(
                 [
-                    config.mf_labels.get(variable, variable),
+                    #config.mf_labels.get(variable, variable),
```
melodb: Remove line instead of commenting.

scripts/PARIS_inversion_results.ipynb
```diff
@@ -341,9 +341,20 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 34,
    "metadata": {},
-   "outputs": [],
+   "outputs": [
+    {
+     "data": {
```
melodb: There are a lot of changes in the notebook that should not be commited. Make sure to clean all output cells and to revert all changes in python version, execution_count, etc. Keep only the changes in the notebook code.

melodb (commented):
Thanks a lot Peter once again! I left some final comments: just small details. I have also "unresolved" one of my previously comments and left a reply. When testing the notebook, all worked fine except one thing: when setting `include` to a single variable, the code breaks because `marker` is not defined.

scripts/PARIS_inversion_results.ipynb
moPeterAndrews: I think I've covered all the changes, but not 100% sure

melodb: Most of it is clean, thanks! Only two metadata need to be reverted at the end (`display_name` and `version`).

melodb:
I took the freedom to commit directly. I have just changed the call to "set_model_colors" to "get_default_colors" as I have suggested in a previous comment. I hope that's ok for you.

moPeterAndrews:
Yeah, absolutely no problem! Sorry for missing that

melodb:
Once the metadata in the notebook is clean, all is good from my side. I will let @lionel42 make the final check and accept the PR. Thanks a lot Peter for all the work!

lionel42:
@melodb @moPeterAndrews I went to some more changes after merging with the main branch.

I also added the new function in the example notebook, and fixed it for what was needed.

Can you test your main notebooks if they still work as intended ?

lionel42 (approved):

fluxy/operators/mf.py
```diff
@@ -74,6 +73,19 @@ def compute_mf_difference(
...
     )
+
+    # check species
+    species_set = {ds_left.attrs["species"], ds_right.attrs["species"]}
+    if len(species_set) != 1:
+        logger.warning(
+            f"Different species found {species_set} between the 2 compared datasets."
+        )
+        species = species_set
+        species_set = "mix"
```
melodb: I think here the better correction would be to replace L.83-84 by:
`species = "mix"`
Otherwise, the attribute in L.98 would be equal to a set. Moreover, how it stands, `species_set = "mix"` is not used anywhere.

melodb (commented):
I will be on sick leave until the end of the day so @mo-aliceramsden and @moPeterAndrews I will leave the testing for you if that's ok. I left only a comment from the revision I started yesterday.

moPeterAndrews:
Any plotting of timeseries has incorrect uncertainties, they're far bigger here than in the "devel" branch. I'd suggest reverting commit #[a64b379](https://github.com/openghg/fluxy/pull/257/commits/a64b379014cd5db07ccf915c58c945b767408ef6), as this will bring it back into alignment with the "devel" branch.

It looks like this commit was made to resolve the discrepancy between "FillBetween" and "Errorbar" uncertainty styles, but the wrong one was corrected. A better fix might be:

`  mean = da.sel(percentile="mean").values

    if plot_type == "FillBetween":
        ax.fill_between(
            time_as_datetime,
            y1=min_unc,
            y2=max_unc,
            alpha=0.2,
            color=da.attrs["plot_color"],
        )

    elif plot_type == "Errorbar":
        ax.errorbar(
            time_as_datetime,
            y=da.sel(percentile="mean"),
            yerr=[mean-min_unc, max_unc-mean],
            alpha=0.4,
            fmt="none",
            color=da.attrs["plot_color"],
        )`

lionel42:
@mo-aliceramsden Have you had time to test this branch ? Can we merge it ?

mo-aliceramsden (approved):
I've tested out the new plotting option and this is all working well for me. I'm happy for this to be merged in to devel

PR #261, opened 2026-01-11
Flux map period comparison

### LINKED ISSUE #260: New flux-map function to compare two periods
Add a new function in flux map to enable comparing two periods with a "difference plot".

hdelongueville:
New function to enable comparing flux maps of two periods.

- [x] close #260 
- [x] add test

### DIFF
--- a/fluxy/operators/flux_map_resample.py
+++ b/fluxy/operators/flux_map_resample.py
@@ -805,7 +805,8 @@ def resample_over_period(
             Interval length for custom periods (e.g., for months or years).
         chop_by (str, list):
             Defines how the dataset should be chopped.
-            Options are: 'year', 'month', 'season', None, a list of dates or months, or a season.
+            Options are: 'year', 'month', 'season', None, a list of dates or months, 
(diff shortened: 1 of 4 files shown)


mo-aliceramsden (approved):
This is all working well for me, happy to have this merged in.

I found some minor updates that could be made to the flux_map_over_time plot titles, that I found whilst testing this PR, but are unrelated to the changes made here. These include using a comma to separate month names instead of a dash (for example, when I asked it to plot Jan and May, the plot title was Jan-May, which was a bit confusing) and only plotting one date when the start and end date are the same (e.g. when plotting one year of data the title is now 2024 instead of 2024-2024). These are only minor issues, so can be reverted and added to another PR later, if that's easier

hdelongueville:
Thanks for testing it! 
Thanks for the suggestions. I think it would be nice to implement them. I suggest making a separate PR. @mo-aliceramsden Can I ask you to create an issue with your suggestions?

hdelongueville:
Oh I see you already made the changes, even better. Thanks!!

PR #268, opened 2026-02-09
Bugfix fluxmap

melodb:
- Revert labels to previous implementation (to be improved by Alice in a later PR)
- Add h5py, numpy and pandas to list of dependencies

**Note:**
The tests pass on GitHub with the following numpy/pandas dependencies:
"numpy==1.26.4"
"pandas==2.2.3"
 
If I don't specify the versions, the following were being installed:
numpy-2.4.2
pandas-3.0.0
 
And the tests were failing. Maybe we need to make the code compatible with the latest numpy/pandas in a later PR. If you agree with the current fix, I will create an issue for this.


lionel42 (approved):
Looks good, we will revert the versions in a future PR

PR #269, opened 2026-02-09
removed strftime use in f-string to fix tests

mo-aliceramsden:
Fix for previous push to devel, that caused the tests to fail. This removes the use of 'strftime' in an f-string line.

This branch can be merged in after Lionel and Daniela have merged in Lionel's updates.


melodb:
Would you like to solve the conflicts with devel so that this fix can be merged in? Or maybe it is easier to simply delete this PR and create a new one that originates from the current status of devel

mo-aliceramsden:
> Would you like to solve the conflicts with devel so that this fix can be merged in? Or maybe it is easier to simply delete this PR and create a new one that originates from the current status of devel

Yes I'll delete this branch as this fix is already in the devel branch. Thanks for sorting this!

PR #272, opened 2026-02-13
Bug fix

melodb:
The for-loop over all axes was replaced by a for-loop over all axes that have a max value associated. This avoids an out-of-bounds error when there are more subplots than regions to plot (for example, when 5 regions are specified but a 3x3 figure is created).


mo-aliceramsden (approved):
Looks good to me! Happy for this to be merged in

PR #277, opened 2026-02-20
Improve align_time

melodb:
- create function "check_times_within_tolerance" which can be used for only_overlapping=True/False.
- replace concatenation of t1 and t2 by pandas date_range array.
- use reindex with a given tolerance (method="nearest").
- simplify error message to only print missing time stamp per model/species.

### DIFF
--- a/fluxy/operators/flux_align_dataset.py
+++ b/fluxy/operators/flux_align_dataset.py
@@ -118,13 +118,9 @@ def align_time(
     # Check for added NaNs
     for ds in aligned_ds_list:
         test_var = list(ds.data_vars)[0]
-
-        if (ds[test_var].ndim) == 1:
-            mask_no_data = ds[test_var].isnull().values[0]
-        else:
-            mask_no_data = []
-            for i in range(ds.sizes["time"]):
-                mask_no_data.append(ds[test_var].isel(time=i).isnull().all().values)
+        mask_no_data = []
+        for i in range(ds.sizes["time"]):
+            mask_no_data.append(ds[test_var].isel(time=i).isnull().all().values)
 
         if np.any(mask_no_data):
             times_no_data = ds.time.values[mask_no_data]
--- a/fluxy/operators/flux_align_dataset.py
+++ b/fluxy/operators/flux_align_dataset.py
@@ -109,7 +109,6 @@ def align_time(
     ).values
 
     # Check if time difference is within reasonal bounds before reindexing
-    print(target_time)
     [check_times_within_tolerance(ds, target_time, tolerance) for ds in ds_list]
     aligned_ds_list = [
         ds.reindex(time=target_time, method="nearest", tolerance=tolerance)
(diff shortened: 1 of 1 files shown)


mo-aliceramsden (approved):
I had to add a minor fix in for the list of target times, adding one period onto the end of the list (otherwise the last time period was skipped and check_times_within_tolerance was failing for me). Otherwise, this all looks ok to me, thanks for working on this!

melodb:
Thanks for the fix! Is it ok if I merge it to your branch?

melodb:
> I had to add a minor fix in for the list of target times, adding one period onto the end of the list (otherwise the last time period was skipped and check_times_within_tolerance was failing for me). Otherwise, this all looks ok to me, thanks for working on this!

Is there a way for me to reproduce your error? What datasets/species did you use?
I'm now wondering if adding a full period is the best option. When `only_overlapping = True`, we subtract/sum half of the period to the min/max dates. When I tried the code, I did not feel the need for it, but maybe that's the saffer approach?

mo-aliceramsden:
> > I had to add a minor fix in for the list of target times, adding one period onto the end of the list (otherwise the last time period was skipped and check_times_within_tolerance was failing for me). Otherwise, this all looks ok to me, thanks for working on this!
> 
> Is there a way for me to reproduce your error? What datasets/species did you use? I'm now wondering if adding a full period is the best option. When `only_overlapping = True`, we subtract/sum half of the period to the min/max dates. When I tried the code, I did not feel the need for it, but maybe that's the saffer approach?

I was trying to plot all_pfc, between any start date and and end_date of '2025-01-01'. And I think the error was happening during 

> > I had to add a minor fix in for the list of target times, adding one period onto the end of the list (otherwise the last time period was skipped and check_times_within_tolerance was failing for me). Otherwise, this all looks ok to me, thanks for working on this!
> 
> Is there a way for me to reproduce your error? What datasets/species did you use? I'm now wondering if adding a full period is the best option. When `only_overlapping = True`, we subtract/sum half of the period to the min/max dates. When I tried the code, I did not feel the need for it, but maybe that's the saffer approach?

I was trying to plot all_pfc for InTEM_NAME_EUROPE_NID2026 between any start_date and and an end_date of '2026-01-01'. I think the error was occuring because the last timestamp in target_ds_time was about '2024-07-01', so the '2025' timestamp had been missed off the list.

melodb:
Oh, I see. So the result of `max_date = max([x.time.max() for x in ds_list])` is 2025-07-01 (or something similar) but when running `pd.date_range()` it stops before 2025. As far as I understand the description of date_range (https://pandas.pydata.org/docs/reference/api/pandas.date_range.html), it includes the end date, but we might have shifts of a couple of days that push the final data point out of the time range.

Would you mind trying a different fix? Adding half the period to max date like for `only_overlapping=True`. That is:
`max_date = max([x.time.max() for x in ds_list]) + period / 2`

This will make it more consistent with case `only_overlapping=True`, will avoid an extra data point at the end for the cases I have tested, and will make `target_time` easier to read.

Just a suggestion. Feel free to leave it as is. Also feel free to merge this branch to yours whenever you find appropriate.

PR #299, opened 2026-04-22
Performance improvements for concentration data

stiglers-eponym:
Improve speed and reduce memory requirement for plotting concentration data:
- Use xr.Dataset.isel instead of xr.Dataset.where when applicable, since isel can be much faster.
- Use ax.plot instead of ax.scatter to plot data points which share the same size, marker and color. This significantly reduces the memory footprint.

Note: ax.plot uses different units to measure marker sizes. The provided marker size ("ms") should match approximatly.

### DIFF
--- a/fluxie/plots/mf_timeseries.py
+++ b/fluxie/plots/mf_timeseries.py
@@ -920,7 +920,7 @@ def plot_sites_timeseries(
                 data,
                 ".",
                 color=model_colors[m][0],
-                ms=5,
+                ms=3,
                 label=model_labels_copy[m],
             )
 
--- a/fluxie/plots/mf_timeseries.py
+++ b/fluxie/plots/mf_timeseries.py
@@ -915,11 +915,12 @@ def plot_sites_timeseries(
             if separate_by_height:
                 mask &= ds_all[m]["intake_height"] == height
             data = ds_all[m]["time"].where(mask, drop=True)
-            ax.scatter(
+            ax.plot(
                 (site_iter + model_offset * i - 0.5 + margin) * np.ones(data.size),
                 data,
-                c=model_colors[m][0],
-                s=2,
+                ".",
+                color=model_colors[m][0],
+                ms=5,
                 label=model_labels_copy[m],
             )
 
--- a/fluxie/plots/mf_timeseries.py
+++ b/fluxie/plots/mf_timeseries.py
@@ -918,9 +918,9 @@ def plot_sites_timeseries(
             ax.plot(
                 (site_iter + model_offset * i - 0.5 + margin) * np.ones(data.size),
                 data,
-                ".",
+                "o",
                 color=model_colors[m][0],
-                ms=3,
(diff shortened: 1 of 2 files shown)


lionel42:
Thanks for the PR. Do you have a small showcase on how much speedup you gain from this change ?

stiglers-eponym:
I did not check the speedup for exactly this case, but a very similar check using fluxie data gave me a speedup by factor 50.

Showcase: Take some fluxie concentration file with many observations. I use our ITMS demonstrator (ICON) simulation covering 4 years.
```python
>>> import xarray as xr
>>> from timeit import timeit
>>> ds = xr.open_dataset("/path/to/model_species_frequency_concentrations.nc")  # adjust this
>>> ds.sizes["index"]
2765033
>>> timeit('ds.where(ds.number_of_identifier==10)', number=2, globals=dict(ds=ds))
2.527924904000429
>>> timeit('ds.isel(index=ds.number_of_identifier==10)', number=100, globals=dict(ds=ds))
2.215147246000015
```

The memory footprint is changed from crashing my computer by using too much memory to <4% total memory usage by jupyter.

fluxie/plots/mf_timeseries.py
```diff
@@ -915,11 +915,12 @@ def plot_sites_timeseries(
...
             data = ds_all[m]["time"].where(mask, drop=True)
-            ax.scatter(
+            ax.plot(
                 (site_iter + model_offset * i - 0.5 + margin) * np.ones(data.size),
                 data,
-                c=model_colors[m][0],
-                s=2,
+                ".",
+                color=model_colors[m][0],
+                ms=5,
```
lionel42: Why changing the marker size from 2 to 5 ?

stiglers-eponym: The scatter function uses different units for the marker size (I don't know why). With 5 it should approximately match. I did not compare directly because the old version led to a crash of my computer.

lionel42: Sorry, just saw your comment above

stiglers-eponym: Probably ms=3 would be closer to the original version.

lionel42:
@stiglers-eponym The default marker for scatter is `'o'`  https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.scatter.html 

What do you think about setting `plot` to use `'o'` instead of '.' as a marker so we can preserve the original size ?

stiglers-eponym:
> @stiglers-eponym The default marker for scatter is `'o'` https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.scatter.html
> 
> What do you think about setting `plot` to use `'o'` instead of '.' as a marker so we can preserve the original size ?

I checked in the documentation and probably for consistency one should use maker="o" and markersize=2**0.5. At least that looks consistent.

lionel42 (approved):

PR #305, opened 2026-05-01
Add new plot of a posteriori covariance between countries

ayanu:
- called with dict of flux datasets
- selects only those models that provide posterior covariance as variable
- further selection of countries for which covariance should be displayed
- aggregation over time (mean)
- normalisation of covariance (all values between -1 and 1)
- plots only lower triangle of matrix

### DIFF
--- a/fluxie/plots/flux_country_covariance.py
+++ b/fluxie/plots/flux_country_covariance.py
@@ -10,10 +10,10 @@ def plot_flux_country_covariance(
     model_labels: dict[str, str] = {},
     cov_max: float = 0.5,
     cmap_diff: str = "coolwarm",
-) -> plt.Figure: 
+) -> plt.Figure:
     """
     Plot posterior country covariance for each model that provides variable 'covariance_flux_total_posterior_country'.
-    Only lower triangle of matrix plotted. Covariance is normalised before plotting. Values will range between -1 and 1. 
+    Only lower triangle of matrix plotted. Covariance is normalised before plotting. Values will range between -1 and 1.
 
     Args:
         ds_all (dictionary of datasets):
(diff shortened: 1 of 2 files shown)


lionel42:
@ayanu Thanks for the contribution,


Do you think you can add your plot in the example notebook ? https://github.com/openghg/fluxie/blob/devel/scripts/example_basics.ipynb 

I am happy to help if needed ;)

ayanu:
Plot added to example notebook. 
Some additional improvements for plot included as well. 
We have a meeting with the PARIS modellers tomorrow. I will ask for feedback on plot content and layout. Depending on that there might be additional adjustments before the branch should be merged.

lionel42:
@ayanu Any update from the meeting ? Should we proceed with the review ?

ayanu:
I presented the plot to the PARIS modellers today. There was no rejection. So I suggest merging if you don't see any technical issues.

lionel42:
@ayanu The comments from the review make sense

fluxie/plots/flux_country_covariance.py
```diff
@@ -0,0 +1,120 @@
...
+        key, value = pair
+        return check_var in value.data_vars
+
+    # filter for datasets that contain covariance
+    ds = dict(filter(filter_cov, ds_all.items()))
+
+    # number of models required to decide on plot layout
+    models = list(ds.keys())
+    nn_models = len(ds)
+    if nn_models == 0:
```
lionel42, line 51: In this case I would raise and error instead of warning. 
It makes more sense that if you cannot

I imagine you warned for the tests and examples to pass, since the test data doesn't contain the covariance variable.
In this case , we could maybe add this variable in the test data if it is not too difficult to add ?

fluxie/plots/flux_country_covariance.py
```diff
@@ -0,0 +1,120 @@
...
+    """
+
+    # function for selecting only datasets with covariance variable
+    def filter_cov(pair):
+        check_var = "covariance_flux_total_posterior_country"
+        key, value = pair
+        return check_var in value.data_vars
+
+    # filter for datasets that contain covariance
+    ds = dict(filter(filter_cov, ds_all.items()))
```
lionel42, line 46: Rename `ds` to `dss` for consistency when there is a list or dict of datasets

fluxie/plots/flux_country_covariance.py
```diff
@@ -0,0 +1,120 @@
...
+        return None
+
+    # start plot layout
+    fig, ax = plt.subplots(ncols=nn_models, figsize=(10, 10 * nn_models), squeeze=False)
+    fig.tight_layout()
+    ax = ax.flatten()
+    for ii, mod in enumerate(ds):
+
+        #  extract the data from each model
+        msk = range(0, len(ds[mod].country))
```
lionel42, line 62: rename to `country_mask` or other name for clarity

melodb:
I didn't look at the code, but I would suggest adding the call to this function to "PARIS_inversion_results.ipynb" to simplify its usage within the PARIS group.

lionel42:
@ayanu  Please let me know if you need help on this

lionel42:
@ayanu  What is the current status on this PR, do you need help ?

Issue #311, opened 2026-06-24
File format fails some CF-convention checks


ayanu:
Latest file format (as uploaded to ICOS Carbon Portal) fails some checks when running convention checker. 

https://mcc.podaac.earthdatacloud.nasa.gov/  (Select CF version 1.8). 

Problems are indicated for flux file:
- order of dimensions for variables using dimensions other than time and space. 
- cell_methods attribute (not valid)
- time_bnds attributes 

Tool failed to run for concentration file.

ayanu:
[commit](https://github.com/openghg/fluxie/commit/d0527b146097607aa3dd496b1129e0e5e22f2fa0) suggests an update to the flux file template that should make CF conform. 
Problem: Order of dimensions for many variables changed
@lionel42 @brendan-m-murphy will this change in order break the IO for fluxie? 

Making the format CF conform is rather critical as the ICOS Carbon Portal informed me that they will make it a requirement for uploads in the future.

lionel42:
@ayanu In most cases fluxie is using xarray for the operations which makes it non dependant on the order of dimensions. 
So in theory it should not be a problem.

It can be however that in some parts of the code we used numpy directly instead of xarray so there it will fail. 

The best would be to add a test netcdf file with the new order of dimension and see where it breaks and check for potential fixes.

ayanu:
Some test netcdf files that follow the format update can be found here:
https://drive.switch.ch/index.php/s/1E2t38HRWbHrQgd 

They all passed the CF checker https://mcc.podaac.earthdatacloud.nasa.gov/  and where created from ELRIS output. 

I ran them through my default notebook, but that's probably not testing everything. 
@lionel42 Can you do some additional tests?

lionel42:
@ayanu  I also tested these files with the test notebook and it works fine.

lionel42:
What should we do from this ? Do we simply change the format specification ?
@ayanu

ayanu:
Good question. I feel we need to change the release version as well as the format is something fundamental. Even if no changes to the code are done, it would allow us to reference the file format change with a fluxie version. What would you suggest?

lionel42:
Yes, since it does not break any code, we could do the minor version change. so it would be for v2.3.0

ayanu:
For me a minor version change makes sense. Will you make the required changes in the "file_format" branch and then merge or how do we proceed?

lionel42:
I just opened a PR, we can merge and then I can setup a release. 

[...]

PR #312, opened 2026-06-29
File format

lionel42:
#311

### DIFF
--- a/data/templates/PAR-AVE-EYE_inversion_concentration_output.cdl
+++ b/data/templates/PAR-AVE-EYE_inversion_concentration_output.cdl
@@ -17,9 +17,6 @@ variables:
         time:bounds = "time_bnds" ;
 		time:standard_name = "time" ; 
     double time_bnds(index, nbnds) ; 		// mandatory 
-		time_bnds:long_name = "start and end points of each time step" ; 
-		time_bnds:calendar = "proleptic_gregorian" ;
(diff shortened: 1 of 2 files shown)


data/templates/PAR-AVE-EYE_inversion_concentration_output.cdl
```diff
@@ -17,9 +17,6 @@ variables:
         time:bounds = "time_bnds" ;
 		time:standard_name = "time" ; 
     double time_bnds(index, nbnds) ; 		// mandatory 
-		time_bnds:long_name = "start and end points of each time step" ; 
-		time_bnds:calendar = "proleptic_gregorian" ;
-		time_bnds:units = "days since 1970-01-01 00:00:00" ;
```
lionel42, line 22: @ayanu  Is it on purpose that you had to remove the doc for the time bounds ?

ayanu: Yes. CF-convention does not accept attributes on bounds variables. Don't ask me why. I suppose because they are linked to the time variable anyway and hence it does not make sense to use different attributes as for time itself.

lionel42: Great, then I guess you can approve the PR and we can merge it ? @ayanu

ayanu (approved):

PR #313, opened 2026-06-29
Fix tests for matplotlib v11

lionel42:
A recent update of matplotlib has made the tests fail.

### DIFF
--- a/fluxie/operators/flux_prepare_inventory.py
+++ b/fluxie/operators/flux_prepare_inventory.py
@@ -1,7 +1,7 @@
 import numpy as np
 import xarray as xr
 
-from matplotlib.cm import get_cmap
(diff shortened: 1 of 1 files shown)


lionel42:
@ayanu If you can approve the PR, we Just need to merge this before #311

ayanu (commented):
Does this mean that the dependency for matplotlib (which version to use with fluxie) needs to be updated as well? 
I assume older versions of matplotlib will not work with this code update.

lionel42:
No it is backward compatible, so older versions will work as well (since mpl 3.6 : https://matplotlib.org/3.6.3/api/_as_gen/matplotlib.pyplot.get_cmap.html)

ayanu (approved):

PR #314, opened 2026-07-03
Add alternative way to calculate posterior country uncertainty for aggregated regions

ayanu:
Currently, if no posterior covariance matrix is available for country-level emissions, fluxie sets posterior uncertainties to NA.

Alternatively, posterior percentiles or standard deviation can be used for the calculation of country posterior uncertainty (same way as it is done for prior uncertainty).

A warning is issued when posterior uncertainty is calculated without taking covariance into account.

### DIFF
--- a/fluxie/operators/regions.py
+++ b/fluxie/operators/regions.py
@@ -183,9 +183,9 @@ def _extract_region_flux_sector(
                     (
                         (
                             ds_region[f"flux_{sector}_posterior_country"]
-                            - ds_region[f"percentile_flux_{sector}_posterior_country"].isel(
-                                percentile=min_percentile_index
-                            )
+                            - ds_region[
+                                f"percentile_flux_{sector}_posterior_country"
+                            ].isel(percentile=min_percentile_index)
                         )
                         ** 2
                     ).sum(dim="country")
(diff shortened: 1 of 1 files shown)


fluxie/operators/regions.py
```diff
@@ -175,7 +175,30 @@ def _extract_region_flux_sector(
...
+                    (
+                        (
+                            ds_region[f"flux_{sector}_posterior_country"]
+                            - ds_region[
+                                f"percentile_flux_{sector}_posterior_country"
+                            ].isel(percentile=min_percentile_index)
+                        )
+                        ** 2
+                    ).sum(dim="country")
+                )
```
lionel42, line 192: Shouldn't this elif be placed in the previous if else statement since it calculates sigma_prior and this if else statement is about the sigma_posterior ?
Or is the assignement going to sigma_posterior ?

ayanu: You are right. I copied that code and missed exchanging all 'priror' with 'posterior'. Thanks for spotting this. 
Will push an update.

lionel42 (approved):
Would be nice to add a test, but I guess it is not so important and not that easy to add one.

lionel42:
@ayanu  If this is ready, you can merge

lionel42:
@ayanu ready to merge this ?

PR #317, opened 2026-08-13
Allow simultaneous plotting of prior/posterior stats

melodb:
- function "stats_mf" now accepts `stats_type` to be a list of strings instead of one single string. The diff does not show up very nicely: I have basically moved everything inside a for loop over the elements of `stats_type` and appended the data frames.
- variable `stats_type` is now a column in the dataframe created in function "stats_observed_vs_simulated".
- plotting function "plot_stats_mf" was updated accordingly.
- option to arrange subplots  horizontally was also added.

### DIFF
--- a/fluxie/operators/mf.py
+++ b/fluxie/operators/mf.py
@@ -10,6 +10,9 @@ from fluxie.operators.stats import stats_observed_vs_simulated
 logger = logging.getLogger(__name__)
 
 
+StatsType = Literal["prior", "posterior", "prior_above_BC", "posterior_above_BC"]
+
+
 def compute_mf_difference(
     ds_all: dict[str, xr.Dataset], models_to_subtract: list[str]
 ) -> dict[str, xr.Dataset]:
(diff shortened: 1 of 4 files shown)


fluxie/operators/mf.py
```diff
@@ -136,58 +137,71 @@ def stats_mf(
...
-        sim_var=sim,
-        sites=sites,
-    )
+    if isinstance(stats_type, str):
+        stats_type = [stats_type]
+
+    dfs = []
+    for stat in stats_type:
+        # assure that stats_type in allowed options
+        type_options = ["prior", "posterior", "prior_above_BC", "posterior_above_BC"]
```
lionel42: use directly the values from the Literal above. see https://docs.python.org/3/library/typing.html#typing.get_args

fluxie/operators/stats.py
```diff
@@ -10,6 +10,7 @@ def stats_observed_vs_simulated(
     obs_var: str,
     sim_var: str,
     sites: list = None,
+    stats_type: str = "Undefined",
```
lionel42: could use the literal type here or `None` instead of a undefined string for clarity

lionel42 (approved):
Just very little changes

PR #318, opened 2026-08-19
Sector flux updates

mo-aliceramsden:
Various updates to make sector-level results calculation and plotting a little easier:
- A notebook containing examples on how to create and plot sector-level results.
- Updated read in of sector-level prior data (which is used to create sector-level inversion results).
- Fixes for aligning sector-level prior data and inversion results and calculating the rolling mean.
- Better error checking for missing sectors.
- Added read in of sector-level inventory results. This data is currently only available for the UK, IRL and NLD (in an older file) but we could update this before the PARIS plotting.
- An option for sector-level spatial flux plotting (rather than the default 'total').

Other changes:
- Updated options to select and plot data from certain months.
- Added new species definitions.
- Added optional print out of trends in region fluxes.
- Added a met office format results table. This is not PARIS related, but very useful for our work, so I've added this to a separate file in the 'cli' directory.
- Fixed a bug that caused 3-monthly inversion spatial flux plotting to fail.
- A few more plotting options to add/remove plot components.


fluxie/operators/flux_prepare_inventory.py
```diff
@@ -1,7 +1,7 @@
 import numpy as np
 import xarray as xr
 
-from matplotlib.pyplot import get_cmap
+from matplotlib.cm import get_cmap
```
micstein89, line 4: This currently raises an ImportError with our current Matplotlib version (get_cmap is no longer available from matplotlib.cm). Could you replace this with the current Matplotlib API, e.g. matplotlib.colormaps[...]?

fluxie/cli/MO_annual_report.py
```diff
@@ -0,0 +1,254 @@
...
+
+            all_sp_res[r][s] = create_str_dataframe(
+                annual_res[s],
+                inventory_years,
+                species_list,
+                model=all_read_in,
+                table_start_date=min(start_date),
+                region=r,
+                include_inventory_uncert=include_inventory_uncert,
+                n_digits=2,
```
micstein89, line 90: n_digits=n_digits,

fluxie/cli/MO_annual_report.py
```diff
@@ -0,0 +1,254 @@
...
+import numpy as np
+import os
+import logging
+from fluxie.cli.utils_annex_plot import create_str_dataframe
+
+logger = logging.getLogger(__name__)
+
+
+def make_AR_table(
+    df: pd.DataFrame,
```
micstein89, line 11: df: dict[str, pd.DataFrame],

fluxie/cli/MO_annual_report.py
```diff
@@ -0,0 +1,254 @@
...
+            if sectors == ["total"]:
+                sector_name = ""
+                sector_name_csv = ""
+
+            else:
+                sector_name = f" {s.capitalize()}"
+                sector_name_csv = f"_{s.capitalize()}"
+
+            if include_inventory:
+                type_title += f" & Inventory{sector_name} & {inv_model}{sector_name}"
```
micstein89, line 132: This causes a LaTEX error if inv_model contains an underscore (e.g. RHIME_ch4).
Possible fix:
inv_model_latex = inv_model.replace("_", `r"\ _"`)
type_title += f" & Inventory{sector_name} & {inv_model_latex}{sector_name}"

fluxie/cli/MO_annual_report.py
micstein89, line 1: This routine could be simplified significantly. Once a DataFrame is constructed, the CSV output could be written with DataFrame.to_csv(). Similarly, DataFrame.to_latex() could simplify the LaTeX generation (this would introduce an additional Jinja2 dependency).

aliceramsden: Thanks for these suggestions (I didn't know about DataFrame.to_latex(), that should be useful thanks!). For now, these functions were just intended for internal Met Office use, so don't need optimising for use by other fluxie users. The other table creation functions (e.g. those in the PARIS annex plot generators) should work with sector-level data, if we want to use sector-level tables in this year's PARIS reports.

data/tests/PRIOR/ch4/PRIOR_EUROPE_EDGAR_v2025_ch4_monthly.nc
lionel42, line 1: This file is quite large to be included in the test suite (we got 100kb for other nc test files). Is there a way to reduce it ? Maybe reducing the time range ?

configs/species_info.json
```diff
@@ -136,6 +136,56 @@
...
+    },
+    "h141b": {
+        "species_print":"HCFC-141b",
+        "gwp": 808,
+        "molar_mass": 116.90
+    },
+    "h142b": {
+        "species_print":"HCFC-142b",
+        "gwp": 2190,
+        "molar_mass": 100.45
```
lionel42, line 162: naming of the hfcf can be confused with the halons below. Can we have hcfc instead of just h ?

fluxie/plots/flux_timeseries.py
```diff
@@ -204,7 +206,8 @@ def prepare_data_to_plot(
     # Add original datasets to plot
     if not any(resample) or plot_resample_and_original:
         ds_original_flux = {
-            m: v for (i, (m, v)) in enumerate(ds_all_region.items()) if plot_separate[i]
+            m: v
+            for (i, (m, v)) in enumerate(ds_all_region.items())  # if plot_separate[i]
         }
```
lionel42, line 211: When you remove code like this. Don't leave the commented code, just delete it. (this applies also below and in other places in the PR)

fluxie/plots/flux_timeseries.py
```diff
@@ -315,12 +322,19 @@ def prepare_data_to_plot(
     return ds_to_plot
 
 
+def linear(x, m, c):
+    return m * x + c
```
lionel42, line 326: document that this is used for curve fit

fluxie/plots/flux_timeseries.py
```diff
@@ -315,12 +322,19 @@ def prepare_data_to_plot(
...
+
 def add_line_plot(
     ax: Axes,
     ds: xr.Dataset,
     variable: Literal["posterior", "prior"],
     highlighted_line: bool = False,
     add_unc: bool = False,
+    unit: str = None,
+    plot_trends: bool = False,
+    use_err: bool = True,
```
lionel42, line 337: document in docstring

fluxie/plots/flux_timeseries.py
```diff
@@ -566,11 +655,19 @@ def add_sector_barplot(
...
+        ds_sector.attrs["model_label"]
+        if "model_label" in ds_sector.attrs
+        else f"inventory"
+    )
+    res["country"] = (
+        ds_sector.attrs["country"] if "country" in ds_sector.attrs else None
+    )
+    res["species"] = (
+        ds_sector.attrs["species"] if "species" in ds_sector.attrs else None
+    )
```
lionel42, line 677: you can use `attrs.get("xxx", None)`

fluxie/plots/flux_timeseries.py
```diff
@@ -566,11 +655,19 @@ def add_sector_barplot(
...
 
-    return res
+    return res, width
```
lionel42, line 679: Dangerous change of signature ! What about adding the width in the res dict instead ?

fluxie/plots/flux_timeseries.py
```diff
@@ -674,9 +771,14 @@ def add_ylim(
...
-            df_country[df_country.columns.intersection(["mean_val", "max_unc"])]
-        )
+        if "max_unc" in df_country:
+            max_country = np.nanmax(
+                df_country[df_country.columns.intersection(["mean_val", "max_unc"])]
+            )
+        else:
+            max_country = np.nanmax(
+                df_country["mean_val"]
+            )
```
lionel42, line 790: put the if else statement directly in the nanmax function

fluxie/plots/flux_timeseries.py
```diff
@@ -938,21 +1047,15 @@ def secondary_unit_to_unit(y):
...
     ds_all: dict[str, xr.Dataset],
     species: str,
     plot_regions: list[str] | str = [],
     config_data: dict[str, dict] = {},
     model_colors: dict[str, list] = {},
     model_labels: dict[str, str] = {},
-    start_date: str | list[str] | None = None,
-    end_date: str | list[str] | None = None,
+    start_date: str | None = None,
+    end_date: str | None = None,
```
lionel42, line 1067: Why is the list not supported anymore ?

fluxie/plots/flux_timeseries.py
```diff
@@ -1060,16 +1167,15 @@ def plot_country_flux(
...
     )
-    date_length = len(ds_all)
-    start_date = parse_date(start_date, date_length)
-    end_date = parse_date(end_date, date_length)
 
-    start_date, end_date = dict(zip(ds_all.keys(), start_date)), dict(
-        zip(ds_all.keys(), end_date)
-    )
     # Sel data
+    if type(start_date) != list:
```
lionel42, line 1181: use isinstance to check types

lionel42 (changes requested):
There would be quite some things to clean (duplicates, legacy code left, ...) 

I already left some comments, but would probably do a second round. 

I can also offer to clean some of the code directly myself if that helps ?

fluxie/operators/flux_scale_by_sector_proportions.py
```diff
@@ -157,104 +164,149 @@ def scale_by_sector_proportions(
...
+                            ds[f"flux_{s}_{suff}_country"] = (
+                                ds[f"flux_{s}_{suff}"]
+                                * ds["country_fraction"]
+                                * ds["cell_area"]
+                            ).sum(dim=["latitude", "longitude"]) * units_factor
+
+                    # add atributes to created variables
+                    for suff in ["posterior", "prior"]:
+                        ds[f"flux_{s}_{suff}_country"].attrs = {
+                            "units": country_flux_units_print,
```
micstein89, line 306: There is a mismatch here regarding the units in the plot labels for the time series. These are read from the variables “flux_total_posterior_country,” “posterior,” “flux_total_prior_country,” and “prior.” So, for example, if you specify the unit ‘Gg yr-1’ for `scale_by_sector_proportions`, the quantities in the time series will be correct, but the plot labels will not.

fluxie/plots/flux_timeseries.py
micstein89, line 1323: start_date and end_date are not arguments of the function, although in the doc string (and in the code)

PR #322, opened 2026-09-04
extract_region_inventory_flux: raise KeyError if country is not found

### LINKED ISSUE #319: extract_region_inventory_flux returns sum of all countries if country is not found
# Current behavior
`fluxie.operators.regions.extract_region_inventory_flux()` selects Inventory data per country. If the function argument `country` does not match any known country name or country code, the function will skip the selection of the country and return the sum of inventories of all countries. This can lead to wrong inventory data in fluxie plots.

# Expected behavior
The function should not return wrong values if the country is not found. Either an exception should be raised, or (even better) the inventory should be skipped in the plot.

# Possible solutions
Possible behavior of `fluxie.operators.regions.extract_region_inventory_flux()` if country is not found:
- raise a KeyError. If this error is not caught, it may be inconvenient for a user who wants to create a plot with multiple countries when inventory data are only available for some countries.
- return an array of NaN values (and test whether the plots look as expected)
- print a warning and return an array of 0 values (and test whether the plots look as expected)
- return None instead of an array and ensure that calling functions handle that correctly.

stiglers-eponym:
Fixes #319: avoid plotting wrong inventory data.

### DIFF
--- a/fluxie/operators/regions.py
+++ b/fluxie/operators/regions.py
@@ -276,7 +276,7 @@ def extract_region_inventory_flux(
 
     Args:
         data_dir: directory which contains the data (should have inside a directory named 'inventory').
-        specie: Gas species, e.g. 'ch4'.
+        species: Gas species, e.g. 'ch4'.
         unit: unit in which the inventory should be converted.
         s_data: Dictionary of species with information for plotting (read from json file).
         r_data: Dictionary with country and region names (read from json file).
@@ -316,7 +316,7 @@ def extract_region_inventory_flux(
             f"No missing_data variable available in inventory files, assuming all data present."
         )
 
-    # first option left for compatability with older inventory netcdfs, can be removed later
+    # first option left for compatibility with older inventory netcdfs, can be removed later
     inv_ds = (
         inv_ds_all["inventory"]
         if "inventory" in inv_ds_all.keys()
@@ -359,6 +359,8 @@ def extract_region_inventory_flux(
             inv_stdev_ds = inv_stdev_ds.sel(country=country_search)
         elif country in inv_ds["country"]:
             inv_stdev_ds = inv_stdev_ds.sel(country=country)
+        else:
+            inv_stdev_ds = None
 
     if country_search in inv_ds["country"]:
         return inv_ds.sel(country=country_search), inv_stdev_ds  # new format
@@ -372,7 +374,9 @@ def extract_region_inventory_flux(
     available_countries = inv_ds["country"].values.astype(str)
     dict_regions: dict[str, str] = r_data.get("regions", {})
 
-    if country_search not in available_countries and country in dict_regions.keys():
+    if country_search in available_countries:
+        inv_ds = inv_ds.sel({"country": country_search})
+    elif country in dict_regions:
         region_search = dict_regions[country]
         country_list = region_search.split("-")
         inv_ds = inv_ds.sel({"country": country_list})
@@ -380,8 +384,8 @@ def extract_region_inventory_flux(
         logger.info(
             f"No inventory data available for {country}. Considering sum of individual countries: {region_search}"
         )
-    elif country_search in available_countries:
-        inv_ds = inv_ds.sel({"country": country_search})
+    else:
+        raise KeyError(f"Country not found in inventory: {country}")
 
     logger.info(
         "There is currently no option to plot inventory uncertainty for grouped countries. This functionality will be added later"
--- a/fluxie/plots/flux_timeseries.py
+++ b/fluxie/plots/flux_timeseries.py
@@ -1482,18 +1482,20 @@ def plot_all_species_stacked_bar(
         )
 
         try:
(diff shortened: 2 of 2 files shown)


fluxie/plots/flux_timeseries.py
```diff
@@ -1571,12 +1584,12 @@ def plot_all_species_stacked_bar(
...
-            flux_sum = ds_to_plot[species][models[s]]["posterior"].values
-            inventory_sum = inventories_to_plot[species][0].values
-        else:
-            flux_sum += ds_to_plot[species][models[s]]["posterior"].values
+        flux_sum += ds_to_plot[species][models[s]]["posterior"].values
+        try:
             inventory_sum += inventories_to_plot[species][0].values
+        except KeyError:
+            # Missing inventory data: Avoid plotting wrong data by just plotting nothing
+            inventory_sum += np.nan
```
lionel42, line 1597: What key error is being tested here ? for clarity it would be better to test if the key is there instead of try except.

fluxie/plots/flux_timeseries.py
```diff
@@ -1127,25 +1127,28 @@ def plot_country_flux(
...
+                    inventory_filename,
+                    sector,
+                    annex_mode,
+                    plot_inventory_uncertainty,
+                )
+                plotted_data_df = pd.concat(
+                    [plotted_data_df, inventory_df], ignore_index=True
+                )
+            except KeyError:
+                logger.warning(f"Missing inventory data for {species} in {country}")
```
lionel42: It is possible to forward the error message and write it in the warning statement:

```

except KeyError as ke:
    logger.warning(f"Failed .... due to {ke}") 

``` 

This would be helpful in this case for the user to know what country is missing from the warning.

fluxie/plots/flux_timeseries.py
```diff
@@ -1477,19 +1481,28 @@ def plot_all_species_stacked_bar(
...
+                    s_data,
+                    r_data,
+                    inventory_years,
+                    inventory_filename,
+                    sectors=sector,
+                )
+            )
+            this_uncert = inventories_uncert_to_plot[0]
+        except KeyError:
+            logger.warning(f"Failed to get inventory data for {species} in {regions}")
```
lionel42: Same comment as above

lionel42 (changes requested):

fluxie/plots/flux_timeseries.py
```diff
@@ -1477,19 +1481,28 @@ def plot_all_species_stacked_bar(
...
+            this_uncert = 0
```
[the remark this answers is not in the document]
stiglers-eponym: `inventories_to_plot` should not be filled with zero for missing values. This could lead to silent plotting of wrong data. It is better to keep `inventories_to_plot` incomplete. It should not crash because it is checked below whether `species`is in `inventories_to_plot`.

fluxie/operators/regions.py
```diff
@@ -372,16 +374,18 @@ def extract_region_inventory_flux(
...
         country_list = region_search.split("-")
         inv_ds = inv_ds.sel({"country": country_list})
 
         logger.info(
             f"No inventory data available for {country}. Considering sum of individual countries: {region_search}"
         )
-    elif country_search in available_countries:
-        inv_ds = inv_ds.sel({"country": country_search})
+    else:
+        raise KeyError(f"Country not found in inventory: {country}")
```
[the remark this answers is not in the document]
lionel42: Makes sense, @stiglers-eponym If you need help to add a test, please let me know

fluxie/plots/flux_timeseries.py
stiglers-eponym: The country is alreay included in the warning, but I can adjust the message of the warning to simplify debugging. For full debugging info one should set exc_info=True, but I do not think that a standard user needs that.

fluxie/plots/flux_timeseries.py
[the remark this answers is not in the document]
stiglers-eponym: One could also just keep the key error unhandled or add a warning. But it should be avoided that the user just sees wrong data. I think that adding zero instead of NaN will make it more likely that a user overlooks the missing inventory. It would therefore increase the risk of false interpretation.

One could also consider a case where the user only provides data for relevant inventory data, knowing that the missing data are (nearly) zero. This case is currently not handled.

stiglers-eponym: This can still be improved, but I do not have a real use case for this function for testing. I will check the key explicitly and add a warning for missing data.

fluxie/operators/regions.py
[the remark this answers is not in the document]
stiglers-eponym: Currently, there is no test function for `extract_region_inventory_flux(...)`. It makes sense to add test functions for the module `operators.regions` module, but I don't have time for that.

lionel42:
I have moved this to #323 so that I could add the tests. 

@stiglers-eponym  would you be interested in getting developers rights on the fluxie repo ? I think it would make sense since you are contributing quite often

